PR #17, opened 2022-12-17
Add Block Detail page

lukaw3d:
block

cc @csillag @buberdds @lubej

### DIFF
--- a/src/app/components/StyledDescriptionList/index.tsx
+++ b/src/app/components/StyledDescriptionList/index.tsx
@@ -1,0 +1,56 @@
+import { styled } from '@mui/material/styles'
+import { ReactNode } from 'react'
+
+interface Props {
+  /**
+   * Width of `<dt>` (first column)
+   *
+   * **Syntax**: `100px | 10% | 10ch | max-content | auto | minmax(min, max)`
+   */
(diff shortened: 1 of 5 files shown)


src/app/components/StyledDescriptionList/index.tsx
```diff
@@ -0,0 +1,56 @@
+import { styled } from '@mui/material/styles'
+import { ReactNode } from 'react'
+
+interface Props {
+  /**
+   * Width of `<dt>` (first column)
+   *
+   * **Syntax**: `100px | 10% | 10ch | max-content | auto | minmax(min, max)`
+   */
+  titleWidth?: string
```
csillag, line 10: React is not happy about the way we are passing `titleWidth` here:

> Warning: React does not recognize the `titleWidth` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `titlewidth` instead. If you accidentally passed it from a parent component, remove it from the DOM element.

csillag (approved):
Reviewed, LGTM.

There is a React warning (about passing some styling parameter), but we can fix that later.

csillag:
I am merging this, since Luka is on vacation

PR #37, opened 2022-12-22
Add duration select

lubej:
Add duration select in Dashboard page

image


src/app/pages/DashboardPage/ParatimeSnaphosts.tsx
```diff
@@ -0,0 +1,50 @@
...
+import { ChartDuration } from '../../utils/chart-utils'
+import { useTranslation } from 'react-i18next'
+import { useConstant } from '../../hooks/useConstant'
+
+export const ParatimeSnaphosts: FC = () => {
+  const { t } = useTranslation()
+  const defaultChartDurationValue = useConstant<ChartDuration>(() => ChartDuration.TODAY)
+  const [chartDuration, setChartDuration] = useState<ChartDuration>(defaultChartDurationValue)
+
+  const handleDurationSelectedChange = useCallback((duration: ChartDuration | null) => {
```
buberdds, line 16: no need to use useCallback if function just use setState

src/app/pages/DashboardPage/ParatimeSnaphosts.tsx
```diff
@@ -0,0 +1,50 @@
...
+import { COLORS } from '../../../styles/theme/colors'
+import { DurationSelect } from './DurationSelect'
+import { TransactionsChartCard } from './TransactionsChartCard'
+import { ChartDuration } from '../../utils/chart-utils'
+import { useTranslation } from 'react-i18next'
+import { useConstant } from '../../hooks/useConstant'
+
+export const ParatimeSnaphosts: FC = () => {
+  const { t } = useTranslation()
+  const defaultChartDurationValue = useConstant<ChartDuration>(() => ChartDuration.TODAY)
```
buberdds, line 13: how that hook works? can't we just use ChartDuration.TODAY ?

buberdds (approved):

src/app/pages/DashboardPage/ParatimeSnaphosts.tsx
lubej: It works the same way if you would define a variable outside of the component. But since we are keeping types and styles in the same file it's messy. I would like to keep it this way.

PR #38, opened 2022-12-22
Transactions List Page

buberdds:
PR adds transactions list page with offset pagination.
Transaction list is extracted from Dashboard Latest Transaction
For pagination I had to use [headless pagination component](https://mui.com/material-ui/react-pagination/#UsePagination.tsx), because it was impossible to swap next/first/prev/last etc.

what's not in this PR
- polling / marking new items with diff background color 

Screenshot 2022-12-22 at 12 59 59

### DIFF
--- a/src/app/components/Pagination/index.tsx
+++ b/src/app/components/Pagination/index.tsx
@@ -1,0 +1,100 @@
+import { FC } from 'react'
+import { useTranslation } from 'react-i18next'
+import usePagination, { type UsePaginationProps } from '@mui/material/usePagination'
+import { styled } from '@mui/material/styles'
+import NavigateNextIcon from '@mui/icons-material/NavigateNext'
+import NavigateBeforeIcon from '@mui/icons-material/NavigateBefore'
+import { COLORS } from '../../../styles/theme/colors'
+import { useSearchParams } from 'react-router-dom'
+
+const StyledList = styled('ul')(({ theme }) => ({
+  listStyle: 'none',
+  padding: `${theme.spacing(5)} 0 0`,
+  margin: 0,
+  display: 'flex',
+}))
+
+const StyledListElement = styled('li')({
+  padding: 0,
+  margin: 0,
+  display: 'inline-flex',
+})
+
+const StyledButton = styled('button')(({ theme }) => ({
+  display: 'inline-flex',
+  background: 'none',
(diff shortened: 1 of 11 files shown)


src/app/components/Table/TablePagination.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+type TablePaginationProps = {
+  count: number
+  rowsNumber: number
+}
+
+export const TablePagination: FC<TablePaginationProps> = ({ count, rowsNumber }) => {
+  const [searchParams, setSearchParams] = useSearchParams()
+  const offsetSearchQuery = searchParams.get('offset')
+  const offset = offsetSearchQuery ? parseInt(offsetSearchQuery, 10) : 0
```
lubej, line 13: This could potentially throw. Would be nice to wrap in try/catch.

src/app/components/Pagination/index.tsx
```diff
@@ -0,0 +1,100 @@
...
+                {t('pagination.first')}
+              </StyledFirstPageButton>
+            )
+          } else if (type === 'last') {
+            children = (
+              <StyledLastPageButton type="button" {...item}>
+                {t('pagination.last')}
+              </StyledLastPageButton>
+            )
+          }
```
lubej, line 88: Would be nice to refactor this into some utility.

lubej (approved):
Overall LGTM

src/app/components/Table/TablePagination.tsx
buberdds: It can throw when this meet what conditions ?

lubej: The string could be invalid integer, if this is coming from url.

src/app/components/Pagination/index.tsx
buberdds: Ok, I will refactor this part in a next PR after holiday season if that's Ok with you.

src/app/components/Table/TablePagination.tsx
lubej: My bad, I meant the bellow code for page calculation.

buberdds: aa Ok I can see it now, thanks

PR #41, opened 2022-12-22
Latest blocks page

csillag:
It's like this:

Marking new blocks with different bg color is not yet implemented.

image


csillag:
I have merged another PR, so now this needs to be rebased, just a minute...

src/app/components/Blocks/index.tsx
```diff
@@ -0,0 +1,129 @@
...
+const gasLimit = 1000000 // TODO temporary value
+
+export const Blocks = (props: BlocksProps) => {
+  const { isLoading, blocks, verbose, pagination, limit } = props
+  const { t } = useTranslation()
+
+  const tableColumns: TableColProps[] = [
+    { content: t('common.table.fill') },
+    { content: t('common.table.block'), align: TableCellAlign.Right },
+    verbose ? { content: t('common.table.timestamp') } : undefined,
```
lubej, line 29:
```suggestion
    ...(verbose ? [{ content: t('common.table.timestamp') }] : []),
```

This would be better, and getting rid of `filter` on array, which would reduce complexity.

csillag:
Rebased.

On the block list, the links to the individual blocks work now. (Since the block details page has been merged.)

src/app/pages/BlocksPage/index.tsx
```diff
@@ -1,19 +1,43 @@
...
+
+import { Blocks } from '../../components/Blocks'
+import { NUMBER_OF_ITEMS_ON_SEPARATE_PAGE } from '../../config'
+
+const PAGE_SIZE = NUMBER_OF_ITEMS_ON_SEPARATE_PAGE
 
 export const BlocksPage: FC = () => {
-  const { data } = useGetEmeraldBlocks({}, { query: { refetchInterval: 5000 } })
+  const { t } = useTranslation()
+  const { offsetString = '0' } = useParams()
```
lubej, line 17: Can we use `useSearchParams` for getting the offset?

src/app/config/index.ts
```diff
@@ -0,0 +1,2 @@
+export const NUMBER_OF_ITEMS_ON_DASHBOARD = 5
+export const NUMBER_OF_ITEMS_ON_SEPARATE_PAGE = 10
```
lubej, line 2: Would be more descriptive with something like `TABLE_PAGE_SIZE`, but either way, we can change that later.

lubej (approved):
Overall LGTM, would just replace `useParams` with `useSearchParams`

src/app/components/Blocks/index.tsx
csillag: Actually I started out that way, but then I decided that enumerating all the columns that way is too much of PITA, so I want for the filter. But it doesn't matter either way, I can revert to the previous method.

lubej: I agree it's not readable, but it is be more efficient. If you want you can add an utility for it.

src/app/pages/BlocksPage/index.tsx
csillag: I have now implemented that  in the API layer, adding a wrapper around the generated API, just like we have it for transactions.

src/app/components/Blocks/index.tsx
csillag: OK I changed it as suggested

PR #44, opened 2023-01-03
Transactions List Page: mark new items

buberdds:
- add optional `markAsNew` prop to a table row 
- compare prev and new state to mark new items in TransactionsPage. api is not indexing new txs so this is not battle tested.


src/app/pages/TransactionsPage/index.tsx
```diff
@@ -3,19 +3,44 @@ import { useTranslation } from 'react-i18next'
...
+          }
+          return {
+            ...nextState,
+            data: {
+              transactions: nextState.data?.transactions?.map(tx => {
+                return {
+                  ...tx,
+                  markAsNew: !previousState.data?.transactions?.some(prevTx => prevTx.hash === tx.hash),
+                }
+              }),
```
lubej, line 33: Looks to me like spread should be in deeper level. Wouldn't this discard all previous transactions in `previousState`?

lubej (approved):

src/app/pages/TransactionsPage/index.tsx
buberdds: In what way it can discard previous transactions? you mean like I am mutating it somewhere or not returning prev state ?

buberdds: spread seems Ok
Screenshot 2023-01-03 at 12 48 49

lubej: By not returning prevState. But I guess that is expected behaviour.

PR #46, opened 2023-01-04
Transaction list: add missing icon 

buberdds:
- add missing from/to icon
- had to lock col width for wide screens
- update address trim based on new mockups

Screenshot 2023-01-04 at 11 08 38


lubej (approved):
LGTM

PR #48, opened 2023-01-04
Add home page

lubej:
Add desktop version of the home page

https://www.loom.com/share/2c9f5779f78d4d19bd741f1344462ba3


src/styles/theme/defaultTheme.ts
```diff
@@ -336,5 +375,61 @@ export const defaultTheme = createTheme({
...
+      styleOverrides: {
+        root: {
+          padding: 0,
+        },
+        input: {
+          padding: '0 !important',
+          backgroundColor: COLORS.white,
+          color: COLORS.grayDark,
+          '&::placeholder': {
+            color: 'rgba(49,67,90,0.7)',
```
buberdds, line 388: let's use hex `#647080` and add it to COLORS

src/styles/theme/defaultTheme.ts
```diff
@@ -336,5 +375,61 @@ export const defaultTheme = createTheme({
...
+      styleOverrides: {
+        root: {
+          padding: 0,
+        },
+        notchedOutline: {
+          border: 'none',
+        },
+      },
+    },
+    MuiTextField: {
```
buberdds, line 406: Search input is highly customised and I am not sure adjusting "generic" inputs will work long term, but we will see.

src/styles/theme/defaultTheme.ts
```diff
@@ -336,5 +375,61 @@ export const defaultTheme = createTheme({
         noSsr: true,
       },
     },
+    MuiInputBase: {
+      styleOverrides: {
+        root: {
+          padding: 0,
+        },
+        input: {
+          padding: '0 !important',
```
buberdds, line 384: why we need important ?

src/app/pages/HomePage/ParaTimeSelector/index.tsx
```diff
@@ -0,0 +1,123 @@
+import React, { FC, memo, useState } from 'react'
```
buberdds, line 1: no need to import React

src/app/pages/HomePage/ParaTimeSelector/index.tsx
```diff
@@ -0,0 +1,123 @@
...
+          >
+            {t('home.zoomOutBtnText')}
+          </ZoomOutBtn>
+        </ZoomOutBtnFade>
+        {ParaTimeSelectorUtils.showExploreBtn(step) && (
+          <ExploreBtn
+            color="secondary"
+            variant="contained"
+            onClick={onExploreClick}
+            aria-label={t('home.exploreBtnText') ?? undefined}
```
buberdds, line 113: remove undef?

buberdds (approved):
ui looks really cool

src/app/components/Search/index.tsx
```diff
@@ -0,0 +1,81 @@
+import React, { FC, FocusEventHandler, FormEvent, memo, useRef, useState } from 'react'
```
buberdds, line 1: kill React and FocusEventHandler

src/app/pages/HomePage/ParaTimeSelector/types.ts
```diff
@@ -0,0 +1,14 @@
+export enum ParaTimeSelectorStep {
+  /**
+   * Animation in progress
+   */
+  ANIMATING,
```
buberdds, line 5: at some point we will have to unify enums, capitalize vs camelcase, using os string enums etc but we don;t have code standards setup yet

src/styles/theme/defaultTheme.ts
lubej: Right, I just put it into theme, so it's easier to refactor in the future. But agree that it is not good long term solution.

src/styles/theme/defaultTheme.ts
lubej: Added the opacity to focus as a solution.

lubej: Actually that is not a solution. Will modify.

src/app/pages/HomePage/ParaTimeSelector/index.tsx
lukaw3d: I don't know how this is possible:
```tsx
  const a1 = t('home.exploreBtnText')
    // hover:
    const t: TFunction <"home.exploreBtnText", string, StringMap>(
      key: "home.exploreBtnText" | "home.exploreBtnText"[]
    ) => string (+11 overloads)
```
```tsx
  const b1 = <div>{t('home.exploreBtnText')}</div>
    // hover:
    const t: TFunction<"home.exploreBtnText", DefaultTFuncReturn, StringMap>(
      key: "home.exploreBtnText" | "home.exploreBtnText"[]
    ) => DefaultTFuncReturn (+11 overloads)
```

This isn't nullable
```tsx
  const a1 = t('home.exploreBtnText')
  const a2 = <div aria-label={a1}></div>
    // hover:
    const a1: string
```

Even this isn't nullable
```tsx
  const c1 = <>{t('home.exploreBtnText')}</>
    // hover:
    const t: TFunction<"home.exploreBtnText", string, StringMap>(
      key: "home.exploreBtnText" | "home.exploreBtnText"[]
    ) => string (+11 overloads)
```

src/app/components/PageLayout/Footer.tsx
```diff
@@ -4,11 +4,13 @@ import Box from '@mui/material/Box'
 import Typography from '@mui/material/Typography'
 import useMediaQuery from '@mui/material/useMediaQuery'
 import { useTheme } from '@mui/material/styles'
+import { useConstant } from '../../hooks/useConstant'
 
 export const Footer: FC = () => {
   const { t } = useTranslation()
   const theme = useTheme()
   const isMobile = useMediaQuery(theme.breakpoints.down('sm'))
+  const currentYear = useConstant(() => new Date().getFullYear())
```
lukaw3d, line 13: I oppose to automated year-in-footer updating

But it's meaningless eitherway

src/styles/theme/defaultTheme.ts
```diff
@@ -336,5 +375,60 @@ export const defaultTheme = createTheme({
...
+        },
+        input: {
+          backgroundColor: COLORS.white,
+          color: COLORS.grayDark,
+          '&::placeholder': {
+            color: COLORS.grayDark07A,
+          },
+        },
+        adornedStart: ({ theme }) => ({
+          padding: theme.spacing(4),
```
lukaw3d, line 391: I don't think this is working

src/app/pages/HomePage/index.tsx
```diff
@@ -1,18 +1,86 @@
...
+  backgroundColor: COLORS.brandDark,
+  '&::before': {
+    content: '" "',
+    position: 'absolute',
+    inset: 0,
+    backgroundImage: `url("${background}")`,
+    backgroundSize: 'cover',
+    backgroundPosition: 'center',
+    backgroundRepeat: 'no-repeat',
+  },
```
lukaw3d, line 28: How come it's not just
```suggestion
  backgroundImage: `url("${background}")`,
  backgroundSize: 'cover',
  backgroundPosition: 'center',
  backgroundRepeat: 'no-repeat',
```

lukaw3d: Ah, neither works well over the fold
image

src/app/pages/HomePage/ParaTimeSelector/index.tsx
lubej: The type it defaults to is 
```
export type DefaultTFuncReturn = string | (TypeOptions['returnNull'] extends true ? null : never);
```

But you are right, not defining it inline, infers the correct type.

src/app/utils/route-utils.ts
```diff
@@ -0,0 +1,12 @@
+import { ParaTime } from '../../config'
+
+export abstract class RouteUtils {
```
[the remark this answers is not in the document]
lubej: To infer usage. `abstract` is here so that utils are not used as `new RouteUtils` - creating an instance of the class(which is not necessary), but rather inferring that it should be used as `RouteUtils.<static-method>`

src/styles/theme/colors.ts
```diff
@@ -18,8 +18,10 @@ export const COLORS: { [colorName: string]: string } = {
   darkSlateBlue: '#3f4284',
   eucalyptus: '#4cd4a9',
   grayDark: '#31435a',
+  grayDark07A: '#31435ab3',
```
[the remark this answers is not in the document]
lubej: To suggest 0.7 Alpha channel.

src/app/pages/HomePage/index.tsx
lubej: The reason for having background as a element on top, is because so that opacity of the background is correctly displayed. Did add min-height for now. But will deal with responsiveness in mobile PR.

PR #50, opened 2023-01-05
Add a comment for using staging API

lukaw3d:


csillag (approved):
Yes, that makes it easier.

PR #53, opened 2023-01-05
Consistently use @mui/material/styles to fix TertiaryButton type

lukaw3d:
Fixed @ts-expect-error in TertiaryButton and replaced all @mui/system references with @mui/material/styles.

This also significantly speeds up typescript in my editor in src/app/components/Select/index.tsx.


lubej (approved):
Much cleaner, thanks for doing this

src/app/components/Select/index.tsx
```diff
@@ -84,15 +80,12 @@ const TertiaryButton = forwardRef(
     {
       children,
       ...restProps
-    }: PropsWithChildren<
-      SlotComponentProps<'button', SelectUnstyledComponentsPropsOverrides, SelectUnstyledOwnerState<{}>>
-    >,
+    }: ButtonProps,
```
lukaw3d, line 83: Oops, this is less accurate when passed into SelectUnstyled slots

PR #56, opened 2023-01-06
Tweak mobile header and sub page responsiveness

buberdds:
Screenshot 2023-01-06 at 10 29 25
vs
Screenshot 2023-01-06 at 10 29 42


src/app/components/PageLayout/Header.tsx
```diff
@@ -1,16 +1,21 @@
...
 export const Header: FC = () => {
+  const theme = useTheme()
+  const isMobile = useMediaQuery(theme.breakpoints.down('sm'))
+
   return (
     <header>
-      <Grid container sx={{ px: 6, pb: 5 }}>
-        <Grid xs={12} sx={{ pb: '50px' }}>
+      <Grid container sx={{ px: isMobile ? 4 : 6, pb: isMobile ? 4 : 5 }}>
+        <Grid md={12} xs={6} sx={{ pb: isMobile ? 0 : '50px' }}>
```
lubej, line 15: This that this would be better if it would be moved to styled component. Can be left for now, we can make a clean refactor once.

src/app/components/Shape/index.tsx
```diff
@@ -0,0 +1,15 @@
+import Box, { BoxProps } from '@mui/material/Box'
```
lubej, line 1: Bounding directory is named wrong - should be Circle.

lubej (commented):
LGTM

src/app/components/PageLayout/Header.tsx
lubej: I mean `isMobile ? 4 : 6` stuff, which makes it hard to read.

buberdds: good idea, another is to split components Header vs MobileHeader etc to avoid all these crap. I will refactor this after filling "Search Placeholder" gap.

lubej: Depends on how much diff there is between the 2. As having 2 separate components would just make our bundle increase. And you need an `if` somewhere anyway to distinguish between them.

lubej (approved):

PR #57, opened 2023-01-06
Mark new items in BlocksPage

csillag:
This is pretty much the same thing we already have for transactions, implemented here: https://github.com/oasisprotocol/explorer/commit/82a05855392e657d3695378844d0cf687a1b3286

For some reason, currently it's impossible to test this, because for some reason,  we only get 9 month old emerald blocks from the backend, when we ask for the latest block.

### DIFF
--- a/src/app/pages/BlocksPage/index.tsx
+++ b/src/app/pages/BlocksPage/index.tsx
@@ -1,12 +1,14 @@
 import { FC } from 'react'
 import { useTranslation } from 'react-i18next'
+import { UseQueryResult } from '@tanstack/react-query'
+import { AxiosResponse } from 'axios'
 import Divider from '@mui/material/Divider'
 import useMediaQuery from '@mui/material/useMediaQuery'
 import { useTheme } from '@mui/material/styles'
 import { PageLayout } from '../../components/PageLayout'
 import { SubPageCard } from '../../components/SubPageCard'
 import { useGetEmeraldBlocks } from '../../../oasis-indexer/api'
-import { Blocks } from '../../components/Blocks'
+import { Blocks, TableRuntimeBlockList } from '../../components/Blocks'
 import { NUMBER_OF_ITEMS_ON_SEPARATE_PAGE, REFETCH_INTERVAL } from '../../config'
 
 const PAGE_SIZE = NUMBER_OF_ITEMS_ON_SEPARATE_PAGE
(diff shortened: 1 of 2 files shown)


src/app/pages/BlocksPage/index.tsx
```diff
@@ -19,7 +19,30 @@ export const BlocksPage: FC = () => {
...
+          if (!previousState) {
+            return nextState
+          }
+          return {
+            ...nextState,
+            data: {
+              blocks: nextState.data?.blocks?.map(block => {
+                const markAsNew = !previousState.data?.blocks?.some(
+                  prevBlock => prevBlock.round === block.round,
+                )
```
lukaw3d, line 35: I'd already optimize this to O(n)

src/app/pages/BlocksPage/index.tsx
```diff
@@ -19,7 +19,30 @@ export const BlocksPage: FC = () => {
...
+                return {
+                  ...block,
+                  markAsNew,
```
lukaw3d, line 38: Unfortunately return type isn't updated :/

src/app/pages/BlocksPage/index.tsx
csillag: OK I did that.

src/app/pages/BlocksPage/index.tsx
csillag: OK I did that, too. (At the query definition itself, because if I do it on this function, the query type is not inferred correctly.)

src/app/pages/BlocksPage/index.tsx
```diff
@@ -15,11 +17,34 @@ export const BlocksPage: FC = () => {
...
+    {
+      query: {
+        refetchInterval: REFETCH_INTERVAL,
+        structuralSharing: (previousState, nextState) => {
+          if (!previousState) {
+            return nextState
+          }
+          const oldBlocks = previousState.data?.blocks || []
+          const oldBlockIds = new Set<number>()
+          oldBlocks.forEach(block => oldBlockIds.add(block.round!))
```
lukaw3d, line 33: should be equal to
```suggestion
          const oldBlockIds = new Set(previousState.data?.blocks?.map(block => block.round!))
```

lukaw3d (approved):

PR #58, opened 2023-01-06
Use stricter types in colors

lukaw3d:
Previously `COLORS.aaaaa` was allowed

### DIFF
--- a/src/stories/Buttons.stories.tsx
+++ b/src/stories/Buttons.stories.tsx
@@ -6,7 +6,7 @@ import { ButtonTypeMap } from '@mui/material/Button/Button'
 export default {
   title: 'Example/Buttons',
   component: Button,
-} as ComponentMeta<typeof Button>
+} satisfies ComponentMeta<typeof Button>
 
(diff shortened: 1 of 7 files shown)


src/stories/Buttons.stories.tsx
```diff
@@ -42,9 +42,9 @@ Tertiary.args = {
   color: 'tertiary',
   size: 'small',
 }
-Primary.argTypes = {
+Tertiary.argTypes = {
   size: {
     control: 'select',
-    options: ['small', 'medium', 'large'] as unknown as ButtonTypeMap['props']['size'],
```
lukaw3d, line 48: I assume it was meant to be 
`options: ['small', 'medium', 'large'] as ButtonTypeMap['props']['size'][],`
to get code-completion inside the array.

`satisfies` gives us code-completion, but also prevents typos.

lubej: Awesome, wasn't aware of `satisfies`.

lubej (approved):
Looks amazing 👍

PR #59, opened 2023-01-07
Fix text alignment in select options

lukaw3d:
| Before | After |
| --- | --- |
|  localhost_1234_emerald (2) | localhost_1234_emerald  |

### DIFF
--- a/src/app/components/Select/index.tsx
+++ b/src/app/components/Select/index.tsx
@@ -15,7 +15,7 @@ import { useTranslation } from 'react-i18next'
 const StyledButton = styled(Button)(({ theme }) => ({
   height: '36px',
   minWidth: '135px',
-  padding: `${theme.spacing(4)} ${theme.spacing(4)}`,
+  padding: `0 ${theme.spacing(4)}`,
   borderRadius: '12px',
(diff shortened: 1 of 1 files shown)


src/app/components/Select/index.tsx
```diff
@@ -52,7 +51,6 @@ const StyledListbox = styled('ul')(({ theme }) => ({
 const StyledOption = styled(OptionUnstyled)(({ theme }) => ({
   boxSizing: 'border-box',
   listStyle: 'none',
-  height: '36px',
   padding: `${theme.spacing(3)} ${theme.spacing(4)}`,
```
lukaw3d, line 54: (if it's meant to be smaller we should adjust padding instead)

lubej: Don't particularly like the solution, as the Button and the Option have different heights now. https://github.com/oasisprotocol/explorer/pull/72

lukaw3d: Ah, if they should have the same height I'd keep height, remove vertical padding, and center text. Updated PR

lubej (approved):

PR #62, opened 2023-01-09
Account details page

buberdds:
init account details page

- jazz icon stuff is taken from web wallet [PR](https://github.com/oasisprotocol/oasis-wallet-web/pull/936)
- `stream-browserify` is added by parcel to compile oasis client

Screenshot 2023-01-10 at 13 56 33

### DIFF
--- a/src/app/components/Account/index.tsx
+++ b/src/app/components/Account/index.tsx
@@ -5,6 +5,7 @@ import { useTheme } from '@mui/material/styles'
 import { StyledDescriptionList } from '../../components/StyledDescriptionList'
 import { CopyToClipboard } from '../../components/CopyToClipboard'
 import { JazzIcon } from '../../components/JazzIcon'
+import { trimLongString } from '../../utils/trimLongString'
 import { Account as ConsensusAccount } from '../../../oasis-indexer/api'
 
 export const addressToNumber = (address: string) => {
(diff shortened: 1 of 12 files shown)


src/app/utils/account.ts
```diff
@@ -0,0 +1,9 @@
+import { staking } from '@oasisprotocol/client'
+
+export const addressToNumber = (address: string) => {
```
lukaw3d, line 3: This is specific to creating an account icon, so I would move it out of general utils.
The returned number is otherwise meaningless/arbitrary, not any kind of account number, and "seed" variable name comes from "a seed for randomization of icon shapes"

src/app/utils/trimLongString.ts
```diff
@@ -9,3 +9,5 @@ export function trimLongString(value: string, trimStart = 10, trimEnd = 8, ellip
 
   return `${value.slice(0, trimStart)}${ellipsis}${trimEnd ? value.slice(-trimEnd) : ''}`
 }
+
+export const defaultTrim = (value: string) => trimLongString(value, 6, 6)
```
lukaw3d, line 13: The default trim is `trimStart = 10, trimEnd = 8, ellipsis = '...'`

This is an alternative default? :smile:

src/app/components/Account/index.tsx
```diff
@@ -0,0 +1,41 @@
+import { FC } from 'react'
+import { staking } from '@oasisprotocol/client'
```
lukaw3d, line 2: (depending on future uses: this is a large dependency for a small utility here. Potentially something to optimize in the future)

lukaw3d (approved):

PR #67, opened 2023-01-10
Fix rotated ArrowIcon width

lukaw3d:
When ArrowDirection.RIGHT it still takes up 8x13 instead of 13x8. Let's just make it square?

### DIFF
--- a/src/app/icons/ArrowIcon.tsx
+++ b/src/app/icons/ArrowIcon.tsx
@@ -22,7 +22,7 @@ const ArrowIcon: FC<ArrowProps> = ({ arrowDirection = ArrowDirection.UP }) => {
   return (
     <ArrowSvgIcon
       arrowDirection={arrowDirection}
-      width="8"
+      width="13"
       height="13"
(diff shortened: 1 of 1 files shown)


src/app/icons/ArrowIcon.tsx
```diff
@@ -22,7 +22,7 @@ const ArrowIcon: FC<ArrowProps> = ({ arrowDirection = ArrowDirection.UP }) => {
   return (
     <ArrowSvgIcon
       arrowDirection={arrowDirection}
-      width="8"
+      width="13"
```
buberdds, line 25: Is icon positioned correctly for all directions when svg `viewBox` is not in sync with this change?

buberdds (approved):

src/app/icons/ArrowIcon.tsx
lukaw3d: Yep, at least on Chrome and Firefox

| Before | After |
| --- | --- |
|  localhost_1234_ (1) |  localhost_1234_  |

PR #69, opened 2023-01-11
Replace custom ArrowIcon with MUI icons

buberdds:
- custom icons are licensed so we will use MUI icons in this project. mockups will be updated in the future.
- using north/south icons in badge instead of up/down because they look more similar to current designs

Screenshot 2023-01-11 at 13 59 18


src/app/components/PercentageGain/percentage-gain-utils.ts
```diff
@@ -1,5 +0,0 @@
-import { Gain } from './types'
```
lukaw3d, line 1: Leftover file

lukaw3d (approved):

PR #71, opened 2023-01-12
Fix react-query's structuralSharing type

lukaw3d:


src/oasis-indexer/api.ts
```diff
@@ -17,42 +16,48 @@ function fromBaseUnits(valueInBaseUnits: string, decimals: number): string {
...
-        data: {
-          ...result.data.data,
-          transactions: result.data.data.transactions?.map(tx => {
+  const result = generated.useGetEmeraldTransactions(
+    { ...params, offset },
+    {
+      ...options,
+      axios: {
+        ...options?.axios,
+        transformResponse: [
```
lukaw3d, line 36: Converted parsing into axios transform instead of parsing after query returns:
- i wanted structural sharing to have the parsed data
- i tried using react-query's select function, but it makes a bunch of weirdness in structuralSharing

buberdds (approved):
optional: move away axios imports from components by moving this stuff to react query api wrappers

PR #75, opened 2023-01-13
Fix TertiaryButton type again

lukaw3d:
Previous change https://github.com/oasisprotocol/explorer/pull/53/files#r1069961535 was inaccurate

This now correctly exposes e.g. `restProps.ownerState.open` that we could use for switching icons, instead of using css

### DIFF
--- a/src/app/components/Select/index.tsx
+++ b/src/app/components/Select/index.tsx
@@ -1,12 +1,18 @@
 import * as React from 'react'
-import SelectUnstyled, { SelectUnstyledProps, selectUnstyledClasses } from '@mui/base/SelectUnstyled'
+import SelectUnstyled, {
+  SelectUnstyledProps,
+  selectUnstyledClasses,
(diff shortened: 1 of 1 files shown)


lubej (approved):

src/app/components/Select/index.tsx
```diff
@@ -82,13 +87,17 @@ const TertiaryButton = forwardRef(
     {
       children,
       ...restProps
-    }: ButtonProps,
+    }: PropsWithChildren<
+      // Extracted type from SelectUnstyledProps<{}>['slotProps']
+      Partial<React.ComponentPropsWithRef<'button'>> &
+        SelectUnstyledComponentsPropsOverrides & { ownerState: SelectUnstyledOwnerState<{}> }
```
buberdds, line 93: can't we use `SelectUnstyledRootSlotProps` type instead ?

PR #76, opened 2023-01-16
Account details: fiat value

buberdds:
The plan is to use Coin Gecko API to fetch ROSE price for MVP. Later, according to Nikola we will fetch ROSE fiat value from our API. 

Screenshot 2023-01-17 at 10 06 45

### DIFF
--- a/src/app/components/Account/index.tsx
+++ b/src/app/components/Account/index.tsx
@@ -1,4 +1,5 @@
 import { FC } from 'react'
+import { useTranslation } from 'react-i18next'
 import { staking } from '@oasisprotocol/client'
 import useMediaQuery from '@mui/material/useMediaQuery'
 import { useTheme } from '@mui/material/styles'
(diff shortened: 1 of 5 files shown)


src/app/components/Account/index.tsx
```diff
@@ -19,11 +20,13 @@ export const addressToNumber = (address: string) => {
 type AccountProps = {
   // switch to Emerald when endpoint is ready
```
csillag, line 21: This is not something we have discussed earlier, but I prefer adding `TODO` to code comments that explains something that we need to do later.

src/app/components/Account/index.tsx
```diff
@@ -19,11 +20,13 @@ export const addressToNumber = (address: string) => {
...
   account: ConsensusAccount
+  roseFiatValue?: number
 }
 
-export const Account: FC<AccountProps> = ({ account }) => {
+export const Account: FC<AccountProps> = ({ account, roseFiatValue }) => {
   const theme = useTheme()
   const isMobile = useMediaQuery(theme.breakpoints.down('sm'))
+  const mockedRuntimeBalance = 1000 // replace with real value when API is ready
```
csillag, line 29: Same thing with TODO here.

csillag (approved):
Reviewed. LGTM.

PR #79, opened 2023-01-17
Lift useSearchParams out of API and TablePagination into pages

lukaw3d:
This allows for multiple paginated tables on the same view without conflicting

### DIFF
--- a/src/app/components/Table/useSearchParamsPagination.ts
+++ b/src/app/components/Table/useSearchParamsPagination.ts
@@ -1,0 +1,18 @@
+import { To, useSearchParams } from 'react-router-dom'
+
+export function useSearchParamsPagination(paramName: string) {
+  const [searchParams] = useSearchParams()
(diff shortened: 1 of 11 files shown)


src/app/components/Table/useSearchParamsPagination.ts
```diff
@@ -0,0 +1,18 @@
+import { To, useSearchParams } from 'react-router-dom'
+
+export function useSearchParamsPagination(paramName: string) {
+  const [searchParams] = useSearchParams()
+  const selectedPage = parseInt(searchParams.get(paramName) ?? '1', 10)
```
buberdds, line 5: not sure it's worth adding a guard against NaN

buberdds (approved):

src/app/components/Table/useSearchParamsPagination.ts
lukaw3d: ah - what do you think of
```ts
if (isNaN(selectedPage)) throw new Error('400 Bad Request')
```

PR #81, opened 2023-01-17
Switch to latest API yaml definition

lukaw3d:
Made on top of https://github.com/oasisprotocol/explorer/pull/79 and https://github.com/oasisprotocol/explorer/pull/80

### DIFF
--- a/src/app/pages/DashboardPage/DurationSelect/index.tsx
+++ b/src/app/pages/DashboardPage/DurationSelect/index.tsx
@@ -1,7 +1,6 @@
 import { FC, memo } from 'react'
 import { Select, SelectOptionBase } from '../../../components/Select'
 import { ChartDuration } from '../../../utils/chart-utils'
-import { useConstant } from '../../../hooks/useConstant'
 import { useTranslation } from 'react-i18next'
 
 interface DurationOption extends SelectOptionBase {
@@ -17,7 +16,7 @@ interface DurationSelectProps {
 const DurationSelectCmp: FC<DurationSelectProps> = ({ defaultValue = ChartDuration.TODAY, handleChange }) => {
   const { t } = useTranslation()
 
-  const options = useConstant<DurationOption[]>(() => [
+  const options = [
     {
       label: t('chartDuration.today'),
       value: ChartDuration.TODAY,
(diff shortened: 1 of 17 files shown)


buberdds (approved):

src/app/pages/DashboardPage/DurationSelect/index.tsx
```diff
@@ -17,7 +16,7 @@ interface DurationSelectProps {
 const DurationSelectCmp: FC<DurationSelectProps> = ({ defaultValue = ChartDuration.TODAY, handleChange }) => {
   const { t } = useTranslation()
 
-  const options = useConstant<DurationOption[]>(() => [
```
lubej, line 20: `useConstant` is actually a performance optimization. As it uses ref behind the scenes. The current implementation will create a new instance of `options` every time rerender happens. To prevent this from happening you can define the constant outside of the Component.

src/app/pages/DashboardPage/TransactionsChartCard.tsx
```diff
@@ -22,27 +22,26 @@ const TransactionsChartCardCmp: FC<TransactionsChartCardProps> = ({ chartDuratio
...
   return (
     <Card sx={{ p: 0 }}>
       <CardHeader component="h5" title={t('transactionsTpsChart.header')} sx={{ pb: 0, pl: 4, pt: 4 }} />
       <CardContent sx={{ pt: 4 }}>
-        {lineChartData !== null && (
-          <LineChart<TpsCheckpoint>
-            dataKey="tx_volume"
+        {lineChartData && (
+          <LineChart
+            dataKey="volume_per_second"
```
lubej, line 44: Would be nice to pass in the type parameter of LineChart. For type safety on `dataKey` & `data` props.

lubej (approved):
LGTM, with couple of nits

src/app/pages/DashboardPage/TransactionsChartCard.tsx
lukaw3d: It's inferred and still typesafe tho

src/app/pages/DashboardPage/DurationSelect/index.tsx
lukaw3d: I would say it's a premature optimization, and in this case, if we had a way to switch the language, it would break translations

PR #90, opened 2023-01-24
Init Tokens sub pages

buberdds:
- adds `RouterTabs` that can be used in TX detail page as well 
- split account details into sub components and sub routes

I will ask Don to simplify tabs a little bit, because imo MUI tabs customisation is limited, or pixel perfect impl will need tons of additional overrides.

Screenshot 2023-01-24 at 13 15 15

### DIFF
--- a/src/app/components/RouterTabs/index.tsx
+++ b/src/app/components/RouterTabs/index.tsx
@@ -1,0 +1,42 @@
+import { FC } from 'react'
+import { useLocation, matchPath, Link as RouterLink, Outlet } from 'react-router-dom'
+import Tabs from '@mui/material/Tabs'
+import Tab from '@mui/material/Tab'
+
+type RouterTabsProps = {
+  tabs: {
+    label: string
+    to: string
+  }[]
+}
+
+function useRouteMatch(patterns: string[]) {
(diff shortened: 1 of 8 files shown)


src/app/components/RouterTabs/index.tsx
```diff
@@ -0,0 +1,42 @@
...
+    }
+  }
+
+  return null
+}
+
+export const RouterTabs: FC<RouterTabsProps> = ({ tabs }) => {
+  const tabsPaths = tabs.flatMap(num => num.to)
+  const routeMatch = useRouteMatch(tabsPaths)
+  const currentTab = routeMatch?.pattern?.path
```
lukaw3d, line 30: You can avoid this matching with

```diff
- component={RouterLink} to={tab.to}
+ component={NavLink} to={tab.to} end

- '&.Mui-selected': {
+ '&.active': {
```

lukaw3d (approved):

src/app/components/RouterTabs/index.tsx
buberdds: MUI component needs `value` prop to work correctly

PR #96, opened 2023-01-26
Add yarn time and build size to CI step summary

lukaw3d:


.github/workflows/build-test.yaml
```diff
@@ -15,8 +15,18 @@ jobs:
         with:
           node-version: 18.x
           cache: 'yarn'
-      - run: yarn install --frozen-lockfile
+      - run: yarn install --frozen-lockfile | tee yarn-install-out
```
lubej, line 18: One thing that could be useful additional step here is, caching the dependencies:

Just some random snippet from the internet:
```
      - uses: actions/cache@v2
        with:
          path: '**/node_modules'
          key: ${{ runner.os }}-modules-${{ hashFiles('**/yarn.lock') }}
```

Source: https://dev.to/mattpocockuk/how-to-cache-nodemodules-in-github-actions-with-yarn-24eh

lukaw3d: We are using the new actions/setup-node@v3 so it uses actions/cache@v3
https://github.com/actions/setup-node/blob/3dbcda8bc2eb5ec6aa3fbde01feaae3236952db8/docs/advanced-usage.md#caching-packages-data

but I'm not sure why yarn still takes 40 seconds :/

lukaw3d: Attempt without cache takes 69 seconds https://github.com/oasisprotocol/explorer/actions/runs/4027056314/jobs/6922325295

lubej (approved):

PR #101, opened 2023-02-01
Derive oasis address from eth address

buberdds:
In case of eth address in url we need to derive oasis address to fetch emerald transactions or account details.  
Validation helpers will be used in other places (to show error pages or in search validation).

### DIFF
--- a/src/app/pages/AccountDetailsPage/index.tsx
+++ b/src/app/pages/AccountDetailsPage/index.tsx
@@ -1,6 +1,6 @@
 import { FC } from 'react'
 import { useTranslation } from 'react-i18next'
-import { useHref, useParams } from 'react-router-dom'
+import { useHref, useLoaderData } from 'react-router-dom'
 import CardContent from '@mui/material/CardContent'
 import Skeleton from '@mui/material/Skeleton'
 import { PageLayout } from '../../components/PageLayout'
@@ -12,7 +12,7 @@ import { useGetRosePrice } from '../../../coin-gecko/api'
 
 export const AccountDetailsPage: FC = () => {
   const { t } = useTranslation()
-  const { address } = useParams()
+  const address = useLoaderData() as string
   // TODO: switch to Emerald when API is ready
   const accountQuery = useGetConsensusAccountsAddress(address!)
   const account = accountQuery.data?.data
--- a/src/app/utils/helpers.ts
+++ b/src/app/utils/helpers.ts
@@ -1,0 +1,35 @@
+import * as oasis from '@oasisprotocol/client'
+import * as oasisRT from '@oasisprotocol/client-rt'
+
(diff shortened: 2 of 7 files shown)


src/app/utils/helpers.ts
```diff
@@ -0,0 +1,25 @@
...
+  try {
+    oasis.staking.addressFromBech32(addr)
+    return true
+  } catch (e) {
+    return false
+  }
+}
+
+export const isValidEthAddress = (hexAddress: string): boolean => {
+  return /^0x[0-9a-fA-F]{40}$/.test(hexAddress)
```
lukaw3d, line 14: Note: This doesn't validate checksum in address. I'll add a full version for search validation

src/app/hooks/useRouteParams.ts
```diff
@@ -0,0 +1,31 @@
...
+
+export const useOasisAddressFromQueryParam = () => {
+  const [oasisAddress, setOasisAddress] = useState<string | undefined>()
+  const { address } = useParams()
+
+  useEffect(() => {
+    async function validateAddress() {
+      if (address) {
+        const oasisAddress = await getOasisAddressFromParams(address)
+        setOasisAddress(oasisAddress)
```
lukaw3d, line 23: > In case of eth address in url we need to derive oasis address to fetch emerald transactions or account details

API is supposed to support querying by eth address too

If we really really need this awkward async conversion anyway: I would move it into router, so it executes before initializing components.

mitjat: FWIW, the `async` is not because this requires a network roundtrip, it's just because hashing is async. This is the root: https://github.com/oasisprotocol/oasis-sdk/blob/b37e6da699df331f5a2ac62793f8be099c68469c/client-sdk/ts-web/core/src/address.ts#L14-L14
So it's not slow, but it might be ugly :)

src/app/pages/AccountDetailsPage/index.tsx
```diff
@@ -1,4 +1,4 @@
-import { FC } from 'react'
+import { FC, useEffect, useState } from 'react'
```
buberdds, line 1: remove

src/app/utils/route-utils.ts
```diff
@@ -31,3 +33,8 @@ export abstract class RouteUtils {
     return RouteUtils.ENABLED_PARA_TIMES
   }
 }
+
+export const addressParamLoader = async ({ params }: LoaderFunctionArgs) => {
+  const address = await getOasisAddress(params.address!)
+  return address
```
lukaw3d, line 39: Perhaps it would be easier to understand if this returned `{oasisAddress, ethAddress?}`

lukaw3d (approved):

PR #102, opened 2023-02-01
Follow latest API changes

csillag:
This PR builds on top of the auto-generated #89 , and attempts to fix what is broken by the API changes.

Since I haven't investigated everything deeply, some review would be very welcome.

### DIFF
--- a/src/app/components/Account/index.tsx
+++ b/src/app/components/Account/index.tsx
@@ -10,7 +10,7 @@ import { CopyToClipboard } from '../../components/CopyToClipboard'
 import { JazzIcon } from '../../components/JazzIcon'
 import { CoinGeckoReferral } from '../../components/CoinGeckoReferral'
 import { trimLongString } from '../../utils/trimLongString'
-import { Account as ConsensusAccount } from '../../../oasis-indexer/api'
+import { FullConsensusAccount } from '../../../oasis-indexer/api'
 import { TokenPills } from './TokenPills'
 
 export const addressToNumber = (address: string) => {
@@ -29,7 +29,7 @@ const StyledBox = styled(Box)(() => ({
 
 type AccountProps = {
   // TODO: switch to Emerald when endpoint is ready
-  account: ConsensusAccount
+  account: FullConsensusAccount
   roseFiatValue?: number
 }
 
@@ -70,7 +70,7 @@ export const Account: FC<AccountProps> = ({ account, roseFiatValue }) => {
       )}
       <dt>{t('account.tokens')}</dt>
       <dd>
-        <TokenPills tokens={account.runtime_evm_balance} />
+        <TokenPills tokens={account.runtime_evm_balances} />
       </dd>
     </StyledDescriptionList>
   )
--- a/src/app/pages/DashboardPage/AverageTransactionSize.tsx
+++ b/src/app/pages/DashboardPage/AverageTransactionSize.tsx
@@ -4,12 +4,13 @@ import Card from '@mui/material/Card'
 import CardHeader from '@mui/material/CardHeader'
 import CardContent from '@mui/material/CardContent'
 import { LineChart } from '../../components/charts/LineChart'
-import { useGetConsensusStatsTxVolume } from '../../../oasis-indexer/api'
+import { Layer, useGetLayerStatsTxVolume } from '../../../oasis-indexer/api'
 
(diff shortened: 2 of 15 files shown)


csillag:
Updated to cover API usage in the recently merged offline indicator PR.

src/app/components/Account/index.tsx
```diff
@@ -70,7 +70,7 @@ export const Account: FC<AccountProps> = ({ account, roseFiatValue }) => {
       )}
       <dt>{t('account.tokens')}</dt>
       <dd>
-        <TokenPills tokens={account.runtime_evm_balance} />
+        <TokenPills tokens={account.runtime_evm_balances || []} />
```
lukaw3d, line 73: This should be non-nullable when fetching a single account

src/app/pages/DashboardPage/AverageTransactionSize.tsx
```diff
@@ -4,12 +4,19 @@ import Card from '@mui/material/Card'
...
   const { t } = useTranslation()
-  // TODO: Replace with real stats when available
-  const dailyVolumeQuery = useGetConsensusStatsTxVolume()
+  const dailyVolumeQuery = useGetLayerStatsTxVolume(
+    Layer.consensus, // TODO: which layer do we want here?
+    {
+      // TODO: Do we want some parameters for the stats? Like these:
+      // limit: 2,
+      // offset: 2,
+      // bucket_size_seconds: 60,
```
lukaw3d, line 17: I think we are happy with current defaults: `limit: 100, bucket_size_seconds: 24*60*60`

src/app/pages/DashboardPage/AverageTransactionSize.tsx
```diff
@@ -4,12 +4,19 @@ import Card from '@mui/material/Card'
...
 import { LineChart } from '../../components/charts/LineChart'
-import { useGetConsensusStatsTxVolume } from '../../../oasis-indexer/api'
+import { Layer, useGetLayerStatsTxVolume } from '../../../oasis-indexer/api'
 
 export const AverageTransactionSize: FC = () => {
   const { t } = useTranslation()
-  // TODO: Replace with real stats when available
-  const dailyVolumeQuery = useGetConsensusStatsTxVolume()
+  const dailyVolumeQuery = useGetLayerStatsTxVolume(
+    Layer.consensus, // TODO: which layer do we want here?
```
lukaw3d, line 12: We want emerald, but it doesn't work, so consensus + // TODO: switch to emerald

src/app/pages/DashboardPage/TransactionsChartCard.tsx
```diff
@@ -17,17 +17,21 @@ const TransactionsChartCardCmp: FC<TransactionsChartCardProps> = ({ chartDuratio
   const theme = useTheme()
   const isMobile = useMediaQuery(theme.breakpoints.down('sm'))
   const statsParams = durationToQueryParams[chartDuration]
-  // TODO: Replace with Emerald stats
-  const { data } = useGetConsensusStatsTxVolume(statsParams)
+  const { data } = useGetLayerStatsTxVolume(Layer.emerald, {
+    // TODO: Do we want some parameters for the stats? Like these:
```
lukaw3d, line 21: we want statsParams

src/app/components/Account/index.tsx
csillag: > This should be non-nullable when fetching a single account

Indeed; the API docs says this:

> NOTE 1: This field is omitted for efficiency when listing multiple accounts.

So I added a wrapper that adjusts the return type of the single account query, so that we will always know that this can never be null.

src/app/pages/DashboardPage/AverageTransactionSize.tsx
csillag: ok

src/app/pages/DashboardPage/AverageTransactionSize.tsx
csillag: ok

src/app/pages/DashboardPage/TransactionsChartCard.tsx
csillag: ok

csillag:
I implemented all the feedback, rebased to latest master, fixed a few more things, and added a simplification (around the Token type).

PR #103, opened 2023-02-02
Show more Emerald TX details

csillag:
This requires the latest API, so is built on top of #102.

(It needs to be rebased after that one is merged.)

Visuals as currently implemented:

image

Latest design for desktop:

image

Implementation and design for mobile:

image

image


csillag:
Some more stuff is missing, but this is already functional and is clearly an improvement, so we could merge this, and continue separately.

src/app/pages/TransactionDetailPage/index.tsx
```diff
@@ -32,39 +107,59 @@ export const TransactionDetailPage: FC = () => {
...
             </dd>
 
+            <dt>{t('common.timestamp')}</dt>
+            <dd>{formattedTimestamp}</dd>
+
             <dt>{t('common.from')}</dt>
-            <dd>{transaction.sender_0}</dd>
+            <dd>
+              <AccountLink address={transaction.sender_0} paratime={ParaTime.Emerald} />
+              <CopyToClipboard value={transaction.sender_0} label={' '} />
```
lukaw3d, line 143: At some point we should refactor `label={' '}`

lukaw3d (approved):

src/app/pages/TransactionDetailPage/index.tsx
csillag: Yes, there are various possibilities here.
- What we want is displaying some potentially shortened text, potentially as a link, and a copy to clipboard icon.
- CopyToClipboard is able to display the value itself, but it doesn't support shortening the the text, or using a link.
- Supporting using shortened text is easy, but links are more complicated, because of the separation of concerns we are using between the MUI and react router. 
- So maybe it would be easiest to add the CopyToLink functionality to the (account) link widget itself....

csillag: > At some point we should refactor label={' '}

Done, see https://github.com/oasisprotocol/explorer/pull/106

PR #104, opened 2023-02-02
Evm tokens section updates

buberdds:
- hardcoded tokens are removed and we show the first three items 
- enable highlighting rows in table, scroll to clicked item

Screenshot 2023-02-02 at 13 48 47

### DIFF
--- a/src/app/components/Account/TokenPills.tsx
+++ b/src/app/components/Account/TokenPills.tsx
@@ -9,25 +9,13 @@ type TokenPillsProps = {
   tokens: Token[]
 }
 
-const prioritizedTokensSymbols = ['ROSE', 'WROSE', 'ETH']
-
 export const TokenPills: FC<TokenPillsProps> = ({ tokens }) => {
   const { t } = useTranslation()
-
-  if (!tokens) {
+  if (!tokens?.length) {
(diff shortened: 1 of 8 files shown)


src/app/components/Account/TokenPills.tsx
```diff
@@ -9,37 +11,30 @@ type TokenPillsProps = {
   tokens: Token[]
 }
 
-const prioritizedTokensSymbols = ['ROSE', 'WROSE', 'ETH']
-
 export const TokenPills: FC<TokenPillsProps> = ({ tokens }) => {
   const { t } = useTranslation()
-
   if (!tokens) {
```
lubej, line 16:
```suggestion
  if (!tokens.length) {
```

src/app/components/Account/TokenPills.tsx
```diff
@@ -9,37 +11,30 @@ type TokenPillsProps = {
...
+  const pills = tokens.slice(0, 3)
 
   return (
     <>
       {pills.map(item => (
         <Chip
+          clickable
+          color="tertiary"
+          component={Link}
+          href={`${item.token_type === 'ERC20' ? useHref('tokens/erc-20') : useHref('tokens/erc-721')}#${
```
lubej, line 28: Since `useHref` is a hook, it should not be conditionally applied.

lubej (approved):

PR #107, opened 2023-02-03
Handle invalid routes 

buberdds:
goes after https://github.com/oasisprotocol/explorer/pull/101

- not adding tx hash validation as it is WIP in search 
- mostly rip off of wallet pattern
- EmptyState should work in theory in search (empty result view)

Screenshot 2023-02-03 at 12 02 23


lubej (approved):
LGTM

PR #119, opened 2023-02-09
Handle invalid pagination parameters via explicit checking

csillag:
~**This depends on #124. (That's the first commit.) Please merge that one first, then rebase.**~

This implementation teaches the pagination hook to return an explicit invalid flag, when encountering an invalid page number.

## Analysis:
 * The drawback is that the happy path and the error path are not completely separated.
 * The benefits is that there is TypeScript-level check to make sure that we don't forget to add the error path. 
image

See #121 for an alternative, using an exception and React Router's error catching feature.


csillag:
Closing in favor of #121 .

PR #123, opened 2023-02-10
Avoid exception in data transformation when there are no transactions

csillag:
When we try load transactions where there are none, sometimes we get an exception on the console.
This one-liner change fixes that.


lukaw3d:
An example of this issue: 1922304 block has 0 transactions

https://index-staging.oasislabs.com/v1/emerald/transactions?block=1922304 returns {transactions: null}

lukaw3d (approved):

PR #126, opened 2023-02-13
Refactor expanded search field styling

lukaw3d:
- Collapsing and expanding logic is all in one place
- Clicking the search icon now passes through and focuses search field
- Reduced search icon padding

| Before | After |
| --- | --- |
| localhost_1234_ (1)  |  localhost_1234_
 |


lubej (approved):
LGTM

PR #127, opened 2023-02-13
Fix emotion's runtime error that :first-child is unsafe for SSR

lukaw3d:
Fixes e1950fc8b29c0e10539117627d835848e337cf0d from https://github.com/oasisprotocol/explorer/pull/126

`The pseudo class ":first-child" is potentially unsafe when doing server-side rendering. Try changing it to ":first-of-type".`

### DIFF
--- a/src/styles/theme/defaultTheme.ts
+++ b/src/styles/theme/defaultTheme.ts
@@ -460,9 +460,14 @@ export const defaultTheme = createTheme({
       styleOverrides: {
         root: ({ theme }) => ({
           borderRadius: '24px',
-          '.MuiInputBase-root > :first-child': {
+          '.MuiInputBase-root': {
             // Prevent first child's edges overflowing due to border-radius
-            marginLeft: theme.spacing(4),
+            '&.MuiInputBase-adornedStart .MuiInputAdornment-positionStart': {
+              marginLeft: theme.spacing(4),
+            },
+            '&:not(.MuiInputBase-adornedStart) .MuiInputBase-input': {
+              marginLeft: theme.spacing(4),
+            },
           },
           ':focus-within': {
             boxShadow: '0px 4px 50px 15px rgba(0, 0, 98, 0.54)',


src/styles/theme/defaultTheme.ts
```diff
@@ -461,9 +461,14 @@ export const defaultTheme = createTheme({
...
-          '.MuiInputBase-root > :first-child': {
+          '.MuiInputBase-root': {
             // Prevent first child's edges overflowing due to border-radius
-            marginLeft: theme.spacing(4),
+            '.MuiInputAdornment-positionStart': {
+              marginLeft: theme.spacing(4),
+            },
+            '.MuiInputBase-input:not(.MuiInputBase-inputAdornedStart)': {
+              marginLeft: theme.spacing(4),
+            },
```
csillag, line 471: I don't get this. This the new code, aren't we applying the same spacing to the start position and all the non-start positions?

src/styles/theme/defaultTheme.ts
```diff
@@ -461,9 +461,14 @@ export const defaultTheme = createTheme({
...
         root: ({ theme }) => ({
           borderRadius: '24px',
-          '.MuiInputBase-root > :first-child': {
+          '.MuiInputBase-root': {
             // Prevent first child's edges overflowing due to border-radius
-            marginLeft: theme.spacing(4),
+            '.MuiInputAdornment-positionStart': {
+              marginLeft: theme.spacing(4),
+            },
+            '.MuiInputBase-input:not(.MuiInputBase-inputAdornedStart)': {
```
buberdds, line 469: Is it possible to avoid or limit custom css selectors? I did not check this, but something like `root: ({ theme }) => ({input: {},` ?

lukaw3d: Doesn't seem so

src/styles/theme/defaultTheme.ts
lukaw3d: Refactored to make it clearer that it applies spacing to the first child:
```ts
            // Prevent first child's edges overflowing due to border-radius
            '&.MuiInputBase-adornedStart .MuiInputAdornment-positionStart': {
              marginLeft: theme.spacing(4),
            },
            '&:not(.MuiInputBase-adornedStart) .MuiInputBase-input': {
              marginLeft: theme.spacing(4),
            },
```

buberdds (approved):

PR #128, opened 2023-02-14
Refactor RouterTabs

lukaw3d:


src/app/components/RouterTabs/index.tsx
```diff
@@ -10,28 +10,13 @@ type RouterTabsProps = {
...
-
-  return null
-}
-
 export const RouterTabs: FC<RouterTabsProps> = ({ tabs }) => {
-  const tabsPaths = tabs.flatMap(num => num.to)
-  const routeMatch = useRouteMatch(tabsPaths)
-  const currentTab = routeMatch?.pattern?.path
+  const { pathname } = useLocation()
+  const currentTab = tabs.find(tab => matchPath(tab.to, pathname))
```
csillag, line 15: Does this take card of the reason why flattening was used before? (in `tabs.flatMap`).

lukaw3d: Looked like a mistake to me

csillag: Maybe, maybe not. @buberdds wrote the previous version of the code, maybe he can explain the reason?

buberdds (approved):

src/app/components/RouterTabs/index.tsx
buberdds: refactor looks Ok

PR #129, opened 2023-02-14
Use type as AppError fallback message

lukaw3d:


csillag (approved):
Well it can't hurt. LGTM.

PR #134, opened 2023-02-14
Refactor transaction details and added "block not found" exception

lukaw3d:


src/app/pages/TransactionDetailPage/index.tsx
```diff
@@ -54,19 +34,18 @@ function compareTxAge(a: RuntimeTransaction, b: RuntimeTransaction) {
...
     return {
       wantedTransaction: transactions[0],
     }
   } else {
-    let wantedTransaction = transactions.find(transaction => transaction.success)
-    if (!wantedTransaction) {
-      wantedTransaction = transactions.sort(compareTxAge).at(-1)
-    }
+    const successfulOne = transactions.find(transaction => transaction.success)
+    const latestOne = transactions.sort((a, b) => b.round - a.round)[0]
```
csillag, line 45: Ahh, I was not aware that compare functions are allowed to return values different from -1, 0 and +1. That makes it simplier.

csillag (approved):
Reviewed. LGTM.

PR #137, opened 2023-02-15
Account details: add missing rows 

buberdds:
We still don't have data in API for three rows, but all of them are required for MVP anyway.

### DIFF
--- a/src/app/components/Account/index.tsx
+++ b/src/app/components/Account/index.tsx
@@ -11,7 +11,11 @@ import { CopyToClipboard } from '../../components/CopyToClipboard'
 import { JazzIcon } from '../../components/JazzIcon'
 import { CoinGeckoReferral } from '../../components/CoinGeckoReferral'
 import { trimLongString } from '../../utils/trimLongString'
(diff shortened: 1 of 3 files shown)


csillag (approved):
Reviewed and tested. LGTM.

src/app/components/Account/index.tsx
```diff
@@ -56,13 +65,19 @@ export const Account: FC<AccountProps> = ({ account, roseFiatValue }) => {
...
+      <dt>{t('common.balance')}</dt>
+      <dd>{t('common.valueInRose', { value: balance })}</dd>
       {roseFiatValue && (
         <>
           <dt>{t('common.fiatValue')}</dt>
           <dd>
             <StyledBox>
               {t('common.fiatValueInUSD', {
-                value: mockedRuntimeBalance * roseFiatValue,
+                value: parseFloat(balance) * roseFiatValue,
```
lukaw3d, line 80: I think all calculations should be done in BigNumber
(cc @csillag, in case you're adding `fee / gas_limit`)

PR #138, opened 2023-02-15
Add eslint

csillag:
This is now ready for review. Some testing would be welcome.


csillag:
@lubej , there is one unresolved warning for the ParaTime selector help screen:

> **src/app/pages/HomePage/ParaTimeSelector/HelpScreen/index.tsx**
>  77:19  warning  The ref value 'swiperElRef.current' will likely have changed by the time this effect cleanup function runs. If this ref points to a node rendered by React, copy 'swiperElRef.current' to a variable inside the effect, and use that variable in the cleanup function  react-hooks/exhaustive-deps

Could you please take a look at that? Thanks

lubej:
> @lubej , there is one unresolved warning for the ParaTime selector help screen:
> 
> > **src/app/pages/HomePage/ParaTimeSelector/HelpScreen/index.tsx**
> > 77:19  warning  The ref value 'swiperElRef.current' will likely have changed by the time this effect cleanup function runs. If this ref points to a node rendered by React, copy 'swiperElRef.current' to a variable inside the effect, and use that variable in the cleanup function  react-hooks/exhaustive-deps
> 
> Could you please take a look at that? Thanks

Right, fixed in https://github.com/oasisprotocol/explorer/pull/142

package.json
```diff
@@ -12,8 +12,9 @@
     "clean": "rm -rf build/ .parcel-cache",
     "start": "parcel --dist-dir build-dev --no-cache",
     "checkTs": "tsc --noEmit",
-    "lint": "prettier src internals --check",
-    "lint:fix": "prettier src internals --write",
+    "lint": "eslint --ext js,ts,tsx ./",
```
lukaw3d, line 15: CI says:
```
$ eslint --ext js,ts,tsx ./

/home/runner/work/explorer/explorer/src/app/pages/HomePage/ParaTimeSelector/HelpScreen/index.tsx
Warning:   [7](https://github.com/oasisprotocol/explorer/actions/runs/4194095200/jobs/7271812264#step:5:8)7:1[9](https://github.com/oasisprotocol/explorer/actions/runs/4194095200/jobs/7271812264#step:5:10)  warning  The ref value 'swiperElRef.current' will likely have changed by the time this effect cleanup function runs. If this ref points to a node rendered by React, copy 'swiperElRef.current' to a variable inside the effect, and use that variable in the cleanup function  react-hooks/exhaustive-deps

✖ 1 problem (0 errors, 1 warning)
```

I think we should throw on warnings too, like wallet
https://github.com/oasisprotocol/oasis-wallet-web/blob/c3da68aba502ba7d26f3067a6ef544b477a282d3/.github/workflows/build-test.yaml#L36

csillag: OK thanks, I cherry-picked the fix.

csillag:
> Right, fixed in #142

OK thanks, I cherry-picked the fix.

package.json
csillag: After the latest commit, the warning is gone.

csillag:
> I think we should throw on warnings too,

Done.

src/app/components/charts/Tooltip.tsx
```diff
@@ -28,7 +28,7 @@ export const TooltipContent = ({ active, payload, formatters }: TooltipContentPr
   if (!active || !payload || !payload.length) {
     return null
   }
-  const { [payload[0].dataKey!]: value, ...rest } = payload[0]?.payload
+  const { [payload[0].dataKey!]: value, ...rest } = payload[0] ? payload[0].payload : undefined
```
lukaw3d, line 31: I don't understand eslint's complaint here, but `= payload[0].payload` should directly work

or, if there wasn't an un-readable `!payload.length` above, then `payload[0] ? payload[0].payload : {}` should be more accurate

.eslintignore
```diff
@@ -0,0 +1,38 @@
...
+build-ext/
+build-dev/
+node_modules/
+stats.json
+.pnp
+.pnp.js
+generated/
+jest.config.js
+
+# misc
```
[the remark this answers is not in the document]
csillag: > I would make it a link to explorer's gitignore

Simply using `.gitignore` won't suffice, since there are files that must be in git, but not linted. But now I have re-created this files based on explorer's gitignore.

package.json
```diff
@@ -12,8 +12,9 @@
...
+    "lint:fix": "yarn run lint --fix",
+    "prettify": "prettier src internals --write",
```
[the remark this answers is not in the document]
csillag: It's unclear to me. According to the docs:

> if you run eslint with the --fix flag, your code will be formatted according to prettier style.

... but this means eslint will also modify/fix other things. I don't see a trivial way to fix the formatting and only the formatting using `eslint-plugin-prettier`. It's still useful to detect non-complience, thought.

src/app/components/charts/Tooltip.tsx
csillag: > I don't understand eslint's complaint here, but `= payload[0].payload` should directly work

Indeed, it does. Simplified.

.eslintrc.js
```diff
@@ -0,0 +1,79 @@
...
+          },
+        ],
+        patterns: [
+          {
+            group: ['grommet/*', '!grommet/es6'],
+            message: "Import 'grommet' or 'grommet/es6/*' instead.",
+          },
+          {
+            group: ['grommet-icons/*', '!grommet-icons/es6'],
+            message: "Import 'grommet-icons' or 'grommet-icons/es6/*' instead.",
```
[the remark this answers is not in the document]
csillag: I'm having trouble finding the right syntax for this one:

> @mui/material, use @mui/material/* instead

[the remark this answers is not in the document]
lukaw3d: Probably simply
```
paths: [
  {
    name: '@mui/material',
    message: "...",
  },
```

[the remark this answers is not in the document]
csillag: OK, all done

.eslintrc.js
```diff
@@ -0,0 +1,79 @@
...
+    'plugin:@typescript-eslint/recommended', // https://typescript-eslint.io/rules/
+    'react-app', // https://github.com/facebook/create-react-app/blob/main/packages/eslint-config-react-app/index.js
+    'plugin:prettier/recommended', // See .prettierrc
+  ],
+  parser: '@typescript-eslint/parser',
+
+  settings: {
+    // Warn about <a target="_blank" rel="noopener"> in components other than "a"
+    // https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-no-target-blank.md#custom-link-components
+    linkComponents: ['Button', 'MediaButton', 'SidebarButton', 'Anchor', 'Link', 'AnchorLink'],
```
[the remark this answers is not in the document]
csillag: > Needs to be updated

Updated.

csillag:
Ok @lukaw3d, I have reacted to everything.

.eslintrc.js
```diff
@@ -0,0 +1,69 @@
...
+            message: "Use '@mui/material/*' instead.",
+          },
+        ],
+        patterns: [
+          {
+            group: ['@mui/system'],
+            message: "Use '@mui/material/styles' instead.",
+          },
+          {
+            group: ['oasis-index/generated/api'],
```
lukaw3d, line 35: This is wrong

.eslintrc.js
```diff
@@ -0,0 +1,69 @@
...
+    'plugin:@typescript-eslint/recommended', // https://typescript-eslint.io/rules/
+    'react-app', // https://github.com/facebook/create-react-app/blob/main/packages/eslint-config-react-app/index.js
+    'plugin:prettier/recommended', // See .prettierrc
+  ],
+  parser: '@typescript-eslint/parser',
+
+  settings: {
+    // Warn about <a target="_blank" rel="noopener"> in components other than "a"
+    // https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-no-target-blank.md#custom-link-components
+    linkComponents: ['Chip', 'Link', 'SocialLink', 'StyledLink'],
```
lukaw3d, line 17: This sucks in a few ways.

- It does not follow renames 
  ```tsx
  import { Link as RouterLink } from 'react-router-dom'
  ```

- 'Link' detects
  ```tsx
  import Link from '@mui/material/Link'
  <Link href="https://aaa.com" target="_blank"></Link>
  ```
  but not 
  ```tsx
  import { Link } from 'react-router-dom'
  <Link to="https://aaa.com" target="_blank"></Link>
  ```

- `{ name: 'Link', linkAttribute: 'to' }, { name: 'Link', linkAttribute: 'href' }` ignores the first

.eslintrc.js
```diff
@@ -0,0 +1,69 @@
...
+  },
+  rules: {
+    'no-restricted-imports': [
+      'error',
+      {
+        paths: [
+          {
+            name: '@mui/material',
+            message: "Use '@mui/material/*' instead.",
```
lukaw3d, line 26: This correctly reduced the bundle size :+1: 3.18 MB => 2.83 MB

.eslintrc.js
lukaw3d: `**/oasis-indexer/generated/api` should work

csillag: Right.

.eslintrc.js
csillag: Indeed, this rule doesn't seem to be that useful. I am for simply dropping it. What do you thing @lukaw3d ?

csillag: Especially since, according to [the docs](https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/jsx-no-target-blank.md#when-to-override-it), 

> Modern browsers (Chrome ≥ 88, Edge ≥ 88, Firefox ≥ 79 and Safari ≥ 12.2) automatically imply rel="noopener". Therefore **this rule is no longer needed**, if legacy browsers are not supported. 

So I would say that we can simply drop this, can't we?

lukaw3d (approved):

PR #139, opened 2023-02-15
Design Feedback: Block details

buberdds:
does not handle max block size (waiting for product decisions)

### DIFF
--- a/src/app/components/Transactions/index.tsx
+++ b/src/app/components/Transactions/index.tsx
@@ -43,14 +43,21 @@ type TransactionProps = {
   isLoading: boolean
   limit: number
   pagination: false | TablePaginationProps
+  verbose?: boolean
 }
 
-export const Transactions: FC<TransactionProps> = ({ isLoading, limit, pagination, transactions }) => {
+export const Transactions: FC<TransactionProps> = ({
+  isLoading,
+  limit,
+  pagination,
+  transactions,
+  verbose = true,
+}) => {
   const { t } = useTranslation()
   const tableColumns = [
     { content: t('common.status') },
     { content: t('common.hash') },
-    { content: t('common.block') },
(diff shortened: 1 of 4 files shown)


src/app/components/Transactions/index.tsx
```diff
@@ -43,14 +43,21 @@ type TransactionProps = {
   isLoading: boolean
   limit: number
   pagination: false | TablePaginationProps
+  verbose?: boolean
 }
 
-export const Transactions: FC<TransactionProps> = ({ isLoading, limit, pagination, transactions }) => {
+export const Transactions: FC<TransactionProps> = ({
```
csillag, line 49: Isn't this the transaction list?
I'm asking because the commit message says

> Introduce verbose mode in blocks list

buberdds: truf, need to edit commit message

src/app/pages/BlockDetailPage/index.tsx
```diff
@@ -87,7 +88,7 @@ export const BlockDetailView: FC<{
 
           <dt>{t('common.hash')}</dt>
           <dd>
-            <CopyToClipboard value={`0x${block.hash}`} />
+            <CopyToClipboard value={block.hash} label={isMobile ? trimLongString(block.hash) : block.hash} />
```
csillag: I have a pending PR that would teach  CopyToClipboard to do this automatically. I think we should discuss what should be best approach here. 
In any case, until that happens, this is fine.

csillag (approved):
Reviewed. LGTM.

src/app/pages/BlockDetailPage/index.tsx
buberdds: I will link PRs. related to https://github.com/oasisprotocol/explorer/pull/106

PR #140, opened 2023-02-16
Search by transaction hash and account address, and show search suggestions

lukaw3d:
Preview https://0a69944a.lw-oasis-explorer-frontend.pages.dev/

Hover and focus states:
search

### DIFF
--- a/src/app/components/Search/search-utils.ts
+++ b/src/app/components/Search/search-utils.ts
@@ -1,3 +1,4 @@
+import { ParaTime } from '../../../config'
 import {
   isValidBlockHeight,
   isValidBlockHash,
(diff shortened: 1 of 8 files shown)


src/app/components/Search/search-utils.ts
```diff
@@ -10,19 +11,18 @@ import { RouteUtils } from '../../utils/route-utils'
...
+    const blockHeight = validateAndNormalize.blockHeight(searchTerm)
+    const txHash = validateAndNormalize.txHash(searchTerm)
+    const evmAccount = validateAndNormalize.evmAccount(searchTerm)
+    const consensusAccount = validateAndNormalize.consensusAccount(searchTerm)
+    if (blockHeight) return RouteUtils.getBlockRoute(parseInt(blockHeight, 10), ParaTime.Emerald)
+    if (txHash) return RouteUtils.getTransactionRoute(txHash, ParaTime.Emerald)
+    if (evmAccount) return RouteUtils.getAccountRoute(evmAccount, ParaTime.Emerald)
+    if (consensusAccount) return RouteUtils.getAccountRoute(consensusAccount, ParaTime.Emerald)
+    // TODO: block hash, contract, validator, event
+    return undefined
```
csillag, line 25: As far as I can tell, this is the code that determines the priority of the various things we can search for. Has this been defined somewhere?

csillag:
Searching for an invalid term on the homepage doesn't result in any visible error messages. just warning on the console. Is that intentional?

csillag:
Scratch that, false alarm, HRM artifact...

csillag:
The placeholder in the search bar mentions "Tnx Hash, Transaction ID". I get the hash, but what is the Transaction Id, and can we really search for it?

csillag (approved):
Reviewed, LGTM.

src/app/components/Search/search-utils.ts
lukaw3d: Oh, no, this is just a simple implementation without supporting multiple results yet. It's not meant to prioritize search types, but just assume that they are all unique.

Multiple results will be a separate PR

PR #142, opened 2023-02-16
Fix useEffect cleanup function eslint warning

lubej:


csillag:
Cosing in favor of #138, which has the same commit now.

PR #145, opened 2023-02-21
Improve charts

lukaw3d:
- Allow line and dot to overflow chart surface
- Reverse charts to display latest data on the right :exclamation:
- Format tooltip number

| Before | After |
| --- | --- |
|  localhost_1234_emerald (1) |  localhost_1234_emerald |

### DIFF
--- a/src/styles/index.css
+++ b/src/styles/index.css
@@ -1,2 +1,6 @@
 /** FigtreeVariable */
 @import '../../node_modules/@fontsource/figtree/variable.css';
+
+svg.recharts-surface {
+  overflow: visible;
+}
--- a/src/app/components/charts/BarChart.tsx
+++ b/src/app/components/charts/BarChart.tsx
@@ -12,7 +12,7 @@ import { COLORS } from '../../../styles/theme/colors'
 
 interface BarChartProps<T extends object> extends Formatters {
   data: T[]
-  dataKey: keyof T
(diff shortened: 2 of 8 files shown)


src/styles/index.css
```diff
@@ -1,2 +1,6 @@
 /** FigtreeVariable */
 @import '../../node_modules/@fontsource/figtree/variable.css';
+
+.recharts-surface {
```
buberdds, line 4: Instead of this custom css selector that no one will remember I adjusted chart rendering area with margins. `https://github.com/oasisprotocol/explorer/blob/master/src/app/pages/DashboardPage/RoseChartCard.tsx#L70`. Anyway, if it's css selector, margin in each chart or in generic  line chart I don't care but it would be nice to unify this.

lukaw3d: I didn't find a good option in recharts to add styles to the svg it produced, but it has this class `<svg class="recharts-surface">`. I'd prefer if we didn't need to involve math `margin = lineStrokeWidth + activeDot.r + activeDot.strokeWidth`

Unified? https://github.com/oasisprotocol/explorer/pull/145/commits/79e60a2f51498c21ddd66f4529f14ad61688a68c

buberdds (approved):

buberdds:
ROSE section does not look good now 
Screenshot 2023-02-22 at 08 53 04

PR #146, opened 2023-02-21
Round ROSE value

buberdds:
In txs list we want to round ROSE to 8 decimal places. Show special label when rounded value is zero.

Screenshot 2023-02-21 at 10 53 34

Screenshot 2023-02-21 at 11 12 18

### DIFF
--- a/src/app/components/RoundedRoseValue/index.tsx
+++ b/src/app/components/RoundedRoseValue/index.tsx
@@ -1,0 +1,28 @@
+import { FC } from 'react'
+import { useTranslation } from 'react-i18next'
+import BigNumber from 'bignumber.js'
+
+type RoundedRoseValueProps = {
+  value?: string
+}
+
+const numberOfDecimals = 8
+
+export const RoundedRoseValue: FC<RoundedRoseValueProps> = ({ value }) => {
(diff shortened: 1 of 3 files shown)


src/app/components/RoundedRoseValue/index.tsx
```diff
@@ -0,0 +1,28 @@
...
+    return null
+  }
+
+  const number = new BigNumber(value)
+  const roundedNumber = number.decimalPlaces(numberOfDecimals, BigNumber.ROUND_HALF_CEIL)
+
+  return (
+    <>
+      {!number.isZero() && roundedNumber.isZero()
+        ? t('common.lessThanRoseAmount', { value: roundedNumber.toFixed(numberOfDecimals) })
```
lubej, line 24: So this one would be `< 0.00000000 ROSE` in case of really small values right?

lubej (approved):
LGTM

src/app/components/RoundedRoseValue/index.tsx
buberdds: yes, I've found sample tx with that case // added to PR desc

PR #147, opened 2023-02-21
Design feedback: latest txs/blocks

buberdds:
- increase global card padding
- adjust skeleton text row to stabilise table height (issue with jumping vertically)
- there is an assumption that if there is a table in a mobile card its content won't feet and we will show horizontal scroll bar. In this case card  "should extend to the right side of the screen to emphasise the horizontal scroll" (screenshot)

Screenshot 2023-02-21 at 12 56 43
- change rounding value

issues with table jumping horizontally moved to separate PRs (rounding / font updates / not sure what else)

### DIFF
--- a/src/app/components/Transactions/index.tsx
+++ b/src/app/components/Transactions/index.tsx
@@ -63,8 +63,8 @@ export const Transactions: FC<TransactionProps> = ({
     { content: t('common.type') },
     { content: t('common.from'), width: '150px' },
     { content: t('common.to'), width: '150px' },
-    { content: t('common.txnFee'), align: TableCellAlign.Right },
-    { align: TableCellAlign.Right, content: t('common.value') },
+    { content: t('common.txnFee'), align: TableCellAlign.Right, width: '250px' },
+    { align: TableCellAlign.Right, content: t('common.value'), width: '250px' },
   ]
   const tableRows = transactions?.map(transaction => ({
     key: transaction.hash,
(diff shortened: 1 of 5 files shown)


src/app/components/Transactions/index.tsx
```diff
@@ -63,8 +63,8 @@ export const Transactions: FC<TransactionProps> = ({
     { content: t('common.type') },
     { content: t('common.from'), width: '150px' },
     { content: t('common.to'), width: '150px' },
-    { content: t('common.txnFee'), align: TableCellAlign.Right },
-    { align: TableCellAlign.Right, content: t('common.value') },
+    { content: t('common.txnFee'), align: TableCellAlign.Right, width: '250px' },
```
csillag, line 66: Why is it that some of the columns headers are aligned to the right, but not all?

csillag (approved):
Reviewed. LGTM.

src/app/components/Transactions/index.tsx
csillag: Yes, I understand the point of right-aligning the numbers. My question is about the column headers. Especially there: 
image

It's a question of personal aesthetics, but I would probably have the column headers aligned uniformly to the middle.  Some them being left and some right feels random to me. Especially "Age" vs "Type", and then "From" -> "To".  Because of the arrow connecting that pair of fields, I would probably align the "from" header to the right, so that both the "from" and the "to" are close to the arrow.

But again, personal asthetics.

PR #151, opened 2023-02-22
Font updates

buberdds:


src/app/components/Transactions/index.tsx
```diff
@@ -75,23 +76,27 @@ export const Transactions: FC<TransactionProps> = ({
       },
       {
         content: (
-          <TrimLinkLabel
-            label={transaction.hash}
-            to={RouteUtils.getTransactionRoute(transaction.hash, ParaTime.Emerald)}
-          />
+          <Typography variant="mono">
+            <TrimLinkLabel
```
csillag, line 80: Wouldn't it make sense to use the Typography inside the link component?

csillag (approved):
Reviewed. LGTM.

src/app/components/Transactions/index.tsx
buberdds: I've done it this way initially (inside the TrimLabel), but moved typography outside the component. Wasn't sure if it's really belong there. Anyway, will gather additional feedback before merging.

lubej: From the usage perspective I prefer if the `Typography` is outside, and the only thing that is actually clickable is the content itself. `Typography` could have some padding, which would make "empty" space clickable.

lubej (approved):

src/styles/theme/defaultTheme.ts
```diff
@@ -409,13 +410,20 @@ export const defaultTheme = createTheme({
     },
     MuiTypography: {
       variants: [
+        {
+          props: { variant: 'mono' },
+          style: () => ({
+            fontFamily: 'Roboto MonoVariable',
```
lukaw3d, line 416: I'd add a monospace fallback

PR #156, opened 2023-02-23
Update search suggestions to match updated mocks

lukaw3d:
| Before | After |
| --- | --- |
| lw-oasis-explorer-frontend pages dev_ (2)  |  localhost_1234_ (1) |

### DIFF
--- a/src/app/components/Search/index.tsx
+++ b/src/app/components/Search/index.tsx
@@ -163,7 +163,6 @@ const SearchCmp: FC<SearchProps> = ({ variant, disabled, onFocusChange }) => {
         }}
         placeholder={searchPlaceholderTranslated}
         fullWidth
-        error={hasError}
         FormHelperTextProps={{
           sx: {
             marginTop: '10px',
--- a/src/app/components/Search/index.tsx
+++ b/src/app/components/Search/index.tsx
@@ -28,11 +28,6 @@ const SearchForm = styled('form', {
   // Approximate size of TextField without helperText, so helperText floats.
   height: '47px',
 
-  // Show helper text when focused or filled.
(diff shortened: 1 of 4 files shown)


src/app/components/Search/index.tsx
```diff
@@ -28,6 +28,11 @@ const SearchForm = styled('form', {
   // Approximate size of TextField without helperText, so helperText floats.
   height: '47px',
 
+  // Show helper text when focused or filled.
+  ':not(:hover, :focus-within) :not(.MuiFormHelperText-filled).MuiFormHelperText-root': {
```
buberdds, line 32: This is so hard to maintain, even comment is saying show when css selector is actually hiding element. Can't we just render SearchSuggestions when needed? Is that a case we are trying to accomplish?

```
 helperText={
          (isHovered || isFocused) && (
            <SearchSuggestions
              onClickSuggestion={suggestion => {
                setValue(suggestion)
              }}
            />
          )
        }
``` ?

buberdds (commented):
not sure it these are parts of this PR, but
there is a weird black border around input
on mockups we have additional yellow indicator (but I don't understand fully understand when to show what looking at mockups)

src/app/components/Search/index.tsx
lukaw3d: Yeah :/

I'm not happy about making it visible on hover, but without it: focus empty, see suggestions, click suggestions, suggestions disappear and don't work

lukaw3d: I'll simplify to `value && <SearchSuggestions` and revisit later :D

buberdds (approved):

PR #159, opened 2023-02-24
Design Feedback: Dashboard

buberdds:
- fix random dashboard CSS issues (more charts issues will be address in another PR)
- prepare bar chart for incoming active accounts endpoint

screenshot with mocked Active Accounts data
```
[
    {
      active_accounts: 634,
      bucket_start: '2022-04-11T22:15:00Z',
    },
    {
      active_accounts: 438,
      bucket_start: '2022-04-11T22:10:00Z',
    },
...
]
```

Screenshot 2023-02-28 at 16 18 12

### DIFF
--- a/src/app/components/Select/index.tsx
+++ b/src/app/components/Select/index.tsx
@@ -20,7 +20,7 @@ const StyledButton = styled(Button)(({ theme }) => ({
   height: '36px',
   minWidth: '135px',
   padding: `0 ${theme.spacing(4)}`,
-  borderRadius: '12px',
+  borderRadius: '36px',
   color: COLORS.white,
   textTransform: 'none',
   justifyContent: 'space-between',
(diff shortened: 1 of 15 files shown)


src/index.tsx
```diff
@@ -14,7 +14,13 @@ import './styles/index.css'
 import './locales/i18n'
 
 Axios.defaults.baseURL = process.env.REACT_APP_API
-const queryClient = new QueryClient({})
+const queryClient = new QueryClient({
+  defaultOptions: {
+    queries: {
+      refetchOnWindowFocus: false,
```
lubej, line 20: Does this affect rose price chart?

src/app/components/Select/index.tsx
```diff
@@ -20,7 +20,7 @@ const StyledButton = styled(Button)(({ theme }) => ({
   height: '36px',
   minWidth: '135px',
   padding: `0 ${theme.spacing(4)}`,
-  borderRadius: '12px',
+  borderRadius: '36px',
```
lubej, line 23: image

So the reason why I didn't set the `border-radius` to the "correct" value according to the designs, is that the dropdown looks a bit off. Not having the same `border-radius` as the trigger button.

src/app/components/charts/BarChart.tsx
```diff
@@ -11,29 +11,58 @@ import { TooltipContent, type Formatters } from './Tooltip'
...
+const BarChartCmp = <T extends object>({
+  cartesianGrid,
+  data,
+  dataKey,
+  formatters,
+  rounded,
+  withLabels,
+  withBarBackground,
+}: BarChartProps<T>) => (
+  <ResponsiveContainer width="100%" height={80}>
```
lubej, line 33: Hardcoding the height here messes up `Transactions per day` chart

src/index.tsx
buberdds: It's intentional global change, affects all queries.

src/app/components/Select/index.tsx
buberdds: dropdown list item hover state ? I will wait for Don's feedback in this case. thx for info

buberdds: Don wants to keep border-radius from designs.

lubej: Right, would be nice to sync the dropdown container as well - I didn't found any good way of doing this.

src/app/components/charts/BarChart.tsx
buberdds: reverted as this is handled in one of your PR

lubej (approved):

PR #161, opened 2023-02-26
Transaction details design improvements

csillag:
Solves the following issues:

- (Probably) The Transactions could do with a bit more space at the top - add 20px padding above the title, please?
- Drop shadow is missing from Transaction title banner - also the wider grey should be removed
- The 'Success' pill text should be green and have space between the text and icon
- 'Failed' pill text should be red and have space between the text and icon
- 'Block' row should be 'Height's 'Block Height' and the text should be styled as link
- 'From' and 'To' should be styled as link

Before:

image

After:

image


csillag:
Design feedback from @donouwens 

> Looks good

src/app/pages/TransactionDetailPage/index.tsx
```diff
@@ -82,12 +83,17 @@ export const TransactionDetailPage: FC = () => {
...
+        <SubPageCard
+          featured
+          title={t('transaction.header')}
+          key={`${transaction.round}_${transaction.index}`}
+        >
+          <StyledDescriptionList titleWidth={isMobile ? '100px' : '200px'}>
             <dt>{t('common.hash')}</dt>
             <dd>
-              {isMobile ? trimLongString(transaction.hash) : transaction.hash}
+              <Typography variant="mono" component="span" sx={{ color: COLORS.brandDark, fontWeight: 700 }}>
```
lukaw3d, line 94: Should these be links to itself, instead of just looking like links? Or should it copy the text? @donouwens

lukaw3d (approved):

src/app/pages/TransactionDetailPage/index.tsx
csillag: So currently, 
 - the hash is formatted as a link ✔️ 
 - the string itself is not a link, so clicking on it does nothing
 - besides the string, we have a "copy to clipboard" button, which does copy it to the clipboard. ✔️ 

Is that OK, @donouwens ?

csillag: OK, so we can close this.

PR #162, opened 2023-02-26
Use Fade animation when showing/hiding Tooltips

csillag:
As requested here: https://app.clickup.com/t/862j7t8az

### DIFF
--- a/src/styles/theme/defaultTheme.ts
+++ b/src/styles/theme/defaultTheme.ts
@@ -1,5 +1,6 @@
 import { createTheme } from '@mui/material/styles'
 import { COLORS } from './colors'
+import Fade from '@mui/material/Fade'
 
 declare module '@mui/material/styles' {
   interface Palette {
@@ -429,6 +430,10 @@ export const defaultTheme = createTheme({
       },
     },
     MuiTooltip: {
+      defaultProps: {
+        TransitionComponent: Fade,
+        TransitionProps: { timeout: 600 },
+      },
       styleOverrides: {
         arrow: {
           color: COLORS.brandExtraDark,


csillag:
Btw, here is the list of supported animations: https://mui.com/material-ui/react-tooltip/#transitions

And feedback from @donouwens : 
> Yeah Fade is the right one

src/app/components/CopyToClipboard/index.tsx
```diff
@@ -47,7 +48,15 @@ export const CopyToClipboard: FC<CopyToClipboardProps> = ({ label, value }) => {
   }, [])
 
   return (
-    <Tooltip arrow onOpen={hideTooltip} open={isCopied} placement="right" title={t('clipboard.success')}>
```
lukaw3d, line 50: Title is inaccurate - we already had a default animation

| Before | After |
| --- | --- |
|  Recording 2023-02-27 at 19 24 26 | r |

lukaw3d: And this is now inconsistent with other tooltips

Recording 2023-02-27 at 19 48 43

csillag: Indeed, so the default animation was Grow, which we are replacing with Fade now, as specified by @donouwens.
I have updated the title of the PR to reflect that.
Indeed, there are other places where we also have similar tooltips, I am guessing that we should replace all of them, too, but I have asked @donouwens to confirm before proceeding.

csillag: I have already updated the code to use `Fade` everywhere. 
And according to @donouwens :

> My preference is Fade - as it’s a bit calmer.

csillag: > I have already updated the code to use Fade everywhere.

I mean, all the _two_ locations.

lukaw3d: Timings aren't consistent :D 

perhaps defaultTheme is a better place for this
```ts
    MuiTooltip: {
      defaultProps: {
        TransitionComponent: Fade,
        TransitionProps: { timeout: 600 },
      },
```

lukaw3d (approved):

src/app/components/CopyToClipboard/index.tsx
csillag: Indeed. Moved in there.

PR #164, opened 2023-02-27
Add percentage calculation for transactions chart

lubej:
## Problem

Percentage value is currently hardcoded to 23%

## Solution

Calculate percentage compared to the average history of the graph, which seems like the most logical solution

### DIFF
--- a/src/app/pages/DashboardPage/TransactionsChartCard.tsx
+++ b/src/app/pages/DashboardPage/TransactionsChartCard.tsx
@@ -30,11 +30,26 @@ const TransactionsChartCardCmp: FC<TransactionsChartCardProps> = ({ chartDuratio
 
   const totalTransactions = data?.data.buckets.reduce((acc, curr) => acc + curr.tx_volume, 0) ?? 0
 
+  const getPercentageDiff = (): number | null => {
+    if (!lineChartData?.length || lineChartData?.length < 2) {
+      return null
+    }
+
+    const [earliestBucket] = lineChartData
+    const latestBucket = lineChartData[lineChartData.length - 1]
+
+    return (
+      ((earliestBucket.volume_per_second - latestBucket.volume_per_second) / latestBucket.volume_per_second) *
+      100
+    )
+  }
+
+  const percentage = getPercentageDiff()
+
   return (
     <SnapshotCard
       title={t('common.transactions')}
-      // TODO: show real percentage value
-      badge={<PercentageGain percentage={23} />}
+      badge={percentage !== null && <PercentageGain percentage={percentage} />}
       label={totalTransactions.toLocaleString()}
     >
       {lineChartData && (


src/app/pages/DashboardPage/TransactionsChartCard.tsx
```diff
@@ -29,11 +29,24 @@ const TransactionsChartCardCmp: FC<TransactionsChartCardProps> = ({ chartDuratio
...
+    }
+
+    const [firstBucket, ...restBuckets] = lineChartData
+    const averageRestBuckets =
+      restBuckets.reduce((acc, curr) => acc + curr.volume_per_second, 0) / restBuckets.length
+
+    return ((averageRestBuckets - firstBucket.volume_per_second) / averageRestBuckets) * -100
+  }
+
+  const percentage = getPercentageDiff()
```
lukaw3d, line 44: > Calculate percentage compared to the average history of the graph, which seems like the most logical solution

Uhhh, I did not expect this to show deviation. Don't we want relative change?

lubej: It is relative change compared to average of the all previous data points? Or am I misinterpreting something.

lukaw3d: Update: discussed on frontend sync that we actually want relative change (final value - starting value)

lubej: Updated, so that it reflect the relative change from final to starting value

lukaw3d: This math is completely wrong

lukaw3d (changes requested):

src/app/pages/DashboardPage/TransactionsChartCard.tsx
lubej: My bad, should be good now.

src/app/pages/DashboardPage/TransactionsChartCard.tsx
```diff
@@ -30,11 +30,25 @@ const TransactionsChartCardCmp: FC<TransactionsChartCardProps> = ({ chartDuratio
...
+  const getPercentageDiff = (): number | null => {
+    if (!lineChartData?.length || lineChartData?.length < 2) {
+      return null
+    }
+
+    const [firstBucket] = lineChartData
+    const lastBucket = lineChartData[lineChartData.length - 1]
+
+    return (
+      ((firstBucket.volume_per_second - lastBucket.volume_per_second) / lastBucket.volume_per_second) * 100
```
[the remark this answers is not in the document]
lukaw3d: Variables are awkwardly named because data is reversed. Perhaps we should reverse it earlier, or rename them to "latestBucket" and "earliestBucket"

lukaw3d (approved):

PR #167, opened 2023-02-27
Router dynamic layer param

lubej:
## Problem

Decision had to be made, if router would use dynamic param or static routes.

## Solution

Decision had been made to go with dynamic param for Layer.


src/routes.tsx
```diff
@@ -27,48 +28,51 @@ export const routes: RouteObject[] = [
       ...RouteUtils.getEnabledParaTimes()
         .map((paraTime): RouteObject[] => [
           {
-            path: `/${paraTime}`,
+            path: `/:${paraTime}`,
```
lubej, line 31: This is a bit hacky, could have been just replaced with `:paraTime` and use custom hook, that would consume `useParams()`

buberdds: imo this "hacky way" is an antipattern. Does it even work correctly?
- there is no point in creating dynamic params specific to each paraTime. 
- generated routes look like: `/:emerald`, `/:sapphire` etc. At the end all URLs will match whatever will be the first route in a loop based on `enabledParaTimes` order. For `/sapphire` param route will match Emerald because `:emerald` param will be true and it will be equal to `sapphire`

buberdds:
side node: we prob should think about naming. We will work on Consensus too and it is not ParaTime. In BE and docs name  "layer" is used  https://github.com/oasisprotocol/explorer/blob/master/src/oasis-indexer/generated/api.ts#L847 ?

src/routes.tsx
lubej: Right, agree that this is an anti-pattern. And as you pointed out, it does not work as expected

src/app/hooks/useLayer.ts
```diff
@@ -0,0 +1,12 @@
+import { useParams } from 'react-router-dom'
+import { Layer } from '../../config'
+
+export const useLayer = () => {
```
buberdds, line 4: not used atm? what would be a use case? In views we will use `const { layer } = useParams()` probably

lubej: Good point, redundant.

buberdds (approved):
Looks good to me, but let's wait for other folks before merging as this will impact future development. 

It looks like the only downside currently is that we will have bunch of  `layer as Layer` type assertion (not sure we care about it)

lukaw3d (approved):

PR #173, opened 2023-03-02
Refactor search and add no results page with suggestions

lukaw3d:
| Before (search -> block detail) | After (search -> search results -> block detail) |
| --- | --- |
| localhost_1234_search_q=a (3)  |  localhost_1234_search_q=a (3) |
| localhost_1234_emerald_blocks_111 | localhost_1234_search_q=a |
| localhost_1234_emerald_blocks_111 (1)  | localhost_1234_search_q=a (1) |

### DIFF
--- a/src/app/components/Search/SearchSuggestionsLinks.tsx
+++ b/src/app/components/Search/SearchSuggestionsLinks.tsx
@@ -1,0 +1,30 @@
+import { FC } from 'react'
+import { searchSuggestionTerms } from '../../components/Search/search-utils'
+import { Trans, useTranslation } from 'react-i18next'
+import { Link as RouterLink } from 'react-router-dom'
+import Link from '@mui/material/Link'
+import { RouteUtils } from '../../utils/route-utils'
+import { OptionalBreak } from '../OptionalBreak'
+
+interface Props {}
+
+export const SearchSuggestionsLinks: FC<Props> = () => {
+  const { t } = useTranslation()
+  const { suggestedBlock, suggestedTransaction, suggestedAccount } = searchSuggestionTerms
+
+  return (
+    <Trans
+      t={t}
+      i18nKey="search.searchSuggestions"
+      components={{
+        OptionalBreak: <OptionalBreak />,
+        BlockIcon: <></>,
+        BlockLink: <Link component={RouterLink} to={RouteUtils.getSearchRoute(suggestedBlock)} />,
+        TransactionIcon: <></>,
+        TransactionLink: <Link component={RouterLink} to={RouteUtils.getSearchRoute(suggestedTransaction)} />,
+        AccountIcon: <></>,
+        AccountLink: <Link component={RouterLink} to={RouteUtils.getSearchRoute(suggestedAccount)} />,
+      }}
+    ></Trans>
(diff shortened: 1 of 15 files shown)


src/app/components/Search/SearchSuggestionsLinks.tsx
```diff
@@ -0,0 +1,31 @@
...
+  const { suggestedBlock, suggestedTransaction, suggestedAccount } = searchSuggestionTerms
+
+  return (
+    <Trans
+      t={t}
+      i18nKey="search.searchSuggestions"
+      components={{
+        OptionalBreak: <OptionalBreak />,
+        BlockIcon: <></>,
+        BlockLink: <Link component={RouterLink} to={RouteUtils.getSearchRoute(suggestedBlock)} />,
```
lukaw3d, line 23: Slightly cheated here: these suggestions redirect to search results, instead of merely modify search input like SearchSuggestionsButtons. I didn't want to maintain emitting an event out from suggestions and drilling setValue into instances of Search.

src/app/pages/SearchResultsPage/index.tsx
```diff
@@ -0,0 +1,113 @@
...
+
+export const SearchResultsView: FC<{
+  searchQueries: SearchQueries
+}> = ({ searchQueries }) => {
+  const { t } = useTranslation()
+  const isAnyLoading = Object.values(searchQueries).some(query => query.isLoading)
+  const hasNoResults = !isAnyLoading && Object.values(searchQueries).every(query => !query.results)
+
+  return (
+    <PageLayout>
```
lukaw3d, line 79: TODO: add special handling for mobile no results

src/app/components/Search/SearchSuggestionsLinks.tsx
lukaw3d: Don confirms THIS IS BETTER

src/app/components/Search/search-utils.ts
```diff
@@ -1,33 +1,25 @@
...
-  }
+export const searchSuggestionTerms = {
+  suggestedTransaction: 'b1e68ca814d913064bd6b9460efcb64b4c6d07f3b98fa659beed46164398a830',
+  suggestedBlock: '1396255',
+  suggestedAccount: '0xBA504818FdD8D3dBA2Ef8fD9B4F4D5c71aD1d1D3',
 }
 
 export const validateAndNormalize = {
   blockHeight: (searchTerm: string) => {
+    const thousandSeparator = new Intl.NumberFormat().formatToParts(1234).find(p => p.type === 'group')?.value
```
lubej, line 19: This `1234` looks like it's hardcoded.

lukaw3d: This is intentional. I couldn't find a thousand separator in Intl.NumberFormat.resolvedOptions, so I use >1000 number to find ","

lukaw3d: From sync: we should filter out everything but 0-9a-z so that copying someone else's localized "1,201,390" would still parse on other localisations

lubej (approved):
LGTM

PR #175, opened 2023-03-02
Use ellipsis instead of ... for trimming long strings (with different font)

lukaw3d:
| Before | After |
| --- | --- |
|  localhost_1234_emerald (2) | localhost_1234_emerald  |


src/app/utils/trimLongString.ts
```diff
@@ -1,8 +1,8 @@
-export function trimLongString(value: string, trimStart = 6, trimEnd = 6, ellipsis = '...') {
+export function trimLongString(value: string, trimStart = 6, trimEnd = 6, ellipsis = '…') {
   if (!value) {
     return
   }
-  const trimmedLength = trimStart + 3 + trimEnd
+  const trimmedLength = trimStart + ellipsis.length + trimEnd
```
buberdds, line 5: isn't it now 1 instead of 3 ?

buberdds (approved):

PR #176, opened 2023-03-02
Update page title

lukaw3d:


public/app.webmanifest
```diff
@@ -1,7 +1,7 @@
 {
   "$schema": "https://json.schemastore.org/web-manifest-combined.json",
-  "short_name": "Oasis Network Explorer",
-  "name": "The official Oasis Network Explorer.",
+  "short_name": "Oasis Explorer",
+  "name": "Official explorer for the Oasis Network.",
```
lukaw3d, line 4: cc @donouwens @tjanez

tjanez (approved):

public/app.webmanifest
tjanez: I think this makes sense :+1:

PR #177, opened 2023-03-03
Dashboard: design feedback # 2

buberdds:
- show proper heading for sapphire
- remove inactive nodes badge 
- tweak line charts in ParaTime snapshot
- Total Transactions defaults to monthly duration
- transactions per day chart 
  - enable duration selector
  - weekly needs unique layout
  - for all time/yearly we want to group stats per month and show avg for a day in a month

### DIFF
--- a/src/app/utils/chart-utils.ts
+++ b/src/app/utils/chart-utils.ts
@@ -1,4 +1,5 @@
-import { GetLayerStatsTxVolumeParams } from '../../oasis-indexer/api'
+import isSameMonth from 'date-fns/isSameMonth'
+import { GetLayerStatsTxVolumeParams, type TxVolume } from '../../oasis-indexer/api'
 
 export enum ChartDuration {
   TODAY = 'TODAY',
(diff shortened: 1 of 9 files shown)


src/app/utils/chart-utils.ts
```diff
@@ -34,3 +35,31 @@ export const chartDurationToDaysMap = {
...
+    acc.push({
+      bucket_start: cur.bucket_start,
+      tx_volume: cur.tx_volume,
+      numberOfDays: 1,
+    })
+    return acc
+  }, [])
+}
+
+export const getMonthlyBucketsDailyAverage = (buckets: Buckets): Buckets => {
```
lubej, line 58: Not sure why this is needed. Number of days in the month could be calculated from https://date-fns.org/v2.29.3/docs/getDaysInMonth. So that above reduce could be removed. Or am I missing something?

lubej: Right, disregard. I thought they were full months.

lubej (approved):

PR #179, opened 2023-03-07
Dashboard design feedback # 3

buberdds:
After a chat with Don I need to
- change Transactions Per Day chart labels 
- update Learning Materials + include Sapphire changes. For now I don't catch errors in any error boundary (not sure we need it as these are going to be implementation not user errors)


lubej (approved):
LGTM

Issue #130, opened 2023-02-14
Fix transaction details gas


lukaw3d:
What should transaction details show for gas?

- API:
  https://index-staging.oasislabs.com/v1/emerald/transactions/f89c23ce07e17c503735f8f6393ae30592ceb7c39d81c24990aff15b5dae1103
  "fee": "800000000000000000",
  "gas_limit": 8000000,

- Blockscout:
  https://explorer.emerald.oasis.dev/tx/0xe976b0d3fadf1197ab8549f5995606605414df94273322a51a57e41ea2ad3855/internal-transactions
  Transaction Fee  0.2979207 ROSE
  Gas Price  100 Gwei
  Gas Limit  8,000,000
  Max Fee per Gas  100 Gwei
  Max Priority Fee per Gas  100 Gwei

- Current explorer incorrectly shows:
  http://localhost:1234/emerald/transactions/f89c23ce07e17c503735f8f6393ae30592ceb7c39d81c24990aff15b5dae1103
  Txn Fee  800000000000000000 ROSE
  Gas limit  8000000 ROSE

> - oasisscan:
>   https://www.oasisscan.com/paratimes/transactions/0xe976b0d3fadf1197ab8549f5995606605414df94273322a51a57e41ea2ad3855?runtime=000000000000000000000000000000000000000000000000e2eaa99fc008f87f
>   "gasPrice": 100000000000,
>   "gasLimit": 8000000,
>

mitjat:
IIUC, from the backend perspective there are ~two~ 1.5 issues:
- Documentation bug/deficiency, where we should state that the fee is in consensus/runtime base units ([here](https://github.com/oasisprotocol/oasis-indexer/blob/96a140711e77d460da56a0e5f5bf4e6bbaf8283b/api/spec/v1.yaml#L1124-L1124) for consensus and ... oh, actually paratime fee units are already documented [here](https://github.com/oasisprotocol/oasis-indexer/blob/96a140711e77d460da56a0e5f5bf4e6bbaf8283b/api/spec/v1.yaml#L1915-L1915))
[...]

PR #141, opened 2023-02-16
Design Feedback: Block details #2

buberdds:
sync with new block details mockups as they were updated yesterday

Screenshot 2023-02-16 at 12 42 10

PR #182, opened 2023-03-07
Fix gas display and document block gas limit values

lukaw3d:
- Documented source of gas limit values (related to https://github.com/oasisprotocol/explorer/pull/141/commits/b01917059397bb668d06a402b1d9d2acf5e4fea8)
- Fixes https://github.com/oasisprotocol/explorer/issues/130

| Before | After |
| --- | --- |
| localhost_1234_emerald_transactions_0205643075fd26f8d0bbd1993c841bd4406809a515b9b4a6638ccbc898042436 (1)  |  localhost_1234_emerald_transactions_0205643075fd26f8d0bbd1993c841bd4406809a515b9b4a6638ccbc898042436 |

### DIFF
--- a/src/oasis-indexer/api.ts
+++ b/src/oasis-indexer/api.ts
@@ -36,7 +36,36 @@ export const useGetRuntimeTransactions: typeof generated.useGetRuntimeTransactio
         (data: generated.RuntimeTransactionList) => {
           return {
             ...data,
-            transactions: (data.transactions || []).map(tx => {
+            transactions: data.transactions.map(tx => {
+              return {
+                ...tx,
+                fee: tx.fee ? fromBaseUnits(tx.fee, paraTimesConfig.emerald!.decimals) : undefined,
+                amount: tx.amount ? fromBaseUnits(tx.amount, paraTimesConfig.emerald!.decimals) : undefined,
+              }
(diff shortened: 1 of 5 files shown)


src/oasis-indexer/api.ts
```diff
@@ -36,7 +36,36 @@ export const useGetRuntimeTransactions: typeof generated.useGetRuntimeTransactio
         (data: generated.RuntimeTransactionList) => {
           return {
             ...data,
-            transactions: (data.transactions || []).map(tx => {
+            transactions: data.transactions.map(tx => {
+              return {
+                ...tx,
+                fee: tx.fee ? fromBaseUnits(tx.fee, paraTimesConfig.emerald!.decimals) : undefined,
```
csillag, line 42: We have a parameter called `runtime`, but we are still using hard-coded emerald values here. Is that intentional?

csillag (approved):
Reviewed. LGTM. 

(I had one small question, but it only relates to formatting, so not crucial.)

src/oasis-indexer/api.ts
lukaw3d: Yeah, I'll leave that to sapphire migration PRs

PR #183, opened 2023-03-07
Make code editors check types in js files by default

lukaw3d:
Previously, if there was a typo in orval.config.js, it would not get noticed. Most other .js files have an explicit `// @ts-check` at the top tho


lukaw3d:
Ah. Pain. I don't really want to install orval deps

PR #186, opened 2023-03-08
Sync ActiveAccounts section with API

buberdds:
Filter param is broken in API so I cannot fully finish this one. For now only week and month duration is working correctly on FE side.


lubej (approved):
LGTM

PR #188, opened 2023-03-09
Fix storybook and add more components

lukaw3d:


.storybook/main.js
```diff
@@ -14,4 +17,10 @@ module.exports = {
   features: {
     emotionAlias: false,
   },
+  async webpackFinal(config) {
+    config.resolve ??= {}
```
lubej, line 21: Not supported in vanilla JS. Which breaks the build.

lubej, line 21:
```suggestion
    config.resolve = config.resolve ?? {}
```

lukaw3d: :+1: Will degrade to support older nodejs versions

.storybook/main.js
```diff
@@ -10,8 +13,16 @@ module.exports = {
   framework: '@storybook/react',
   core: {
     builder: '@storybook/builder-webpack5',
+    disableTelemetry: true,
```
lukaw3d, line 16: they had telemetry enabled by default :O

lubej (approved):

PR #191, opened 2023-03-10
Split detail views into reusable components for search

lukaw3d:


src/app/pages/AccountDetailsPage/index.tsx
```diff
@@ -18,7 +18,11 @@ export const AccountDetailsPage: FC = () => {
   return (
     <PageLayout>
       <SubPageCard featured title={t('account.title')}>
-        <Account account={account} isLoading={accountQuery.isLoading} roseFiatValue={rosePriceQuery.data} />
+        <AccountDetailsView
+          isLoading={accountQuery.isLoading}
+          account={account}
+          roseFiatValue={rosePriceQuery.data}
+        ></AccountDetailsView>
```
buberdds, line 25: we rather use self-closing format

src/app/pages/AccountDetailsPage/index.tsx
```diff
@@ -30,3 +34,11 @@ export const AccountDetailsPage: FC = () => {
     </PageLayout>
   )
 }
+
+export const AccountDetailsView: FC<{
+  isLoading: boolean
+  account: RuntimeAccount | undefined
+  roseFiatValue: number | undefined
+}> = ({ isLoading, account, roseFiatValue }) => {
+  return <Account account={account} isLoading={isLoading} roseFiatValue={roseFiatValue} />
```
buberdds, line 43: What's a point of `AccountDetailsView` wrapper in general?

src/app/pages/AccountDetailsPage/index.tsx
```diff
@@ -30,3 +34,11 @@ export const AccountDetailsPage: FC = () => {
     </PageLayout>
   )
 }
+
+export const AccountDetailsView: FC<{
+  isLoading: boolean
```
buberdds, line 39: We have this type defined. Do we need to duplicate it?

buberdds (approved):

src/app/pages/AccountDetailsPage/index.tsx
lukaw3d: I wanted a consistent pattern to reuse on search: AccountDetailsView, BlockDetailView, TransactionDetailView, since search should always almost match detail pages

src/app/pages/AccountDetailsPage/index.tsx
lukaw3d: I would keep it decoupled, unless we rename Account component to AccountDetailsView (and later split out Account component if we ever need a more flexible reusable one)

src/app/pages/AccountDetailsPage/index.tsx
lukaw3d: I hate it. I'll add eslint rule

PR #192, opened 2023-03-10
Mock sapphire APIs

lukaw3d:
"Redirects" currently used APIs from sapphire to emerald.

Thoughts on this approach?


lubej (approved):

csillag (approved):
On the technical level I see that this is OK, but on the conceptual level, I'm sure how interchangeable are the data of the different runtimes...

PR #193, opened 2023-03-10
Render hourly/daily chart in ActiveAccounts

buberdds:
Daily chart needs some customisation unfortunately (hourly buckets won't be added to API)

### DIFF
--- a/src/app/utils/chart-utils.ts
+++ b/src/app/utils/chart-utils.ts
@@ -1,5 +1,5 @@
 import isSameMonth from 'date-fns/isSameMonth'
-import { GetLayerStatsTxVolumeParams, type TxVolume } from '../../oasis-indexer/api'
+import { GetLayerStatsTxVolumeParams, type TxVolume, type ActiveAccounts } from '../../oasis-indexer/api'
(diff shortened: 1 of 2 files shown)


src/app/utils/chart-utils.ts
```diff
@@ -10,8 +10,8 @@ export enum ChartDuration {
 
 export const durationToQueryParams = {
   [ChartDuration.TODAY]: {
-    bucket_size_seconds: 60 * 60,
-    limit: 24,
+    bucket_size_seconds: 60 * 5,
+    limit: (60 / 5) * 24, // full day for 5 minutes buckets
```
buberdds, line 14: @lubej this is changing defaults you created. I can move it to ActiveAccounts if you want, because this will affect your tx chart work. API needs 300 for 5 mins buckets. And we need fetch full day of 5mins buckets. For active accounts we need to use every 12th item (or sum groups of 12, in case of tx_volume)

lubej: Disregard for now. Can be merged in, as is.

PR #194, opened 2023-03-10
Show proper values in TotalTransactions

buberdds:
Based on Don's feedback we want to have a sum of TXs over time here to differentiate Total Transactions chart a little bit as we have three charts that are using the same endpoint.

### DIFF
--- a/src/app/utils/chart-utils.ts
+++ b/src/app/utils/chart-utils.ts
@@ -64,6 +64,16 @@ export const getMonthlyBucketsDailyAverage = (buckets: Buckets): Buckets => {
   }))
 }
 
+export const getTotalTxValues = (buckets: Buckets): Buckets => {
+  return buckets?.reduce((acc: TxVolume[], cur, index) => {
+    acc.push({
+      bucket_start: cur.bucket_start,
+      tx_volume: index > 0 ? cur.tx_volume + acc[index - 1].tx_volume : cur.tx_volume,
+    })
+    return acc
+  }, [])
+}
+
 export const filterHourlyActiveAccounts = (
   windows: ActiveAccounts[] | undefined,
 ): ActiveAccounts[] | undefined => {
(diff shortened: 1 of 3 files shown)


lubej (commented):
This looks weird to me, cumulative `tx_volume`. But ok, could we change the label to something else at least, as [date] [tx_volume] pair does not correctly represent [tx_volume] on a given date.

buberdds:
Label updated. Agree that charts with line that always goes up is weird and I was talking about it with Don. Overall strategy is that we will replace this part whenever we will have a new stats endpoint ready (after MVP)

lubej (approved):

src/app/utils/chart-utils.ts
```diff
@@ -64,6 +64,16 @@ export const getMonthlyBucketsDailyAverage = (buckets: Buckets): Buckets => {
...
 
+export const getTotalTxValues = (buckets: Buckets): Buckets => {
+  return buckets?.reduce((acc: TxVolume[], cur, index) => {
+    acc.push({
+      bucket_start: cur.bucket_start,
+      tx_volume: index > 0 ? cur.tx_volume + acc[index - 1].tx_volume : cur.tx_volume,
+    })
+    return acc
+  }, [])
+}
```
lukaw3d, line 75: I think this would be much more readable:
```ts
export const getTotalTxValues = (buckets: Buckets): Buckets => {
  let acc = 0
  return buckets?.map(bucket => {
    acc += bucket.tx_volume
    return {
      bucket_start: bucket.bucket_start,
      tx_volume: acc,
    }
  })
}
```

and it can be generalized:
```ts
export function cumulativeSum<T>(
  items: T[],
  accumulateField: { [F in keyof T]: T[F] extends number ? F : never }[keyof T],
): T[] {
  let acc = 0
  return items.map(item => {
    acc += item[accumulateField] as number
    return { ...item, [accumulateField]: acc }
  })
}

buckets ? cumulativeSum(buckets, 'tx_volume') : undefined
```

PR #195, opened 2023-03-10
Daily transactions chart

lubej:
## Problem

Daily duration changes in https://github.com/oasisprotocol/explorer/pull/193

## Solution

Group buckets by 12 length windows

### DIFF
--- a/src/app/utils/chart-utils.ts
+++ b/src/app/utils/chart-utils.ts
@@ -1,4 +1,7 @@
+import startOfHour from 'date-fns/startOfHour'
+import startOfDay from 'date-fns/startOfDay'
 import isSameMonth from 'date-fns/isSameMonth'
+import startOfMonth from 'date-fns/startOfMonth'
 import { GetLayerStatsTxVolumeParams, type TxVolume, type ActiveAccounts } from '../../oasis-indexer/api'
 
 export enum ChartDuration {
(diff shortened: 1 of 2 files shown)


src/app/utils/chart-utils.ts
```diff
@@ -69,3 +69,38 @@ export const filterHourlyActiveAccounts = (
...
+  )
+
+  return chunkArray
+    .map(groupedBy => {
+      return {
+        [sumKey]: groupedBy.reduce((acc, curr) => acc + (curr[sumKey] as number), 0),
+        [dateKey]: groupedBy[0][dateKey],
+      }
+    })
+    .flat()
```
buberdds, line 105: Is there a need to call flat method at this point ?

src/app/utils/chart-utils.ts
```diff
@@ -69,3 +69,38 @@ export const filterHourlyActiveAccounts = (
 ): ActiveAccounts[] | undefined => {
   return windows?.filter((value, index) => index % 12 === 0)
 }
+
+export const sumBucketsByChucks = <T>(
```
buberdds, line 73: looking at all loops here this logic is so custom that I don't think we need sumKey dataKey args and generics, but Ok.

src/app/utils/chart-utils.ts
```diff
@@ -69,3 +69,38 @@ export const filterHourlyActiveAccounts = (
...
+  dateKey: keyof T,
+  numOfChunks: number,
+) => {
+  if (!buckets) {
+    return []
+  }
+
+  const chunkArray = buckets.reduce(
+    (accArray, item, index) => {
+      const chunkIndex = Math.floor(index / numOfChunks)
```
buberdds, line 85: shouldn't we group by hour, something like isSameHour ?

src/app/utils/chart-utils.ts
lubej: Good point, removed.

src/app/utils/chart-utils.ts
lubej: I was looking into active accounts chart, which has the "offset" as well. It would make more sense. Should I add it to Active accounts chart as well?

src/app/utils/chart-utils.ts
lubej: Was just thinking that it might be useful to reuse, for all time chart, or something similar.

src/app/utils/chart-utils.ts
buberdds: It's not the same. In active accounts we show every 12th item and we do not group/sum them as windows overlaps for 5 minutes buckets.

lubej: Right, after talking with Don, summing by hour makes the most sense.

buberdds (approved):

PR #199, opened 2023-03-10
Don't transform error responses

lukaw3d:
Fixes "TypeError: Cannot read properties of undefined (reading 'map')"

### DIFF
--- a/src/oasis-indexer/api.ts
+++ b/src/oasis-indexer/api.ts
@@ -34,7 +34,8 @@ export const useGetRuntimeTransactions: typeof generated.useGetRuntimeTransactio
       ...options?.axios,
       transformResponse: [
         ...arrayify(axios.defaults.transformResponse),
-        (data: generated.RuntimeTransactionList) => {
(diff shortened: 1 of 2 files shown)


src/oasis-indexer/api.ts
```diff
@@ -34,7 +34,8 @@ export const useGetRuntimeTransactions: typeof generated.useGetRuntimeTransactio
       ...options?.axios,
       transformResponse: [
         ...arrayify(axios.defaults.transformResponse),
-        (data: generated.RuntimeTransactionList) => {
+        (data: generated.RuntimeTransactionList, headers, status) => {
+          if (!(status && 200 <= status && status < 300)) return data
```
lubej, line 38: Any reason why we would need to check all 2xx codes?

lukaw3d: Not really

lubej: Would prefer if we would only check for code 200. If it will cause problems, we can always change.

lubej (approved):

PR #206, opened 2023-03-14
Sync Nodes component with incoming API changes

buberdds:
not able to test it, but it should be Ok when new API is released


lubej (commented):
LGTM. Can we keep this PR open, so it does not brake the dashboard page.

lubej (approved):

buberdds:
Discussed on Slack, I will merge it and we will have temp 404s until BE release.

PR #212, opened 2023-03-15
Implement displaying sapphire block and transaction list and details

csillag:
This handles 
 - blocks lists
 -  transaction lists
 -  block details
 -  transaction details.
 - network header

Does _not_ handle:
 - Accounts
 - Tokens
 - Stats
 - Graphs


lukaw3d:
Search results PR breaks assumptions made here

csillag:
> Search results PR breaks assumptions made here

Reworked according to the agreed solution

lukaw3d (approved):

PR #220, opened 2023-03-17
Break circular imports in config vs api

csillag:
If _can_ work, but let's avoid the pain of hidden bugs.

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -1,4 +1,5 @@
-import { Layer } from './oasis-indexer/api'
+// eslint-disable-next-line no-restricted-imports
+import { Layer } from './oasis-indexer/generated/api' // We get this from the generated code to avoid circular imports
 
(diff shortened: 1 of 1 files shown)


lubej (approved):

src/config.ts
```diff
@@ -1,4 +1,5 @@
-import { Layer } from './oasis-indexer/api'
+// eslint-disable-next-line no-restricted-imports
+import { Layer } from './oasis-indexer/generated/api' // We get this from the generated code to avoid circular imports
```
lukaw3d, line 2: Should also revert e6657af7a1e65bf9374a0a738fcf9974cff5e892

csillag: > Should also revert e6657af7a1e65bf9374a0a738fcf9974cff5e892

See https://github.com/oasisprotocol/explorer/pull/227

PR #225, opened 2023-03-17
Simplify links

csillag:


src/app/components/Account/AccountLink.tsx
```diff
@@ -6,16 +6,26 @@ import Link from '@mui/material/Link'
...
   const isMobile = useMediaQuery(theme.breakpoints.down('sm'))
   const to = RouteUtils.getAccountRoute(address, layer)
-  return isMobile ? (
-    <TrimLinkLabel label={address} to={to} />
-  ) : (
-    <Link component={RouterLink} to={to}>
-      {address}
-    </Link>
+  return (
+    <Typography variant="mono" component="span" sx={{ color: COLORS.brandDark, fontWeight: 700 }}>
```
lukaw3d, line 21: Do we no longer need `sx={{ color: COLORS.brandDark, fontWeight: 700 }}`? BlockLink doesn't have it

lukaw3d (approved):
Nice

src/app/components/Account/AccountLink.tsx
csillag: I am not sure, I just copied the code as it was used in the current code-base.

I agree that it would probably make sense to make them more similar to each other.

PR #229, opened 2023-03-20
CopyToClipboard update

buberdds:
Built on top of https://github.com/oasisprotocol/explorer/pull/225

Based on discussion only an icon triggers copy to clipboard action and we want create links instead labels

Also fixes css issue in Account Details. Avatar row now has the same height as all other rows in a list.


lubej (approved):
LGTM

PR #230, opened 2023-03-20
Api sync

buberdds:
- show nodes number even if it's 0
- both total_sent and total_received are now available in API


csillag (approved):
Lgtm.

Issue #169, opened 2023-02-28
How to find the same transaction in other explorers?


lukaw3d:
e.g. https://lw-oasis-explorer-frontend.pages.dev/emerald/transactions/1ada4eb14e50d7c5426f707b6d636e2222280308f09356c1cb2c14503fe927e3

Issue is that Oasis Network Explorer currently uses tx.hash, not tx.eth_hash like other explorers

Andrew7234:
https://github.com/oasisprotocol/oasis-indexer/pull/315 added support for looking up runtime transactions by either oasis hash or eth hash, does this help?

mitjat:
In case it matters: the indexer endpoint https://index-staging.oasislabs.com/v1/emerald/transactions/:hash works for both Oasis and eth hashes.

lukaw3d:
Yaay, that's finally deployed!

PR #239, opened 2023-03-24
In TX details, display ETH hash (if available)

csillag:
Fixes https://github.com/oasisprotocol/explorer/issues/169

### DIFF
--- a/src/app/pages/TransactionDetailPage/index.tsx
+++ b/src/app/pages/TransactionDetailPage/index.tsx
@@ -142,8 +142,8 @@ export const TransactionDetailView: FC<{
           )}
           <dt>{t('common.hash')}</dt>
           <dd>
-            <TransactionLink hash={transaction.hash} layer={transaction.layer} />
-            <CopyToClipboard value={transaction.hash} />
+            <TransactionLink hash={transaction.eth_hash || transaction.hash} layer={transaction.layer} />
+            <CopyToClipboard value={transaction.eth_hash || transaction.hash} />
           </dd>
 
           <dt>{t('common.status')}</dt>
--- a/src/app/components/Transactions/LogEvent.tsx
+++ b/src/app/components/Transactions/LogEvent.tsx
@@ -117,6 +117,8 @@ export const TransactionLogEvent: FC<{ layer: Layer; event: RuntimeEvent; isFirs
   const isMobile = useMediaQuery(theme.breakpoints.down('sm'))
   const decoded = true // TODO: how do I know if this has been successfully decoded?
 
+  // TODO: also consider tx_eth_hash when https://github.com/oasisprotocol/oasis-indexer/issues/363 is resolved
+
   return (
     <>
       {!isFirst && <Divider variant={'card'} />}
--- a/src/app/components/Transactions/index.tsx
+++ b/src/app/components/Transactions/index.tsx
@@ -73,7 +73,13 @@ export const Transactions: FC<TransactionProps> = ({
         key: 'success',
       },
       {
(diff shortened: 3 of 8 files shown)


src/app/pages/TransactionDetailPage/index.tsx
```diff
@@ -145,6 +145,15 @@ export const TransactionDetailView: FC<{
             <TransactionLink hash={transaction.hash} layer={transaction.layer} />
             <CopyToClipboard value={transaction.hash} />
           </dd>
+          {transaction.eth_hash && (
+            <>
+              <dt>{t('common.ethHash')}</dt>
```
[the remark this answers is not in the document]
csillag: Done. One difficulty: when loading events from the API ( via the `/paratime/events` ), we only get a single `tx_hash`, which is an Oasis hash not an ETH hash, so we need to use it for linking.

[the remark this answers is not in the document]
lukaw3d: Pain

src/app/components/Transactions/TransactionLink.tsx
```diff
@@ -8,22 +8,27 @@ import useMediaQuery from '@mui/material/useMediaQuery'
...
+  eth_hash: string | undefined
+}> = ({ alwaysTrim, hash, eth_hash, layer }) => {
   const theme = useTheme()
   const isMobile = useMediaQuery(theme.breakpoints.down('sm'))
-  const to = RouteUtils.getTransactionRoute(hash, layer)
+  if (!hash && !eth_hash) {
+    throw new Error('Either HASH or ETH_HASH must be provided!')
+  }
+  const to = RouteUtils.getTransactionRoute(eth_hash || hash!, layer)
+  const label = eth_hash ? `0x${eth_hash}` : hash!
```
lukaw3d, line 23: I think even links should contain 0x prefix for evm addresses.

(so maybe we should add 0x prefix in api.ts, instead of everywhere it is used)

src/app/components/Transactions/TransactionLink.tsx
```diff
@@ -8,22 +8,27 @@ import useMediaQuery from '@mui/material/useMediaQuery'
 import { TrimLinkLabel } from '../TrimLinkLabel'
```
lukaw3d, line 8: Also 
https://github.com/oasisprotocol/explorer/blob/da4bde6ad7aed9b0e4cc5164e1f3c7a902ed2d95/src/app/pages/SearchResultsPage/useRedirectIfSingleResult.ts#L22-L25

https://github.com/oasisprotocol/explorer/blob/da4bde6ad7aed9b0e4cc5164e1f3c7a902ed2d95/src/app/pages/SearchResultsPage/index.tsx#L132

src/app/components/Transactions/TransactionLink.tsx
csillag: OK, if we want `0x` in the URL, then it changes everything. I will move the transformation into the API layer....

csillag: done

src/app/components/Transactions/TransactionLink.tsx
csillag: OK, no we are handling search, too.

csillag:
There is one more thing we should consider. When displaying TX details, we can say "Hash" or "ETH hash" based on what data we find. But how to handle tables? In the header, we currently say "Hash", even if for some (or all) of the rows will have and ETH hash instead. Is this acceptable? Can we simply call both Oasis and ETH hashes "hash"? If yes, then maybe we shouldn't even try to differentiate in TX details, either..

lukaw3d (approved):
I think yes

PR #245, opened 2023-03-27
Update Rose price display, following design change

csillag:

PR #247, opened 2023-03-27
Adjust the vertical layout of some dashboard cards

csillag:
As requested by @donouwens.

This depends on #245.

Before:

image

After:

image


src/app/components/logo/SmallLogo.tsx
```diff
@@ -0,0 +1,7 @@
+import * as React from 'react'
```
buberdds, line 1: remove

src/app/components/logo/SmallLogo.tsx
```diff
@@ -0,0 +1,7 @@
...
+import { FC } from 'react'
+import logo from './images/logo.png'
+
+export const SmallLogo: FC<{ size?: number }> = ({ size = 25 }) => (
+  <img src={logo} alt={'logo'} height={`${size}px`} width={`${size}px`} />
```
buberdds, line 6: - HTML height and width attrs should be defined as integer without a unit.
- alt value does not make sense + it should be taken from i18next

src/app/pages/DashboardPage/SnapshotCard.tsx
```diff
@@ -1,4 +1,4 @@
-import { FC, PropsWithChildren, ReactNode } from 'react'
+import React, { FC, PropsWithChildren, ReactNode } from 'react'
```
buberdds, line 1: no need to import React

src/app/pages/DashboardPage/SnapshotCard.tsx
```diff
@@ -27,31 +27,33 @@ const StyledCardContent = styled(CardContent)(({ theme }) => ({
 type SnapshotCardProps = PropsWithChildren & {
   badge?: ReactNode
   label?: string
-  title: string
+  title: React.ReactNode
```
buberdds, line 30: ReactNode is already imported

src/app/pages/DashboardPage/SnapshotCard.tsx
```diff
@@ -35,23 +35,25 @@ export const SnapshotCard: FC<SnapshotCardProps> = ({ badge, children, title, la
...
-            pb: 3,
-          }}
-        >
-          <Box>{badge}</Box>
-          <Typography variant="h2" sx={{ fontWeight: 'fontWeightRegular' }}>
-            {label}
-          </Typography>
-        </Box>
-      </CardActions>
+      {(badge || label) && (
```
csillag, line 38: The idea is that if there is neither a badge, nor a label, so the whole bottom section is empty, then don't even render these elements, in order to save space for the main (middle) section, where the children go.

PR #250, opened 2023-03-29
Make sure to always scroll to transactions when requested

csillag:
Task is [here](https://app.clickup.com/t/862jbmukx).

You can test the behavior [here](https://csillag-late-scrolling.oasis-explorer.pages.dev/emerald/blocks/1396255#transactions).

(Suggestion: make then window smaller, so that scrolling is required to see the transactions.)

Also, please note that reload can behave differently compared to the initial loading (i.e. when opening on a new tab), so test both.

### DIFF
--- a/src/app/components/PageLayout/ScrollingCard.tsx
+++ b/src/app/components/PageLayout/ScrollingCard.tsx
@@ -1,0 +1,16 @@
+import Card from '@mui/material/Card'
+import { FC } from 'react'
+import { ScrollingDiv } from './ScrollingDiv'
+
+/**
+ * Native scrolling to an anchor name doesn't work if element is not rendered yet (e.g. after reload).
+ * This checks if id matches anchor name in URL fragment and scrolls to it when component is mounted.
+ */
+export const ScrollingCard: FC<Parameters<typeof Card>[0] & { id: string }> = props => {
+  const { id, ...rest } = props
+  return (
+    <ScrollingDiv id={id}>
+      <Card {...rest} />
+    </ScrollingDiv>
+  )
+}
--- a/src/app/components/PageLayout/ScrollingDiv.tsx
+++ b/src/app/components/PageLayout/ScrollingDiv.tsx
@@ -1,0 +1,29 @@
+import { FC, useEffect, useRef } from 'react'
+import { useLocation } from 'react-router-dom'
+
+export const ScrollingDiv: FC<JSX.IntrinsicElements['div'] & { id: string }> = props => {
+  const { id } = props
+  const { hash } = useLocation()
+  const divRef = useRef<HTMLDivElement>(null)
+  const element = divRef.current
+
+  useEffect(() => {
+    if (hash === `#${id}`) {
+      if (element) {
+        element.scrollIntoView({
+          block: 'start',
+        })
+      } else {
+        setTimeout(() => {
+          if (divRef.current) {
+            divRef.current.scrollIntoView({
+              block: 'start',
+            })
+          }
+        })
+      }
+    }
+  }, [id, hash, element])
+
+  return <div {...props} ref={divRef} />
+}
--- a/src/app/pages/BlockDetailPage/TransactionsCard.tsx
+++ b/src/app/pages/BlockDetailPage/TransactionsCard.tsx
@@ -1,6 +1,6 @@
 import { FC } from 'react'
 import { useTranslation } from 'react-i18next'
-import Card from '@mui/material/Card'
+import { ScrollingCard } from '../../components/PageLayout/ScrollingCard'
 import CardHeader from '@mui/material/CardHeader'
 import CardContent from '@mui/material/CardContent'
 
@@ -47,13 +47,13 @@ const TransactionList: FC<{ layer: Layer; blockHeight: number }> = ({ layer, blo
 export const TransactionsCard: FC<{ layer: Layer; blockHeight: number }> = ({ layer, blockHeight }) => {
   const { t } = useTranslation()
   return (
-    <Card id={transactionsContainerId}>
+    <ScrollingCard id={transactionsContainerId}>
       <CardHeader disableTypography component="h3" title={t('common.transactions')} />
       <CardContent>
         <ErrorBoundary light={true}>
           <TransactionList layer={layer} blockHeight={blockHeight} />
         </ErrorBoundary>
       </CardContent>
-    </Card>
+    </ScrollingCard>
   )
 }


src/app/components/PageLayout/ScrollingCard.tsx
```diff
@@ -0,0 +1,20 @@
...
+export const ScrollingCard: FC<Parameters<typeof Card>[0]> = props => {
+  const { id } = props
+  const { hash } = useLocation()
+  const cardRef = useRef<HTMLDivElement>(null)
+  const card = cardRef.current
+
+  useEffect(() => {
+    if (id && hash === `#${id}` && card) {
+      card.scrollIntoView({
+        behavior: 'smooth',
```
csillag, line 14: This line might be unnecessary after https://github.com/oasisprotocol/explorer/pull/251 is merged.

src/app/components/PageLayout/ScrollingCard.tsx
```diff
@@ -0,0 +1,20 @@
+import Card from '@mui/material/Card'
+import { FC, useEffect, useRef } from 'react'
+import { useLocation } from 'react-router-dom'
+
+export const ScrollingCard: FC<Parameters<typeof Card>[0]> = props => {
```
lukaw3d, line 5: add some documentation why this is needed

src/app/components/PageLayout/ScrollingCard.tsx
```diff
@@ -0,0 +1,20 @@
...
+
+  useEffect(() => {
+    if (id && hash === `#${id}` && card) {
+      card.scrollIntoView({
+        behavior: 'smooth',
+        block: 'start',
+      })
+    }
+  }, [id, hash, card])
+  return <Card {...props} ref={cardRef} />
```
lukaw3d, line 19: Could this be a completely general abstraction, instead of tying it to Card?

src/app/components/PageLayout/ScrollingCard.tsx
csillag: Indeed. Removed.

src/app/components/PageLayout/ScrollingCard.tsx
csillag: OK, added documentation.

src/app/components/PageLayout/ScrollingCard.tsx
csillag: On some level, yes, but 
 - Card is special in the sense that it's automatically creating an HTML anchor based on the ID. So it has a nice, compact configuration surface, easy to interface with. The generic interface would be be different, not immediately usable with Card.
 - Our UI is mostly divided to Cards, so when we want to scroll somewhere, there is a good chance that it will be a card.

csillag: Re-implemented as a hook, independent of card. (And then re-created ScrollingCard using the hook.)

src/app/components/PageLayout/ScrollingCard.tsx
```diff
@@ -0,0 +1,19 @@
...
+ *
+ * If a Card has an "id" prop, it will add an HTML anchor to the DIV.
+ * It can be added to the URL, and we can scroll to this div.
+ *
+ * However, when loading the page initially, sometimes the HTML element for the card is not yet available in the DOM
+ * when the scrolling is supposed to happen, so it fails. This wrapper component can fix that.
+ *
+ * When the component is mounted, it will check (with reach router) if we are supposed to be scrolled to this element,
+ * and executes the scrolling when necessary.
+ */
```
lukaw3d, line 16: I would shorten to e.g.:
> Native scrolling to an anchor name doesn't work if element is not rendered yet (e.g. after reload). This checks if `id` matches anchor name in URL fragment and scrolls to it when component is mounted.

src/app/components/PageLayout/ScrollingCard.tsx
```diff
@@ -0,0 +1,19 @@
...
+export const ScrollingCard: FC<Parameters<typeof Card>[0]> = props => (
```
lukaw3d, line 17: Is `Parameters<typeof Card>[0]` the same as `CardProps`?

lukaw3d: id prop is not required

src/app/components/PageLayout/ScrollingCard.tsx
```diff
@@ -0,0 +1,31 @@
...
+/**
+ * This is a wrapper around MUI card that take care of (fixes) automatic scrolling.
+ *
+ * If a Card has an "id" prop, it will add an HTML anchor to the DIV.
+ * It can be added to the URL, and we can scroll to this div.
+ *
+ * However, when loading the page initially, sometimes the HTML element for the card is not yet available in the DOM
+ * when the scrolling is supposed to happen, so it fails. This wrapper component can fix that.
+ *
+ * When the component is mounted, it will check (with reach router) if we are supposed to be scrolled to this element,
```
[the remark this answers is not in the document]
csillag: OK

src/app/components/PageLayout/ScrollingCard.tsx
csillag: OK

src/app/components/PageLayout/ScrollingCard.tsx
csillag: > id prop is not required

Yes. And the `useScrolling` hook can handle cases when the input parameter (`targetHash`) is undefined. In that case, it won't do anything.

lukaw3d: Alright, tho that feels like a bug to me

src/app/hooks/useScrolling.ts
```diff
@@ -0,0 +1,28 @@
...
+ * React hook that take care of (fixes) automatic scrolling.
+ *
+ * Normally, if an HTML element has an anchor, then it can be added to the URL, and we can scroll to this div.
+ *
+ * However, when loading the page initially, sometimes the HTML element for the card is not yet available in the DOM
+ * when the scrolling is supposed to happen, so it fails. This hook can fix that.
+ *
+ * Use this hook to implement scrolling. See the ScrollingCard component for an example on how to use.
+ */
+export function useScrolling<ElementType extends Element = Element>(targetHash: string | undefined) {
```
lukaw3d, line 14: Should these docs be changed too?

Without looking at ScrollingCard, I would not know this is intended to be passed into `ref=`, but I like the abstraction

csillag:
>  but I like the abstraction

Unfortunately, the abstraction makes the behavior somewhat unpredictable / unreliable, since the order of execution of hooks and effects is different this way. I will have to work on it a bit more.

lukaw3d:
An easy general alternative would be to revert hook, and replace `<Card>` with `<div>`

csillag:
> An easy general alternative would be to revert hook, and replace `<Card>` with `<div>`

I have "fixed" the hook now so that it's reliable, while keeping it as general as possible.
But I don't like the abstraction, the interface is unclean.

Yes, I think I will go with just one wrapper component, no hooks, and it will be a div.

csillag:
> An easy general alternative would be to revert hook, and replace `<Card>` with `<div>`

Yes, I did that. Merging now.

lukaw3d:
Are there any other places we should handle this

PR #251, opened 2023-03-29
Use smooth scrolling everywhere

csillag:
Solves [this task](https://app.clickup.com/t/862jbmuhy).

### DIFF
--- a/src/styles/index.css
+++ b/src/styles/index.css
@@ -1,6 +1,10 @@
 @import '../../node_modules/@fontsource/figtree/variable.css';
 @import '../../node_modules/@fontsource/roboto-mono/variable.css';
 
+html {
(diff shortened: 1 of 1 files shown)


src/styles/index.css
```diff
@@ -19,3 +19,7 @@ svg.recharts-surface {
 .recharts-cartesian-grid-horizontal line:last-child {
   stroke-opacity: 0;
 }
+
+html {
```
lukaw3d, line 23: (i'd prefer if this appeared higher in the file)

lukaw3d (approved):

src/styles/index.css
csillag: OK, moved to the top (after the imports)

PR #252, opened 2023-03-29
Add transactions link to account details

csillag:
On the account detail pages, "transactions" should be a link, just like it is on the block details page.

~This depends on #250.~


lukaw3d (approved):
Nice!

PR #268, opened 2023-04-11
Follow API changes

csillag:
This builds on top of #267 , and fixes that's broken.

A notable new feature is that TX status can now also be unknown. @donouwens has created new designs for that, and this PR also incorporates that.

### DIFF
--- a/src/app/components/TransactionStatusIcon/index.tsx
+++ b/src/app/components/TransactionStatusIcon/index.tsx
@@ -1,48 +1,80 @@
-import { FC } from 'react'
+import React, { FC, ReactNode } from 'react'
 import { useTranslation } from 'react-i18next'
 import Box from '@mui/material/Box'
 import CheckCircleIcon from '@mui/icons-material/CheckCircle'
 import CancelIcon from '@mui/icons-material/Cancel'
 import { styled } from '@mui/material/styles'
 import { COLORS } from '../../../styles/theme/colors'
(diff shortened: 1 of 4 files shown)


csillag:
Here is two screenshots of the new "unknown" state:

In the details view:

image

On transaction list:

image


Design feedback from @donouwens:

>  All in all looks good to me!

src/app/components/TransactionStatusIcon/index.tsx
```diff
@@ -1,48 +1,80 @@
...
+  unknown: COLORS.grayMedium,
+  success: COLORS.eucalyptus,
+  failure: COLORS.errorIndicatorBackground,
+}
+
+const statusIcon: Record<TxStatus, () => ReactNode> = {
+  unknown: () => <Help color={'inherit'} fontSize="inherit" />,
+  success: () => <CheckCircleIcon color="success" fontSize="inherit" />,
+  failure: () => <CancelIcon color="error" fontSize="inherit" />,
+}
```
lukaw3d, line 28: Is there a benefit to generating these every time?

csillag:
This is so old, I will just close and re-create it.

csillag:
All comments from here has been addressed in the follow-up PR, #340.

Except this one:

> RuntimeTransactionMethod is no longer a good name

Does anyone have a better suggestion?

PR #333, opened 2023-04-25
Format all numbers with BigNumber

csillag:
Not handled:
 - Currency amounts
 - Percentages
 - Anything inside EVM logs (will handle in separate PR)
 - Labels on the Y axis of line charts

### DIFF
--- a/src/app/components/Account/index.tsx
+++ b/src/app/components/Account/index.tsx
@@ -17,6 +17,7 @@ import { AccountLink } from './AccountLink'
 import { RouteUtils } from '../../utils/route-utils'
 import { accountTransactionsContainerId } from '../../pages/AccountDetailsPage/TransactionsCard'
 import Link from '@mui/material/Link'
+import { useFormatNumber } from '../../hooks/useNumberFormatter'
 
 export const StyledAvatarContainer = styled('dt')(({ theme }) => ({
   '&&': {
@@ -53,6 +54,7 @@ type AccountProps = {
 
 export const Account: FC<AccountProps> = ({ account, isLoading, roseFiatValue, showLayer }) => {
   const { t } = useTranslation()
+  const formatNumber = useFormatNumber()
   const theme = useTheme()
   const isMobile = useMediaQuery(theme.breakpoints.down('sm'))
   const balance = account?.balances[0]?.balance ?? '0'
@@ -108,7 +110,7 @@ export const Account: FC<AccountProps> = ({ account, isLoading, roseFiatValue, s
           <dt>{t('common.transactions')}</dt>
           <dd>
             <Link component={RouterLink} to={transactionsAnchor!}>
-              {t('common.transactionsNumber', { count: account.stats.num_txns })}
+              {formatNumber(account.stats.num_txns, { countKey: 'common.transactionsNumber' })}
             </Link>
           </dd>
 
--- a/src/app/hooks/useNumberFormatter.ts
+++ b/src/app/hooks/useNumberFormatter.ts
@@ -1,0 +1,60 @@
+import { BigNumber } from 'bignumber.js'
+import { useTranslation } from 'react-i18next'
+
+export type NumberFormattingParameters = Partial<BigNumber.Format> & {
+  // Additional features which are not natively supported by BigNumber
+
+  decimalPlaces?: number
+  maximumFractionDigits?: number
+  roundingMode?: BigNumber.RoundingMode
+  unit?: string
+  countKey?: string
+}
+
+export const useFormatNumber = () => {
+  const { t } = useTranslation()
+  return (
+    inputNumber: number | string | BigNumber.Instance | undefined,
+    format: NumberFormattingParameters = {},
+  ): string | undefined => {
+    if (inputNumber === undefined) return
+    const { decimalPlaces, maximumFractionDigits, roundingMode, unit, countKey, ...formatting } = format
+    if (!!unit && !!countKey) {
+      throw new Error("Please don't try to use unit and countKey together! They are incompatible.")
+    }
+    let number =
+      typeof inputNumber === 'number'
+        ? BigNumber(inputNumber.toString(2), 2) // This is required to keep all precision
+        : BigNumber(inputNumber)
+    if (maximumFractionDigits !== undefined) {
+      number = BigNumber(number.toFixed(maximumFractionDigits, roundingMode))
+    }
+    const wantedFormat = { ...BigNumber.config().FORMAT, ...formatting }
+    const numberString =
+      decimalPlaces === undefined
+        ? number.toFormat(wantedFormat)
+        : roundingMode === undefined
+        ? number.toFormat(decimalPlaces, wantedFormat)
+        : number.toFormat(decimalPlaces, roundingMode, wantedFormat)
+    if (unit) {
+      const formattedUnit = Intl.NumberFormat(undefined, {
+        style: 'unit',
+        unit,
+        unitDisplay: 'long',
+      })
+        .formatToParts(number.toNumber())
+        .find(p => p.type === 'unit')!.value
+      return `${numberString} ${formattedUnit}`
+    } else if (countKey) {
+      const num: number = number.toNumber()
+      if (num === 1) {
+        return t(countKey as any, { count: 1 })
+      } else {
+        const i18nForm = t('common.number', { value: num })
+        return t(countKey as any, { count: num }).replace(i18nForm, numberString)
+      }
+    } else {
+      return numberString
+    }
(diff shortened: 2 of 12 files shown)


src/app/utils/numberFormatter.ts
```diff
@@ -0,0 +1,42 @@
...
+
+  decimalPlaces?: number
+  maximumFractionDigits?: number
+  roundingMode?: BigNumber.RoundingMode
+  unit?: string
+}
+
+export const formatNumber = (
+  inputNumber: number | string | BigNumber.Instance | undefined,
+  format: NumberFormattingParameters = {},
```
lukaw3d, line 14: If the future we'll probably need to add formatting based on locale.

Some approaches for that:
- Param: `formatNumber(111, { useTranslation() })`
- Hook: `const formatNumber = useFormatNumber()` and internally use useTranslation
- i18n: Does i18n have a way to expand `{{value, number}}` with custom `{{value, bignumber}}`

src/app/components/Account/index.tsx
```diff
@@ -108,7 +109,7 @@ export const Account: FC<AccountProps> = ({ account, isLoading, roseFiatValue, s
           <dt>{t('common.transactions')}</dt>
           <dd>
             <Link component={RouterLink} to={transactionsAnchor!}>
-              {t('common.transactionsNumber', { count: account.stats.num_txns })}
+              { formatNumber(account.stats.num_txns) }
```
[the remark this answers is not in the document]
csillag: OK

src/app/utils/numberFormatter.ts
```diff
@@ -0,0 +1,42 @@
...
+): string | undefined => {
+  if (inputNumber === undefined) return
+  const { decimalPlaces, maximumFractionDigits, roundingMode, unit, ...formatting } = format
+  let number =
+    typeof inputNumber === 'number'
+      ? BigNumber(inputNumber.toString(2), 2) // This is required to keep all precision
```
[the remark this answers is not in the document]
csillag: Read [this](https://github.com/MikeMcl/bignumber.js/blob/d87c6606f905668445b1f5782f94adf8c02c4618/README.md?plain=1#L97) and [this](https://github.com/MikeMcl/bignumber.js/blob/d87c6606f905668445b1f5782f94adf8c02c4618/README.md?plain=1#L115) together.

> If the limited precision of Number values is not well understood, it is recommended to create BigNumbers from String values rather than Number values to avoid a potential loss of precision.

> When creating a BigNumber from a Number, note that a BigNumber is created from a Number's decimal `toString()` value not from its underlying binary value. If the latter is required, then pass the Number's `toString(2)` value and specify base 2.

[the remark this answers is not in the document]
csillag: So, to summarize: according to the docs, in order to get all the data from a Number, with no loss of precision, we need to create a binary string representation. Otherwise it will just create a string (with whatever precision) and that will be the input of BigNumber.

src/app/utils/numberFormatter.ts
csillag: OK, using a hook, for now.

src/locales/en/translation.json
```diff
@@ -34,7 +34,7 @@
     "active": "Active",
     "balance": "Balance",
     "block": "Block",
-    "bytes": "{{value, number}}",
+    "bytes": "{{value}}",
```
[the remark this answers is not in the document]
csillag: Right. Removed.

src/app/utils/numberFormatter.ts
```diff
@@ -0,0 +1,42 @@
...
+    decimalPlaces === undefined
+      ? number.toFormat(wantedFormat)
+      : roundingMode === undefined
+      ? number.toFormat(decimalPlaces, wantedFormat)
+      : number.toFormat(decimalPlaces, roundingMode, wantedFormat)
+  if (unit) {
+    const formattedUnit = number
+      .toNumber()
+      .toLocaleString(undefined, { style: 'unit', unit, unitDisplay: 'long' })
+      .split(' ')
```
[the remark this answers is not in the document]
csillag: OK, improved using `formatToParts()`, as suggested.

csillag:
Finished reacting to the first round of comments. Ready for second round.

csillag:
Fixing the broken test case...

src/app/hooks/useNumberFormatter.ts
```diff
@@ -0,0 +1,58 @@
...
+        unitDisplay: 'long',
+      })
+        .formatToParts(number.toNumber())
+        .find(p => p.type === 'unit')!.value
+      return `${numberString} ${formattedUnit}`
+    } else if (countKey) {
+      if (number.toNumber() === 1) {
+        return t(countKey as any, { count: 1 })
+      } else {
+        return t(countKey as any, { count: 2 }).replace('2', numberString)
```
lukaw3d, line 52: Hm, could be unreliable with some translations `countKey_two: "two"`

lukaw3d: Oh but this doesnt matter

src/app/hooks/useNumberFormatter.ts
```diff
@@ -0,0 +1,58 @@
...
+      }
```
lukaw3d, line 53: This doesn't pluralize well

src/app/utils/numberFormatter.ts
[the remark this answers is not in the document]
lukaw3d: okay it loses precision *differently* 
```ts
BigNumber((10000000000/3).toString(2), 2).toFixed()
'3333333333.33333349227905273438'

BigNumber(10000000000/3).toFixed()
'3333333333.3333335'
```

src/app/hooks/useNumberFormatter.ts
```diff
@@ -0,0 +1,58 @@
...
+    const { decimalPlaces, maximumFractionDigits, roundingMode, unit, countKey, ...formatting } = format
+    if (!!unit && !!countKey) {
+      throw new Error("Please don't try to use unit and countKey together! They are incompatible.")
+    }
+    let number =
+      typeof inputNumber === 'number'
+        ? BigNumber(inputNumber.toString(2), 2) // This is required to keep all precision
+        : BigNumber(inputNumber)
+    if (maximumFractionDigits !== undefined) {
+      number = BigNumber(number.toFixed(maximumFractionDigits, roundingMode))
```
lukaw3d, line 30: how come not `number.decimalPlaces`

src/app/hooks/useNumberFormatter.ts
```diff
@@ -0,0 +1,58 @@
...
+    }
+    const wantedFormat = { ...BigNumber.config().FORMAT, ...formatting }
+    const numberString =
+      decimalPlaces === undefined
+        ? number.toFormat(wantedFormat)
+        : roundingMode === undefined
+        ? number.toFormat(decimalPlaces, wantedFormat)
+        : number.toFormat(decimalPlaces, roundingMode, wantedFormat)
```
lukaw3d, line 38: If bignumber types weren't broken I would modify to:
```ts
  return (
    inputNumber: number | string | BigNumber.Instance | undefined,
    format: NumberFormattingParameters = { roundingMode: BigNumber.ROUND_DOWN },
  ): string | undefined => {
  
    const numberString = number.toFormat(decimalPlaces, roundingMode, wantedFormat)
```

src/app/hooks/useNumberFormatter.ts
lukaw3d: The new solution is
```tsx
const i18nForm = t('common.number', { value: num })
return t(countKey as any, { count: num }).replace(i18nForm, numberString)
```

I've been thinking if there are any solutions that don't involve string manipulation, and I have one:
```tsx
return t(countKey as any, {
  count: number.toNumber(), // To pluralize the translation strings
  value: numberString, // To display in translation strings (without losing precision, and with consistent formatting)
})
```

src/app/hooks/useNumberFormatter.ts
```diff
@@ -0,0 +1,60 @@
...
+      })
+        .formatToParts(number.toNumber())
+        .find(p => p.type === 'unit')!.value
+      return `${numberString} ${formattedUnit}`
+    } else if (countKey) {
+      const num: number = number.toNumber()
+      if (num === 1) {
+        return t(countKey as any, { count: 1 })
+      } else {
+        const i18nForm = t('common.number', { value: num })
```
lukaw3d, line 53: https://github.com/oasisprotocol/explorer/pull/731#discussion_r1262425451
We might be able to simplify a lot

Issue #259, opened 2023-04-03
[Bug]: DecodedLogEvent would print "0" if `event.evm_log_params=[]`


lukaw3d:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### STEPS TO REPRODUCE

`{event.evm_log_params?.length && (` can return "0".

For usual `if (array.length)` I would prefer a lint rule to enforce `if (array.length > 0)`.
But in case of `event.evm_log_params?.length > 0` that's incorrect. It needs
`event.evm_log_params && event.evm_log_params.length > 0`

lukaw3d:
Maybe
https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/jsx-no-leaked-render.md
and https://typescript-eslint.io/rules/strict-boolean-expressions/

PR #334, opened 2023-04-25
Fix handling empty evm_log_params

lukaw3d:
Part of https://github.com/oasisprotocol/explorer/issues/259

### DIFF
--- a/src/app/components/Transactions/LogEvent.tsx
+++ b/src/app/components/Transactions/LogEvent.tsx
@@ -55,7 +55,7 @@ const DecodedLogEvent: FC<{ layer: Layer; event: RuntimeEvent }> = ({ layer, eve
             <dd>
               <b>{event.evm_log_name}</b>
             </dd>
-            {event.evm_log_params?.length && (
+            {event.evm_log_params && event.evm_log_params.length > 0 && (
               <>
                 {/*<dt>Params</dt>*/}
                 <dd>


src/app/components/Transactions/LogEvent.tsx
```diff
@@ -51,7 +51,7 @@ const DecodedLogEvent: FC<{ layer: Layer; event: RuntimeEvent }> = ({ layer, eve
             <dd>
               <b>{event.evm_log_name}</b>
             </dd>
-            {event.evm_log_params?.length && (
```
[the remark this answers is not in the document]
lukaw3d: https://github.com/oasisprotocol/explorer/issues/259
> DecodedLogEvent would print "0" if `event.evm_log_params=[]` because `{event.evm_log_params?.length && (` renders "0".

src/app/components/Transactions/LogEvent.tsx
```diff
@@ -51,7 +51,7 @@ const DecodedLogEvent: FC<{ layer: Layer; event: RuntimeEvent }> = ({ layer, eve
...
+            {event.evm_log_params && event.evm_log_params.length > 0 && (
```
buberdds, line 54: Can we keep optional chaining `event.evm_log_params?.length > 0`  to make it shorter?

buberdds (approved):

src/app/components/Transactions/LogEvent.tsx
lukaw3d: nope, typescript doesn't pass `nullable > number` comparison

PR #340, opened 2023-04-27
Follow API changes

csillag:
This builds on top of #425, and do the necessary updates in the rest of the code.

This is the successor of both the (now obsolete and closed) #243 and #268.

A notable new feature is that TX status can now also be unknown. @donouwens has created new designs for that, and this PR also incorporates that.

### DIFF
--- a/src/app/components/RuntimeTransactionLabel/index.tsx
+++ b/src/app/components/RuntimeTransactionLabel/index.tsx
@@ -2,28 +2,20 @@ import { FC } from 'react'
 import { useTranslation } from 'react-i18next'
 import { TFunction } from 'i18next'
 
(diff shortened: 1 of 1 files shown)


src/locales/en/translation.json
```diff
@@ -157,6 +158,10 @@
   "transactions": {
     "latest": "Latest Transactions",
     "method": {
+      "unavailable": "Malformed transaction, method unavailable",
+      "accounts": {
+        "transfer": "Account transfer"
```
lukaw3d, line 163: I think "Transfer" is better. Or something odd like "Native non-EVM transfer"

lukaw3d (approved):

PR #347, opened 2023-04-28
Eliminate duplicate translations

csillag:


src/locales/en/translation.json
```diff
@@ -9,8 +9,7 @@
     "evmTokens": "EVM tokens",
```
lukaw3d, line 9: I disagree with removing duplicates. We should do the opposite and split reused translation strings by semantic context.

csillag:
OK then.

PR #377, opened 2023-05-11
Build and publish release via GitHub releases

buberdds:
[Sample](https://github.com/buberdds/explorer/releases/tag/v0.0.29 ) output triggered by tag push (fork)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,56 @@
+# NOTE: This name appears in GitHub's Checks API and in workflow's status badge.
+name: release
+
+# Trigger the workflow when:
+on:
+  # A push occurs to one of the matched tags.
+  push:
+    tags:
+      # Pattern that roughly matches Oasis Core's version tags.
+      # For more details on GitHub Actions' pattern match syntax, see:
+      # https://help.github.com/en/actions/automating-your-workflow-with-github-actions/workflow-syntax-for-github-actions#patterns-to-match-branches-and-tags.
+      - 'v[0-9]+.[0-9]+*'
+
+jobs:
+  release:
+    # NOTE: This name appears in GitHub's Checks API.
+    name: prepare-release
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+      - name: Set up Node.js 18
+        uses: actions/setup-node@v3
+        with:
+          node-version: '18.x'
+          cache: yarn
+      - name: Install dependencies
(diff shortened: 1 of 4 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,57 @@
...
+          node-version: '18.x'
+          cache: yarn
+      - name: Install dependencies
+        run: yarn install --frozen-lockfile
+      - name: Build app
+        run: yarn build
+      - name: Create zip file
+        run: |
+          cd build/
+          zip -r ../explorer-build.zip .
```
tjanez, line 36: Let's name the release artifacts `oasis-explorer-<VERSION>.zip`.

.github/workflows/release.yml
```diff
@@ -0,0 +1,57 @@
...
+        run: |
+          echo "VERSION=$(echo ${{ github.ref_name }} | sed 's/^v//')" >> "$GITHUB_OUTPUT"
+      - name: Release
+        uses: softprops/action-gh-release@v1
+        with:
+          files: |
+            explorer-build.zip
+          name: Oasis Explorer ${{ steps.vars.outputs.VERSION }}
+          body: |
+            For a list of changes in this release, see the [Change Log].
```
tjanez, line 50: Let's try to copy the part of the Change Log pertaining to the release here directly so it is more convenient for the users.

tjanez (commented):
Great work!

.github/workflows/release.yml
buberdds: The copy task is not enough. md file created by towncirer used in release notes looks like https://github.com/buberdds/explorer/releases/tag/v0.0.32 

1) Easy solution would be extract version related changes, but also adjust JINJA2 template to
```
- {{ lines[0] }} {{ referenced_issues }}
``` 
to keep entries in one line, but we will hit lint errors in CHANGELOG.md as lines will be too long

2) other option is to call helper script in node which extract version related changes from md file, but also tweak formatting (not sure we will handle all edge cases for formatting tweaks).  
sample output: https://github.com/buberdds/explorer/releases/tag/v1.1.0
sample script: https://github.com/buberdds/explorer/blob/master/internals/scripts/version-changelog.js

3) Keep current generic template

4) Some other JINJA2 workarounds maybe?

tjanez: I like option 2. And the script you wrote for the extraction looks pretty robust.

buberdds: PR updated, sample output https://github.com/buberdds/explorer/releases/tag/v1.2.0

tjanez: Great, thanks!

PR #378, opened 2023-05-11
 Update version based on change log fragments

buberdds:
Instead of python punch tool we could use `yarn version` which bumps version in package.json and tag.

### DIFF
--- a/internals/scripts/changelog.js
+++ b/internals/scripts/changelog.js
@@ -1,0 +1,37 @@
+const execSync = require('child_process').execSync
+const semver = require('semver')
+const glob = require('glob')
+const packageJson = require('../../package.json')
+
+const folderPath = '.changelog/'
+const majorPattern = `${folderPath}'*breaking*.md'`
+const minorPattern = `${folderPath}'*feature*.md'`
+const patchPattern = `${folderPath}*{process,cfg,bugfix,doc,internal,trivial}*.md`
+
+console.log('Assembling Change Log and updating package version')
+
+let version = packageJson.version
+if (glob.sync(majorPattern).length > 0) {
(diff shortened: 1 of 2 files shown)


internals/scripts/changelog.js
```diff
@@ -0,0 +1,37 @@
...
+const semver = require('semver')
+const glob = require('glob')
+const packageJson = require('../../package.json')
+
+const folderPath = '.changelog/'
+const majorPattern = `${folderPath}'*breaking*.md'`
+const minorPattern = `${folderPath}'*feature*.md'`
+const patchPattern = `${folderPath}*{process,cfg,bugfix,doc,internal,trivial}*.md`
+
+console.log('Assembling change log and updating package version')
```
tjanez, line 11:
```suggestion
console.log('Assembling Change Log and updating package version')
```

internals/scripts/changelog.js
```diff
@@ -0,0 +1,37 @@
...
+let version = packageJson.version
+if (glob.sync(majorPattern).length > 0) {
+  console.log('\x1b[31m%s\x1b[0m', 'Warning: This release contains breaking changes.')
+  version = semver.inc(version, 'major')
+} else if (glob.sync(minorPattern).length > 0) {
+  version = semver.inc(version, 'minor')
+} else if (glob.sync(patchPattern).length > 0) {
+  version = semver.inc(version, 'patch')
+} else {
+  console.log('No change log fragments for a release found.')
```
tjanez, line 22:
```suggestion
  console.log('No Change Log fragments found. Aborting release...')
```

tjanez (approved):
> Instead of python punch tool we could use `yarn version` which bumps version in package.json and tag.

Makes sense!

PR #256, opened 2023-04-02
Implement paratime picker

lubej:

PR #395, opened 2023-05-17
Remove ParaTime picker

buberdds:


lubej (approved):

src/app/components/PageLayout/NetworkHeader.tsx
```diff
@@ -20,24 +21,6 @@ const StyledNetworkHeader = styled(Box)(({ theme }) => ({
   padding: 0,
   border: 'none',
   borderRadius: 'none',
```
lukaw3d, line 23: Looking at https://github.com/oasisprotocol/explorer/pull/256, I think `position: 'relative'` above can be reverted too

lukaw3d (approved):
Reverts most of https://github.com/oasisprotocol/explorer/pull/256

src/app/components/PageLayout/NetworkHeader.tsx
buberdds: thx, it looks like I can remove `StyledNetworkHeader` completly.

PR #402, opened 2023-05-19
Include network dynamic param in FE router

csillag:
https://app.clickup.com/t/862ju7u5d

Currently, only the UI is changed,
but the displayed data is always from mainnet.


csillag:
Can potentially be tested together with #404, which can also use data from testnet when required.

csillag:
Rebased, now comes with automatic theme selection.

lukaw3d (approved):

src/index.tsx
```diff
@@ -29,10 +29,12 @@ const router = createBrowserRouter(routes)
 
 const root = ReactDOM.createRoot(document.getElementById('root') as HTMLElement)
 
+const wantedTheme = router.state.location.pathname.startsWith('/testnet') ? testnetTheme : defaultTheme
```
lukaw3d, line 32: This should be moved into a component, and use useParams

csillag: We is kind of a special case, since we are outside the router (and any other react components), and so useParams is not yet available.

But in any case, we have some dedicated components like that, in upcoming PRs.

PR #408, opened 2023-05-22
Search across all networks

csillag:
Task is https://app.clickup.com/t/862ju7v0a.

~This depends on #404.~

Some explanation: with this, we support two kinds of search. 

- One is launched from within a page belonging to a specific network. The URL is like `/mainnet/ember/search?q=whatever`. In this case, we are mainly showing the results from that network. (And maybe some indicators about results elsewhere.)
- Then there is the so-called "global" search, which is what you get if you launch a search right from the opening homepage, without actually entering any paratime first. The URL for this is `/global/search?q=whatever`. In this case, you see all results, grouped by networks of course.

In the URL, there is always something that specifies the flavor of the search. That is either a name of a network (mainnet or testnet), or "global" for global search.

### DIFF
--- a/src/app/pages/SearchResultsPage/ResultsOnNetwork.tsx
+++ b/src/app/pages/SearchResultsPage/ResultsOnNetwork.tsx
@@ -1,0 +1,61 @@
+import { FC } from 'react'
+import { Network } from '../../../types/network'
+import { useTranslation } from 'react-i18next'
+import { ResultsGroupByType } from './ResultsGroupByType'
+import { BlockDetailView } from '../BlockDetailPage'
+import { RouteUtils } from '../../utils/route-utils'
+import { TransactionDetailView } from '../TransactionDetailPage'
+import { AccountDetailsView } from '../AccountDetailsPage'
+import { SearchQueries } from './hooks'
+
+/**
+ * Component for selectively displaying a subset of search results that belongs to a specific network
+ *
+ * It doesn't actually run a search query, but uses existing results.
+ */
+export const ResultsOnNetwork: FC<{
+  network: Network
+  searchQueries: SearchQueries
+  roseFiatValue: number | undefined
+}> = ({ network, searchQueries, roseFiatValue }) => {
+  const { t } = useTranslation()
+  return (
+    <>
+      <ResultsGroupByType
+        title={t('search.results.blocks.title')}
+        results={searchQueries.blockHeight.results.filter(result => result.network === network)}
+        resultComponent={item => <BlockDetailView isLoading={false} block={item} showLayer={true} />}
+        link={item => RouteUtils.getBlockRoute(item.network, item.round, item.layer)}
+        linkLabel={t('search.results.blocks.viewLink')}
+      />
+
+      <ResultsGroupByType
+        title={t('search.results.transactions.title')}
+        results={searchQueries.txHash.results.filter(result => result.network === network)}
+        resultComponent={item => (
+          <TransactionDetailView isLoading={false} transaction={item} showLayer={true} />
+        )}
+        link={item => RouteUtils.getTransactionRoute(item.network, item.eth_hash || item.hash, item.layer)}
+        linkLabel={t('search.results.transactions.viewLink')}
+      />
+
+      <ResultsGroupByType
+        title={t('search.results.accounts.title')}
+        results={[
+          ...(searchQueries.oasisAccount.results ?? []).filter(result => result.network === network),
+          ...(searchQueries.evmBech32Account.results ?? []).filter(result => result.network === network),
+        ]}
+        resultComponent={item => (
+          <AccountDetailsView
+            isLoading={false}
+            account={item}
+            roseFiatValue={roseFiatValue}
+            showLayer={true}
+          />
+        )}
+        link={item => RouteUtils.getAccountRoute(item.network, item.address_eth ?? item.address, item.layer)}
+        linkLabel={t('search.results.accounts.viewLink')}
+      />
+    </>
+  )
+}
--- a/src/app/pages/SearchResultsPage/index.tsx
+++ b/src/app/pages/SearchResultsPage/index.tsx
@@ -1,143 +1,73 @@
-import { FC } from 'react'
+import React, { FC } from 'react'
 import Divider from '@mui/material/Divider'
 import { PageLayout } from '../../components/PageLayout'
 import { useParamSearch } from '../../components/Search/search-utils'
-import { useTranslation } from 'react-i18next'
-import {
-  RuntimeAccount,
-  RuntimeBlock,
-  RuntimeTransaction,
-  useGetRuntimeAccountsAddress,
-  useGetRuntimeBlockByHeight,
-  useGetRuntimeTransactionsTxHash,
-  Runtime,
-} from '../../../oasis-indexer/api'
+import { HasNetwork } from '../../../oasis-indexer/api'
 import { useGetRosePrice } from '../../../coin-gecko/api'
-import { AccountDetailsView } from '../AccountDetailsPage'
-import { BlockDetailView } from '../BlockDetailPage'
-import { TransactionDetailView } from '../TransactionDetailPage'
 import { SubPageCard } from '../../components/SubPageCard'
 import { TextSkeleton } from '../../components/Skeleton'
 import { useRedirectIfSingleResult } from './useRedirectIfSingleResult'
 import { NoResults } from './NoResults'
 import { RouteUtils } from '../../utils/route-utils'
-import { ResultsGroup } from './ResultsGroup'
-import { GlobalNetwork, Network, NetworkOrGlobal } from '../../../types/network'
+import { Network } from '../../../types/network'
 import { useNetworkParam } from '../../hooks/useNetworkParam'
-
-function isDefined<T>(item: T): item is NonNullable<T> {
-  return item != null
-}
-
-type ConditionalResults<T> = { isLoading: boolean; results: T[] }
-export type SearchQueries = {
-  blockHeight: ConditionalResults<RuntimeBlock>
-  txHash: ConditionalResults<RuntimeTransaction>
-  oasisAccount: ConditionalResults<RuntimeAccount>
-  evmBech32Account: ConditionalResults<RuntimeAccount>
-}
-function useBlocksConditionally(
-  network: NetworkOrGlobal,
-  blockHeight: string | undefined,
-): ConditionalResults<RuntimeBlock> {
-  const queries = [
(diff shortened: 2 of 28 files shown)


src/styles/index.css
```diff
@@ -23,3 +23,7 @@ svg.recharts-surface {
 .recharts-cartesian-grid-horizontal line:last-child {
   stroke-opacity: 0;
 }
+
+.foreignNote strong {
+  font-weight: 700;
+}
```
lukaw3d, line 29: This should not be needed. @buberdds do you remember why we have https://github.com/oasisprotocol/explorer/blob/c46707429ec456c4d4badf9c7c2421eaa04cf327/src/styles/theme/defaultTheme.ts#L110

csillag:
Rebased this again, all dependencies has been merged, now this is ready for review.

Please note that there are two known issues which this PR doesn't handle:

- Theme reactivity issues when switching between networks -- solved in #418 (which depends on this one)
- Disabling emerald on testnet -- solved in #411 (which depends on #418 and this one)

src/styles/index.css
csillag: I'm happy to simplify, if we find a good way.

lukaw3d: I would change it here. I only noticed it affected "Type     Transfer" in  https://explorer.dev.oasis.io/mainnet/emerald/transactions/0x2d3f414f402eb83e99a59ae7b08d849e05884f36bacf5e2d2cc8b967566176cf

src/app/pages/SearchResultsPage/ResultsOnNetwork.tsx
```diff
@@ -0,0 +1,52 @@
...
+import { getThemesForNetworks } from '../../../styles/theme'
+import useMediaQuery from '@mui/material/useMediaQuery'
+import { NetworkThemeBubble } from '../../components/NetworkThemeBubble'
+import Box from '@mui/material/Box'
+import { SearchResultsList } from './SearchResultsList'
+import { SearchQueries } from './hooks'
+import Typography from '@mui/material/Typography'
+
+export const ResultsOnNetwork: FC<{
+  network: Network
```
lukaw3d, line 14: Ah, this is network from results, not from search query. Rename? Comment?

src/app/pages/SearchResultsPage/index.tsx
```diff
@@ -148,51 +78,46 @@ export const SearchResultsView: FC<{
...
+            searchQueries={searchQueries}
+            roseFiatValue={roseFiatValue}
           />
         </>
       )}
+      {!isGlobal &&
+        !isAnyLoading &&
+        matchingNetworkList
+          .filter(n => n !== wantedNetwork)
+          .map(n => (
```
lukaw3d, line 97: I'd rename `n` to `net` like below

src/app/pages/SearchResultsPage/ResultsOnNetwork.tsx
```diff
@@ -0,0 +1,52 @@
...
+              pt: 4,
+              px: isMobile ? 0 : '4%',
+              border: isMobile ? 'none' : `solid 15px ${otherTheme.palette.layout.border}`,
+              background: otherTheme.palette.background.default,
+            }}
+          >
+            {content}
+          </Box>
+        </NetworkThemeBubble>
+      )}
```
lukaw3d, line 49: Can this be unified? Redeclaring mainnet theme shouldn't hurt, and differentiating it at theme level seems nicer (unless the border shouldn't be for not-mainnet results, but for foreign results)

src/app/pages/SearchResultsPage/SearchResultsList.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+export const SearchResultsList: FC<{
+  network: Network
+  searchQueries: SearchQueries
+  roseFiatValue: number | undefined
+}> = ({ network, searchQueries, roseFiatValue }) => {
+  const { t } = useTranslation()
+  return (
+    <>
+      <ResultsGroup
```
lukaw3d, line 19: ResultsGroup is now no longer a great name. It groups by result type.
But ResultsOnNetwork and ResultsOnForeignNetwork (and SearchResultsList) now also group by network.

src/app/pages/SearchResultsPage/ResultsOnForeignNetwork.tsx
```diff
@@ -0,0 +1,104 @@
...
+            i18nKey={
+              hasLocalMatches ? 'search.otherResults.clickToShowMore' : 'search.otherResults.clickToShowThem'
+            }
+            values={{
+              countLabel: t(hasLocalMatches ? 'search.results.moreCount' : 'search.results.count', {
+                count: matches,
+              }),
+              networkName,
+            }}
+          />
```
lukaw3d, line 73: image
Wrap this into a `<span>` to fix the spacing

lukaw3d (approved):

src/app/pages/SearchResultsPage/SearchResultsList.tsx
```diff
@@ -0,0 +1,56 @@
...
+        title={t('search.results.blocks.title')}
+        results={searchQueries.blockHeight.results.filter(result => result.network === network)}
```
lukaw3d, line 21: Eww

src/app/components/NetworkThemeBubble/index.tsx
```diff
@@ -0,0 +1,9 @@
+import React, { FC } from 'react'
+import { NetworkOrGlobal } from '../../../types/network'
+import { ThemeProvider } from '@mui/material/styles'
+import { getThemesForNetworks } from '../../../styles/theme'
+
+export const NetworkThemeBubble: FC<{ network: NetworkOrGlobal; children: React.ReactNode }> = ({
```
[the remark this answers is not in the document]
csillag: Done.

src/app/pages/SearchResultsPage/ResultsOnForeignNetwork.tsx
```diff
@@ -0,0 +1,104 @@
...
+  const networkName = getNetworkNames(t)[network]
+  if (!matches) {
+    return null
+  }
+  const otherTheme = getThemesForNetworks()[network]
+
+  if (!open) {
+    return (
+      <NetworkThemeBubble network={network}>
+        <NotificationBox className={'foreignNote'} onClick={() => setOpen(true)}>
```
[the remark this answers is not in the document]
csillag: OK, which component do you suggest we use here, insted of Box?

src/app/pages/SearchResultsPage/index.tsx
csillag: OK

src/app/pages/SearchResultsPage/ResultsOnNetwork.tsx
csillag: OK, adding comments to all similar components.

src/app/pages/SearchResultsPage/SearchResultsList.tsx
csillag: OK, it became `ResultGroupByType`. (Also,
 - `ResultsOnNetwork` --> `ResultsOnNetworkThemed`
 - `ResultsOnForeignNetwork' -> `ResultsOnForeignNetworkThemed`
 - `SearchResultList` -> `ResultsOnNetwork`.

src/app/pages/SearchResultsPage/SearchResultsList.tsx
csillag: Would you please elaborate? Is this about performance, and code readability?

src/styles/index.css
csillag: OK

src/app/pages/SearchResultsPage/ResultsOnNetwork.tsx
csillag: Actually, I had some problem with the theme "leaking out". (Even though the component used to be  called a bubble..) So for now I got rid of the theme around the network results, and just styling the Box.

The theme "leaking out" seems to be a generic issue, not related to this PR, so I think we should investigate it elsewhere.

src/routes.tsx
```diff
@@ -14,72 +14,90 @@ import {
...
       {
         path: '/',
-        element: <HomePage />,
-      },
-      {
-        path: '/:network/search', // ?q=
-        element: <SearchResultsPage />,
-        loader: searchParamLoader,
+        element: <ThemedHomePage />,
       },
```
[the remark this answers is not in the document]
csillag: OK, so it's undoubtedly true that the testnet/mainnet setting (of the home page) should be reflected in the URL, but I think it's more like a query parameter, and not part of the path, since we are not actually going anywhere, just planning to go... will do it separately.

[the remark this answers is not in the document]
lukaw3d: I think it is part of the path, and it should affect useGetStatus, globe colors, globe navigation
(and I would propose that is should affect search too)

src/app/pages/SearchResultsPage/SearchResultsList.tsx
lukaw3d: We have one implementation in a higher layer for counting results, and completely different implementation in this layer for showing the results

lukaw3d: *different implementation of counting/filtering all results

src/app/pages/SearchResultsPage/ResultsOnForeignNetwork.tsx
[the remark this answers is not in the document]
lukaw3d: Probably Collapse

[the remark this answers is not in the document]
lukaw3d: And Button

src/routes.tsx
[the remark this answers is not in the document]
lukaw3d: Okay, apparently it should only affect globe, and we are putting into query parameter now

src/app/pages/SearchResultsPage/SearchResultsList.tsx
lukaw3d: Counting and filtering logic was unified in https://github.com/oasisprotocol/explorer/pull/449/commits/050a9557f6966937a9fbd798b14c9ff2755280c0


https://github.com/oasisprotocol/explorer/pull/449/commits/7e9666708e8275b2fd724a0f20c34f2796306a6b and https://github.com/oasisprotocol/explorer/pull/449/commits/f060f9cb9d901968c16a19aae75b4c754a4867e4 refactored from passing in `all results + filter` to passing `filtered results`.

:tada: very readable

PR #409, opened 2023-05-22
Make EmptySpace / NoResults theme-aware

csillag:
This PR improves how the "no results found" message looks on the TestNet.

It can't be directly tested, unless you also have the changes from #404, but I'm opening this a separate PR to avoid later conflicts with other theming work.

Before:

image

After:

after

@donouwens says that the visual aspects are acceptable.


buberdds (approved):

csillag:
> lint error

Sorry, fixed the lint error.

PR #417, opened 2023-05-22
Fix accessibility of buttons by enabling focus outlines and focus styles

lukaw3d:
Tabbing through top elements:
| Before | After |
| --- | --- |
| before | after |


csillag (approved):
Reviewed. LGTM.

PR #418, opened 2023-05-22
Fix theme reactivity issues

csillag:
This depends on #408. ~(Which depends on #404.)~


csillag:
I am merging this into #408.

PR #421, opened 2023-05-23
Update Orval to latest (for API wrapper code generation)

csillag:
There is a problem around number / bigint in the version of Orval we are currently using, as Luka has explained [here](https://github.com/oasisprotocol/explorer/pull/267#discussion_r1195385202).

This PR updates Orval to the latest release.

(Which I have verified to work.)

Doing this is necessary to get the latest API changes.

### DIFF
--- a/src/oasis-indexer/package.json
+++ b/src/oasis-indexer/package.json
@@ -4,6 +4,6 @@
     "generate": "orval"
   },
   "dependencies": {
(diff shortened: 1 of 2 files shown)


lukaw3d (approved):

csillag:
Ooops, this broke this action. Follow-up fix is in #422

PR #426, opened 2023-05-23
ParaTime picker 

buberdds:
PR covers only desktop version

~~Need to go through this with Don tomorrow + there are some missing descriptions in lang files.~~ This is done


src/app/components/ParaTimePicker/ParaTimeMenu.tsx
```diff
@@ -0,0 +1,100 @@
...
+        }}
+        onClick={enabled ? () => setSelectedLayer(layer) : undefined}
+      >
+        <ListItemText>
+          {labels[layer]}
+          <Typography
+            component="span"
+            sx={{ fontSize: '10px', fontStyle: 'italic', color: COLORS.grayMedium, ml: 2 }}
+          >
+            {activeLayer === layer && t('paraTimePicker.selected')}
```
lukaw3d, line 52: Bug:
- open picker on testnet https://d35c4db7.oasis-explorer.pages.dev/testnet/emerald
- it shows Emerald Mainnet as currently selected
image

src/app/components/ParaTimePicker/ParaTimeMenu.tsx
```diff
@@ -0,0 +1,100 @@
...
+        divider={divider}
+        sx={{
+          cursor: enabled ? 'pointer' : 'default',
+        }}
+        onMouseEnter={() => {
+          if (layer !== selectedLayer) {
+            setSelectedLayer(undefined)
+          }
+          setHoveredLayer(layer)
+        }}
```
lukaw3d, line 43: I get annoyed with this UX:
ux

src/app/components/ParaTimePicker/index.tsx
```diff
@@ -0,0 +1,119 @@
...
+          </Button>
+          <Button
+            onClick={() => onConfirm(selectedNetwork!, selectedLayer!)}
+            disabled={!selectedLayer || !selectedNetwork}
+            color="primary"
+            variant="contained"
+          >
+            {t('common.select')}
+          </Button>
+        </Box>
```
lukaw3d, line 115: (smaller screen layout issue: buttons sometimes overlay the text:
image

src/app/components/ParaTimePicker/index.tsx
```diff
@@ -0,0 +1,119 @@
...
+            />
+          </Grid>
+          <Grid xs={6}>
+            <ParaTimeDetails
+              activeLayer={layer}
+              hoveredLayer={hoveredLayer}
+              selectedLayer={selectedLayer}
+              network={selectedNetwork || network}
+            />
+          </Grid>
```
lukaw3d, line 100: (smaller screen layout issue: with my scrollbars, buttons are sometimes jittery:
button positions

csillag:
I checked this out, and have a few difficulties with with, and suggestions.

1. By default, we only see the level where we can select Paratimes within the network. There is no indication whatsoever that it should be possible to switch between networks. (Even when I am running in TestNet mode!) There is only this little arrow at the top, which is nor clear at all. I think at least a work of explanation would be very useful there.
2. When I click that semi-hidden button, the network selector is revealed, but it only shows the Mainnet, even if I am currently at TestNet! I understand that we don't want to promote TestNet too much, but when we are already on TestNet, what is the point of hiding it's existence? I think it's confusing that the opening state is not synced to the current setting. I think if we are at TestNet, it should shot TestNet selected by default.
3. Why is the currently selected state not highlighted by a different bg color? (Both network and Paratime.)

@donouwens I'm sorry to bring the feedback so late in the game, but I think it's better late then never.

csillag:
- For point 1: Having it collapsed is OK. All I am advocating for is having a single word of explanation besides the arrow.
- For point 3: For me, on my monitor, the difference between the selected and not selected stats is undetectable. About confusion with the hover state: the hover state has both the moving arrow _and_ the background color change. I think we could take one of the two to indicate the selected state. (Or something else, but the current way I just can not see.)

lukaw3d (approved):

csillag:
>  the active state does have a background in the designs. Has it not been implemented correctly? I feel there are significant color differences and visual indicators between default/selected/hover/disabled (which also has the tooltip) - see screenshot.

Screenshot 2023-05-25 at 21 26 43

   * * *

No, it doesn't look that way. This is the current implementation:




So,
 -  In the design, both the lighter shade of highlight, and the arrow signifies the current selection, and the hover is a stronger shade of highlight.
 - However, in the implementation, there is only one shade of highlight, and both it and the arrow serve the hover, and the selection only has the small "(currently active)" flag.

Should be easy to fix, though.

csillag (changes requested):
Please make it so that the selected state is highlighted, like it is in the design. Maybe my eyes  👀  are not what they used to be, but I can hardly see which one is selected..

csillag:
>> When I click that semi-hidden button, the network selector is revealed, but it only shows the Mainnet, even if I am currently at TestNet! I understand that we don't want to promote TestNet too much, but when we are already on TestNet, what is the point of hiding it's existence? I think it's confusing that the opening state is not synced to the current setting. I think if we are at TestNet, it should shot TestNet selected by default.

> This is true - if the user is one Testnet we should show the Testnet icon or Testnet icon with 'Testnet' title.

That part works, we can see the Testnet icon and title, that's not what I'm talking about.
My issue is that when I bring out the network selector, it only shows Mainnet, and hides Testnet, _**which is the current selection**_. 

image

So, to summarize: we are on Testnet, why is testnet not listed as one of the options, and marked as selected?

src/app/components/ParaTimePicker/LayerMenu.tsx
```diff
@@ -0,0 +1,134 @@
...
+  selectedNetwork,
+  setHoveredLayer,
+  setSelectedLayer,
+}) => {
+  const options = Object.values(Layer)
+    .map(layer => ({
+      layer,
+      enabled: RouteUtils.getEnabledLayersForNetwork(selectedNetwork || network).includes(layer),
+    }))
+    .sort((a, b) => menuSortOrder.indexOf(a.layer) - menuSortOrder.indexOf(b.layer))
```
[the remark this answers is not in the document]
buberdds: thx, I will go with 1st proposal, but without negative values

buberdds:
@csillag 
- Network menu is open and expanded for non mainner networks
- hover and selected states updated based on Don's feedback. 

If I did not catch something else from your comment let me know 🖖

csillag:
>     * network menu is open and expanded for non mainnet networks

Yep, that works great now!

>     * hover and selected states updated based on Don's feedback

I'm sorry I still don't see the highlight color on the bg of the selected option.
I'm looking for this: Screenshot 2023-05-25 at 21 26 43

I do see the highlight for hover, but not for selection. In Don's visuals, there were two distinct highlights, one for hover, and one for selection.

buberdds:
I've received this from Don today. I will double check with him again.

1. default
2. hover
3. selected
4. disabled 

Screenshot 2023-05-26 at 08 31 28

buberdds:
Turns out there was an issue with mockups and some colors were wiped out :D Background to selected state added.

csillag:
I think it would make total sense to reassign the arrow for selection type 2 instead of type 1.
It's a nice connection between the left side (the option selected) and the right side (where the option is described).

buberdds:
Here's answer from Don: We have default, selected, disabled, active states. Let’s stick with those. The ‘mouse down’ state Kristof is describing is a split second state between hover and active. We are okay without it.

csillag:
OK, can clear up the dictionary? I am suspecting that we might assing different meaning to some of the terms here.
The state I'm talking about (selection type 2) is definitely not a split second state; it starts when the users clicks on the list item, and exists indefinitely, at least until the user clicks the "select" button at the other side of the dialog, hence closing the dialog.

buberdds:
imo it doesn't make sense to discuss such design changes via code review process. It's better to discuss this during Sync meeting. I will keep PR open until then.

csillag (approved):
LGTM

PR #436, opened 2023-05-24
Never automatically redirect from global search to testnet

csillag:


csillag:
Let's close this for now, since the whole scoping of search is under rethinking anyway. (We can just re-implement this on the finalized search result display code)

PR #437, opened 2023-05-25
Revamp everything according to new understanding of search

csillag:
(This depends on #439)

The main changes are:
- /:network/search has moved to /:network/:paratime/search
- search is actually scoped to work only inside the given paratime, optionally offering results from elsewhere.
- Since now we know that selecting a network and a paratime always goes together (ie. we can never select one without the other), we are calling this combined selection a SearchScope.
- All components that used to expect a network and layer parameter together, now expect a scope parameter instead.

Tests still need to be updated.


csillag:
Closing this in favor of #438.

PR #438, opened 2023-05-25
Revamp everything according to new understanding of search, v2

csillag:
~This is an alternate version of #437.~
This depends on https://github.com/oasisprotocol/explorer/pull/439

The main changes are:

- /:network/search has moved to /:network/:paratime/search
- search is actually scoped to work only inside the given paratime, optionally offering results from elsewhere.
- Other results from other networks and paratimes are offered separately, as discussed.
- Since now we know that selecting a network and a paratime always goes together (ie. we can never select one without the other), we are calling this combined selection a SearchScope.
- All components that used to expect a network and layer parameter together, now expect a scope parameter instead.


src/app/pages/SearchResultsPage/ResultsElsewhere.tsx
```diff
@@ -0,0 +1,144 @@
...
+        <ResultsInNetworkThemed
+          network={wantedScope.network}
+          title={t('search.otherResults.otherParatimesOnNetwork', {
+            network: networkNames[wantedScope.network],
+          })}
+          searchQueries={searchQueries}
+          numberOfResults={resultsInNetworks[wantedScope.network]}
+          roseFiatValue={roseFiatValue}
+        />
+      )}
```
lukaw3d, line 129: - https://1232a66b.oasis-explorer.pages.dev/mainnet/emerald/search?q=0xBA504818FdD8D3dBA2Ef8fD9B4F4D5c71aD1d1D3
- expand more results
- see the same Emerald Mainnet account listed again

1232a66b oasis-explorer pages dev_mainnet_emerald_search_q=0xBA504818FdD8D3dBA2Ef8fD9B4F4D5c71aD1d1D3

csillag: Just pushed a big simplification and code removal. Should take care of that. too.

csillag: Yes, now solved.

csillag:
Half of this has been merged; closing in favor of #449.

PR #439, opened 2023-05-25
Simplify query generation

csillag:
Instead of manually enumerating all possible combinations of networks and runtimes, creating queries for them, and then checking if they are enabled, we can just compile the list of enabled combinations, and generated queries just for them.

Doing this simplifies our code. (And now we can add more networks and runtimes without manually having to touch this code.)

Normally, calling React hooks from callbacks and other conditional code is not a good idea, but in this case, we can be sure that the number of enabled combinations will never change during runtime (since it's hard-coded in route-utils.ts), su we can just ignore the TS warning about abusing the rules of hooks.

### DIFF
--- a/src/app/pages/SearchResultsPage/hooks.ts
+++ b/src/app/pages/SearchResultsPage/hooks.ts
@@ -6,8 +6,8 @@ import {
   useGetRuntimeBlockByHeight,
   useGetRuntimeTransactionsTxHash,
   Runtime,
+  Layer,
 } from '../../../oasis-indexer/api'
-import { Network } from '../../../types/network'
 import { RouteUtils } from '../../utils/route-utils'
 
 function isDefined<T>(item: T): item is NonNullable<T> {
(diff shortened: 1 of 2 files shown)


src/app/pages/SearchResultsPage/hooks.ts
```diff
@@ -23,32 +23,16 @@ export type SearchQueries = {
...
-      query: {
-        enabled:
-          !!blockHeight && RouteUtils.getEnabledLayersForNetwork(Network.testnet).includes(Runtime.sapphire),
-      },
-    }),
-  ]
+  const queries = RouteUtils.getEnabledScopes()
+    .filter(scope => scope.layer !== Layer.consensus)
+    .map(scope =>
+      // eslint-disable-next-line react-hooks/rules-of-hooks
```
lukaw3d, line 29: This seems very risky. Are you sure this is fine?

lukaw3d (approved):
Alright

src/app/utils/route-utils.ts
```diff
@@ -72,6 +72,12 @@ export abstract class RouteUtils {
     return RouteUtils.ENABLED_LAYERS_FOR_NETWORK[network] || []
   }
 
+  static getEnabledScopes(): { network: Network; layer: Layer }[] {
+    return RouteUtils.getEnabledNetworks().flatMap(network =>
```
lukaw3d, line 76: Can we have a readonly const at the top of the file instead

src/app/pages/SearchResultsPage/hooks.ts
```diff
@@ -89,32 +59,17 @@ export function useTransactionsConditionally(
...
-      query: {
-        enabled:
-          !!address && RouteUtils.getEnabledLayersForNetwork(Network.testnet).includes(Runtime.sapphire),
-      },
-    }),
-  ]
+  const queries = RouteUtils.getEnabledScopes()
+    .filter(scope => scope.layer !== Layer.consensus)
+    .map(scope =>
+      // eslint-disable-next-line react-hooks/rules-of-hooks
```
lukaw3d, line 65: `// eslint-disable-next-line react-hooks/rules-of-hooks -- copy an explanation`

csillag: OK

src/app/utils/route-utils.ts
csillag: Yes, but this part is already being touched by the big search overhaul PR, so I'll do it there instead.

Issue #356, opened 2023-05-03
[Bug]: If API is broken we only detect it on homepage


lukaw3d:
### URGENCY

- [ ] High
- [ ] Medium
- [X] Low
- [ ] I don't know

### DESCRIPTION OF ISSUE

| Home Page | Emerald dashboard |
| --- | --- |
| ![oasis-explorer pages dev_](https://user-images.githubusercontent.com/3758846/235970892-437547cb-4a2e-4ca0-a209-a4eef180472f.png) | ![oasis-explorer pages dev_ (1)](https://user-images.githubusercontent.com/3758846/235970895-fcc3e250-1b4d-4ed3-80b0-47ee40c3db6b.png) |

### DEVICE / VIEWPORT

- [X] Desktop
- [X] Mobile

### STEPS TO REPRODUCE

Open explorer with intentionally broken API: https://9543da24.oasis-explorer.pages.dev/emerald (deployed from https://github.com/oasisprotocol/explorer/commit/0a22166922b9fcb0235d203e8ae68ba9d0f736fe)

lukaw3d:
Yes, thanks!
TODO: implement to match the mocks

lukaw3d:
TODO: should we use getRuntimeStatus on homepage too?

csillag:
> TODO: should we use getRuntimeStatus on homepage too?

Do you mean the Dashboards?

csillag:
TODO: make sure to track the status of the correct deployment (TestNet vs MainNet)

lukaw3d:
> 

That too;
I think Homepage should probably check:
- getStatus(mainnet /* consensus */)
- getRuntimeStatus(mainnet, emerald)
- getRuntimeStatus(mainnet, sapphire)

Mainnet Emerald Dashboard should check:
- getRuntimeStatus(mainnet, emerald)

csillag:
See #462.

csillag:
This has already been implemented. Closing.

PR #440, opened 2023-05-25
Add missing wrappers for API calls and block accidental mainnet fallback

csillag:
Since we need to be able to override the URL for testnet, and this functionality is currently only implemented in our manual API wrappers, we need manual wrappers for all API functions we want to use.

We already had them for most, but a few were missing, and thus those calls always went to mainnet, instead of testnet.

This change fixes that by adding the missing wrappers, and updating the app code to always add the network param.

**Update:** it also makes it impossible to accidentally use a wrapper-less API call in the future.

### DIFF
--- a/src/app/pages/HomePage/index.tsx
+++ b/src/app/pages/HomePage/index.tsx
@@ -107,7 +107,8 @@ export const HomePage: FC = () => {
   const { t } = useTranslation()
   const infoAriaLabel = t('home.helpScreen.infoIconAria')
   const isMobile = useMediaQuery(theme.breakpoints.down('sm'))
-  const apiStatusQuery = useGetStatus()
+  const { network } = useSearchQueryNetworkParam()
+  const apiStatusQuery = useGetStatus(network)
   const isApiOffline = apiStatusQuery.isFetched && !apiStatusQuery.isSuccess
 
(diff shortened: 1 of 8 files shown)


csillag:
See #442 for an alternative.

src/app/pages/HomePage/index.tsx
```diff
@@ -105,7 +106,7 @@ export const HomePage: FC = () => {
   const { t } = useTranslation()
   const infoAriaLabel = t('home.helpScreen.infoIconAria')
   const isMobile = useMediaQuery(theme.breakpoints.down('sm'))
-  const apiStatusQuery = useGetStatus()
+  const apiStatusQuery = useGetStatus(Network.mainnet) // TODO: what is testnet is off-line? We need a combined status
```
lukaw3d, line 109: typo in the comment

my answer: I'd just check mainnet on homepage. We can cover testnet when user navigates into it https://github.com/oasisprotocol/explorer/issues/356

csillag: Now that we have proper network selection on this screen, I made it so that we are checking the status of the selected network.

lukaw3d (approved):

PR #441, opened 2023-05-25
Add more syntactic sugar around API wrapper calls

csillag:
Until we can replace these manual wrappers with auto-generated code, at least automate as much as possible from the plumbing.


csillag:
Let's just forget this. Not important.

PR #411, opened 2023-05-22
Selectively support different runtimes per network.

csillag:
... since Emerald is not available on TestNet.

PR #443, opened 2023-05-25
Control home page section theme via query param

buberdds:
Home page needs a theme override only in a specific section (graph area).

Now with https://github.com/oasisprotocol/explorer/pull/411 there is more work to do around graph. I will create a ClickUp task to handle that part

### DIFF
--- a/src/app/hooks/useSearchQueryNetworkParam.ts
+++ b/src/app/hooks/useSearchQueryNetworkParam.ts
@@ -1,0 +1,30 @@
+import { useSearchParams } from 'react-router-dom'
+import { Network } from '../../types/network'
+import { RouteUtils } from '../utils/route-utils'
+import { AppErrors } from '../../types/errors'
+
+/**
+ * Get the network from the URL search params and return it if it is valid,
+ * otherwise fallback to mainnet
+ */
+export const useSearchQueryNetworkParam = (): {
+  network: Network
+  setNetwork: (network: Network) => void
+} => {
+  const [searchParams, setSearchParams] = useSearchParams()
+  const networkQueryParam = searchParams.get('network') ?? Network.mainnet
+  if (!RouteUtils.getEnabledNetworks().includes(networkQueryParam as any)) {
+    throw AppErrors.InvalidUrl
+  }
+  return {
+    network: networkQueryParam as Network,
+    setNetwork: network => {
+      if (network === Network.mainnet) {
+        searchParams.delete('network')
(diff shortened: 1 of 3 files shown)


csillag (approved):
Reviewed. LGTM.

(Tweaked it so that the default setting is hidden in the URL.)

csillag (approved):
Reviewed. LGTM.

(Tweaked it so that the default setting is hidden in the URL.)

src/app/hooks/useSearchQueryNetworkParam.ts
```diff
@@ -0,0 +1,29 @@
...
+  }
+  return {
+    network: networkQueryParam as Network,
+    setNetwork: network => {
+      if (network === Network.mainnet) {
+        searchParams.delete('network')
+        setSearchParams(searchParams)
+      } else {
+        setSearchParams({ network })
+      }
```
lukaw3d, line 26: Slightly more accurate:
```suggestion
      if (network === Network.mainnet) {
        searchParams.delete('network')
      } else {
        searchParams.set('network', network)
      }
      setSearchParams(searchParams)
```

(if we ever used it with other query parameters)

lukaw3d (approved):

src/app/hooks/useSearchQueryNetworkParam.ts
```diff
@@ -0,0 +1,29 @@
...
+ */
+export const useSearchQueryNetworkParam = (): {
+  network: Network
+  setNetwork: (network: Network) => void
+} => {
+  const [searchParams, setSearchParams] = useSearchParams()
+  let networkQueryParam = searchParams.get('network')
+  if (!networkQueryParam || !RouteUtils.getEnabledNetworks().includes(networkQueryParam as any)) {
+    networkQueryParam = Network.mainnet
+  }
```
lukaw3d, line 17: Also?
```suggestion
  const networkQueryParam = searchParams.get('network') ?? Network.mainnet
  if (!RouteUtils.getEnabledNetworks().includes(networkQueryParam as any)) {
    throw new Error('Unexpected network')
  }
```

csillag: Yes... but we have a dedicated error type to throw for invalid network, so I will use that.

csillag:
This 40-line PR now has code from 3 devs, _and_ two approvals. I think that's enough, so I squashed and merged it.

PR #447, opened 2023-05-25
Replace network and layer filters with SearchScope

lukaw3d:
Extracted from https://github.com/oasisprotocol/explorer/pull/438


csillag (approved):
LGTM (since I wrote it initially.)
Let's hope it actually works without the reset of the pending changes.

PR #455, opened 2023-05-26
Add option to hide empty tabs on accounts (disabled by default)

csillag:
Actually, most of the work here is about cleaning up the infrastructure, and moving common logic into hooks.
The actual behavior change is just two small commits.

I suggest reviewing commit by commit.

Before: 
image
image

After:

image
image

### DIFF
--- a/src/app/components/RouterTabs/index.tsx
+++ b/src/app/components/RouterTabs/index.tsx
@@ -7,6 +7,7 @@ type RouterTabsProps = {
   tabs: {
     label: string
     to: string
+    visible?: boolean
   }[]
 }
 
@@ -21,9 +22,11 @@ export const RouterTabs: FC<RouterTabsProps> = ({ tabs }) => {
   return (
     <>
       <Tabs value={currentTab?.to}>
-        {tabs.map(tab => (
-          <Tab key={tab.to} component={RouterLink} value={tab.to} label={tab.label} to={tab.to} />
-        ))}
+        {tabs
+          .filter(tab => tab === currentTab || tab.visible !== false)
+          .map(tab => (
+            <Tab key={tab.to} component={RouterLink} value={tab.to} label={tab.label} to={tab.to} />
+          ))}
       </Tabs>
       <Outlet />
     </>
--- a/src/app/pages/AccountDetailsPage/hook.ts
+++ b/src/app/pages/AccountDetailsPage/hook.ts
@@ -1,0 +1,51 @@
+import { Layer, useGetRuntimeAccountsAddress, useGetRuntimeTransactions } from '../../../oasis-indexer/api'
+import { AppErrors } from '../../../types/errors'
+import { useSearchParamsPagination } from '../../components/Table/useSearchParamsPagination'
+import { NUMBER_OF_ITEMS_ON_SEPARATE_PAGE } from '../../config'
+import { SearchScope } from '../../../types/searchScope'
+
+export const useAccount = (scope: SearchScope, address: string) => {
+  const { network, layer } = scope
+  if (layer === Layer.consensus) {
+    // There can be no ERC-20 or ERC-721 tokens on consensus
+    throw AppErrors.UnsupportedLayer
+  }
+  const query = useGetRuntimeAccountsAddress(network, layer, address!)
+  const account = query.data?.data
+  const isLoading = query.isLoading
+  return { account, isLoading }
+}
+
+export const useTransactions = (scope: SearchScope, address: string) => {
+  const { network, layer } = scope
+  const pagination = useSearchParamsPagination('page')
+  const offset = (pagination.selectedPage - 1) * NUMBER_OF_ITEMS_ON_SEPARATE_PAGE
+  if (layer === Layer.consensus) {
+    throw AppErrors.UnsupportedLayer
+    // Loading transactions on the consensus layer is not supported yet.
+    // We should use useGetConsensusTransactions()
+  }
+  const query = useGetRuntimeTransactions(
+    network,
+    layer, // This is OK since consensus has been handled separately
+    {
+      limit: NUMBER_OF_ITEMS_ON_SEPARATE_PAGE,
+      offset: offset,
+      rel: address,
+    },
+  )
+  const { isFetched, isLoading, data } = query
+
+  const transactions = data?.data.transactions || []
+  const totalCount = data?.data.total_count
+  const isTotalCountClipped = data?.data.is_total_count_clipped
+
+  return {
+    isLoading,
+    isFetched,
+    transactions,
+    pagination,
+    totalCount,
+    isTotalCountClipped,
+  }
+}
--- a/src/app/pages/AccountDetailsPage/index.tsx
+++ b/src/app/pages/AccountDetailsPage/index.tsx
@@ -6,10 +6,11 @@ import { SubPageCard } from '../../components/SubPageCard'
 import { Account } from '../../components/Account'
 import { RouterTabs } from '../../components/RouterTabs'
 import { useGetRosePrice } from '../../../coin-gecko/api'
-import { RuntimeAccount } from '../../../oasis-indexer/api'
+import { EvmTokenType, RuntimeAccount } from '../../../oasis-indexer/api'
 import { accountTokenContainerId } from './TokensCard'
 import { useAccount } from './hook'
 import { useRequiredScopeParam } from '../../hooks/useScopeParam'
+import { showEmptyAccountDetails } from '../../../config'
 
 export const AccountDetailsPage: FC = () => {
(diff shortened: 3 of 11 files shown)


csillag:
@donouwens approves the UI change.

src/app/components/RouterTabs/index.tsx
```diff
@@ -21,9 +22,11 @@ export const RouterTabs: FC<RouterTabsProps> = ({ tabs }) => {
   return (
     <>
       <Tabs value={currentTab?.to}>
-        {tabs.map(tab => (
-          <Tab key={tab.to} component={RouterLink} value={tab.to} label={tab.label} to={tab.to} />
-        ))}
+        {tabs
+          .filter(tab => tab === currentTab || tab.visible !== false)
```
lukaw3d, line 26: I don't like this direction. Why hide the tabs? @donouwens

src/app/pages/AccountDetailsPage/hook.ts
```diff
@@ -0,0 +1,73 @@
...
+import { useSearchParamsPagination } from '../../components/Table/useSearchParamsPagination'
+import { NUMBER_OF_ITEMS_ON_SEPARATE_PAGE } from '../../config'
+
+export const useAccount = () => {
+  const { network, layer } = useRequiredScopeParam()
+  if (layer === Layer.consensus) {
+    // There can be no ERC-20 or ERC-721 tokens on consensus
+    throw AppErrors.UnsupportedLayer
+  }
+  const address = useLoaderData() as string
```
lukaw3d, line 20: I think components referenced from routes (page components) should be the ones that obtain parameters from the URL, not API hooks

src/app/pages/AccountDetailsPage/hook.ts
```diff
@@ -0,0 +1,73 @@
...
+    ERC1155: [],
+    ERC20: [],
+    ERC721: [],
+    OasisSdk: [],
+  }
+  if (!isLoading && account) {
+    account.evm_balances.forEach(balance => {
+      if (balance.token_type) tokenBalances[balance.token_type].push(balance)
+    })
+  }
```
lukaw3d, line 34: Can we avoid another hooks layer of API processing?
I would move this into api.ts, or make it a pure function 
`parseAccount(query) => { isLoading, account: { address, balances, tokenBalances, ... } }`

src/app/components/RouterTabs/index.tsx
csillag: My idea was, why offer an option which is not really there, or at least irrelevant. The less visual clutter, the better.

src/app/pages/AccountDetailsPage/hook.ts
csillag: Fine. Will refactor.

src/app/pages/AccountDetailsPage/hook.ts
csillag: Good idea, moving it to  api.ts

src/app/components/RouterTabs/index.tsx
lukaw3d: So far in most places we've been disabling options if unavailable (+explain in tooltip). I guess that let's user see what features we have, and if they do something that unlocks an option, they'll know where to find it.

src/app/pages/AccountDetailsPage/hook.ts
csillag: Done.

src/app/pages/AccountDetailsPage/hook.ts
csillag: Done.

src/app/components/RouterTabs/index.tsx
csillag: I'm leaning towards hiding it, but I'm find both ways. We can discuss later. 

In any case, I have now added a config flag (in `AccountDetaisPage` code) and disabled this feature.
So, no change of behavior occurs.

_However_ I think we could still merge this, since it improves the code in various small ways.

lukaw3d (approved):

src/oasis-indexer/api.ts
```diff
@@ -282,19 +300,23 @@ export const useGetRuntimeAccountsAddress: typeof _f7 = (network, runtime, addre
...
+                balance: token.balance // TODO: why do we test this? According to the API it must be there
+                  ? fromBaseUnits(token.balance, token.token_decimals)
+                  : '0',
               }
             }),
             balances: data.balances?.map(token => {
               return {
                 ...token,
-                balance: token.balance ? fromBaseUnits(token.balance, token.token_decimals) : undefined,
+                balance: token.balance // TODO: why do we test this? According tothe API it must be there
```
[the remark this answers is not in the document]
csillag: OK, but the previously used `undefined` was not a valid value for this type according to [the definition](https://github.com/oasisprotocol/explorer/blob/031268b6e311773d9666c29f552c283f07f1daf8/src/oasis-indexer/generated/api.ts#L951). So we need a better fallback.

As far as I understand, we can extend the auto-generated types using the "declare module" trick, but we can't redefine an existing field as potentially undefined, so we can't change this definition to allow it.

We could introduce a new field for our re-calculated balances, but since the old value would still be part of the record, then there would be no way to make sure that all our code uses the new value.

One way would be to simply drop the the whole record where the balance is missing. Would that be OK with you?

src/app/pages/AccountDetailsPage/index.tsx
```diff
@@ -6,40 +6,45 @@ import { SubPageCard } from '../../components/SubPageCard'
...
 import { RouterTabs } from '../../components/RouterTabs'
 import { useGetRosePrice } from '../../../coin-gecko/api'
-import { Layer, RuntimeAccount, useGetRuntimeAccountsAddress } from '../../../oasis-indexer/api'
-import { AppErrors } from '../../../types/errors'
+import { EvmTokenType, RuntimeAccount } from '../../../oasis-indexer/api'
 import { accountTokenContainerId } from './TokensCard'
+import { useAccount } from './hook'
 import { useRequiredScopeParam } from '../../hooks/useScopeParam'
 
+const SHOW_EVERYTHING = true
```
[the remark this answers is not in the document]
csillag: Renamed to `showEmptyAccountDetails`, and moved to `config.ts`.

src/oasis-indexer/api.ts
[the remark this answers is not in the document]
csillag: OK, so for now, we will simply drop the balances with missing actual balances.  (That's better than showing 0, I guess... and `undefined` was not an option.)

src/app/components/RouterTabs/index.tsx
csillag: Sure thing, that's how it behaves now.

src/app/pages/AccountDetailsPage/hook.ts
```diff
@@ -0,0 +1,51 @@
...
+    layer, // This is OK since consensus has been handled separately
+    {
+      limit: NUMBER_OF_ITEMS_ON_SEPARATE_PAGE,
+      offset: offset,
+      rel: address,
+    },
+  )
+  const { isFetched, isLoading, data } = query
+
+  const transactions = data?.data.transactions || []
```
lukaw3d, line 39: This broke loading state, and I think it's usually good to avoid these fallbacks

Was there a reason for this one?

csillag: Why is loading state not detected using the dedicated flag? In any case I will check it

csillag: To answer my question: I gather that Table wants to keep the existing data even while loading, and only show the skeleton rows when we are loading, _and_ we have no previous data. This is why setting `isLoading` to true is not sufficient to trigger the skeleton rows.

csillag: Fix is available at https://github.com/oasisprotocol/explorer/pull/475

src/oasis-indexer/api.ts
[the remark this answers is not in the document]
csillag: Unfortunately, `undefined` is not a valid fallback value according to the type definition...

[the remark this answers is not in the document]
lukaw3d: Ah i see what you mean. The type comes from typeof generated.useGetRuntimeAccountsAddress, not inferred from the transformed value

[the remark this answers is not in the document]
csillag: >  The type comes from typeof generated.useGetRuntimeAccountsAddress, not inferred from the transformed value

Exactly. And I see no way of overriding that type with another type that allows an existing non-nullable field to be nullable.

PR #462, opened 2023-05-27
Add offline banners everywhere (both for networks and paratimes)

csillag:
The current version only checks the availability of the indexer instances; doesn't yet check the status of the runtimes.

For the sake of testing: 
 - in [this test deployment](https://fcf34b98.oasis-explorer.pages.dev), Mainnet is broken, but testnet is available.
 - in [this test deployment](https://d722d778.oasis-explorer.pages.dev/), Testnet Sapphire is offline, everything else is OK.

## Network unavailable at home page:

image

## Network unavailable at Dashboard (and all other internal pages):

image

## Runtime out of data at Homepage

image

## Runtime out of date at Dashboard

image


src/locales/en/translation.json
```diff
@@ -226,7 +227,8 @@
     }
   },
   "home": {
-    "apiOffline": "Our API is offline. We’re trying to reconnect",
+    "apiOffline": "Our {{ target }} API is offline. We’re trying to reconnect",
+    "runtimeOffline": "Our {{ target }} Paratime is offline. Please try again later.",
```
lukaw3d, line 231: I don't think we'll have a way to detect if a Paratime is offline. It'll either be analyzer or indexer being offline

csillag: Maybe, but in theory, the runtime status API endpoint could have some useful info. I have alreadly opened a slack thread about this.

lukaw3d: Yes.. I mean we will detect either paratime analyzer or paratime indexer is offline, not paratime itself

csillag: Right, what we can easily do is checking how much are we behind the current time, and if it's above a configured threshold, we can display a warning saying that a data is not up to date, processing has only done up to this timestamp...

csillag: OK, we no longer say that the paratime is offline, but that that we don't have fully up-to-date data.

image

src/app/components/OfflineBanner/hook.ts
```diff
@@ -0,0 +1,48 @@
...
+export const useRuntimeFreshness = (scope: SearchScope): FreshnessInfo => {
+  const isApiOffline = useIsApiOffline(scope.network)
+  if (scope.layer === Layer.consensus) {
+    throw new AppError(AppErrors.UnsupportedLayer)
+  }
+  const query = useGetRuntimeStatus(scope.network, scope.layer)
+  const timeDistance = useFormattedTimestampString(query?.data?.data.latest_update)
+  if (isApiOffline || query.isLoading) {
+    return {
+      outOfDate: false,
```
lukaw3d, line 27: False if isApiOffline?

csillag: Yes; if the API is offline, then we basically have no data. 

In this case I don't think it makes sense to add another banner, saying that the data is obsolete....

csillag: Added comment as requested.

src/app/utils/renderWithProviders.tsx
```diff
@@ -2,8 +2,22 @@ import { MemoryRouter } from 'react-router-dom'
...
+const upToDate: FreshnessInfo = {
+  outOfDate: false,
+}
 
 export function renderWithProviders(component: React.ReactElement) {
+  // We need to disable API offline checks
+  const hook1: any = useIsApiOffline
+  hook1.mockReturnValue(false)
+  const hook2: any = useRuntimeFreshness
+  hook2.mockReturnValue(upToDate)
```
lukaw3d, line 18:
```suggestion
  jest.mocked(useIsApiOffline).mockReturnValue(false)
  jest.mocked(useRuntimeFreshness).mockReturnValue({ outOfDate: false })
```

src/config.ts
```diff
@@ -63,6 +71,7 @@ const cipherConfig: LayerConfig = {
   },
   decimals: 9,
   type: RuntimeTypes.Oasis,
+  outOfDateThreshold: 2 * 60 * 1000, // 5 minutes
```
lukaw3d, line 74: Lies in a comment

csillag: Fixed

lukaw3d (approved):

src/app/components/PageLayout/Header.tsx
```diff
@@ -56,7 +61,7 @@ export const Header: FC = () => {
       </Grid>
       {!isMobile && (
         <Box sx={{ mb: 6, px: isMobile ? 0 : '4%' }}>
-          <Search scope={scope} variant="icon" />
+          <Search scope={scope} variant="icon" disabled={isApiOffline || runtimeFreshness.outOfDate} />
```
lukaw3d, line 64: We could/should allow searching when API is out of date

PR #463, opened 2023-05-28
Testnet: teach balances to display TEST not ROSE

csillag:


src/locales/en/translation.json
```diff
@@ -64,6 +64,7 @@
     "smartContract": "Smart Contract",
     "success": "Success",
     "status": "Status",
+    "test": "TEST",
```
lukaw3d, line 67: `common.test` path is so bad.

src/app/pages/TransactionDetailPage/index.tsx
```diff
@@ -136,6 +137,8 @@ export const TransactionDetailView: FC<{
   const isMobile = useMediaQuery(theme.breakpoints.down('sm'))
   const formattedTimestamp = useFormattedTimestampString(transaction?.timestamp)
 
+  const ticker = transaction ? getNetworkTickers(t)[transaction.network] : ''
```
lukaw3d, line 140: Will backend ever add token symbol/ticker too all APIs? If yes, then I would add it to api.ts

csillag:
I realized that some more stuff is needed (around converting to fiat value), so I put this back to draft.

src/locales/en/translation.json
csillag: My biggest question is, do we even want to use i18n for the tickers? It would be so easy if we could just use the name natively...

src/app/pages/TransactionDetailPage/index.tsx
csillag: Good question, I have no idea. In any case, yes, we could add the ticker in api.ts

src/locales/en/translation.json
lukaw3d: I'm not sure if it needs to be translated :shrug:
For good i18n we would at least need `"{value} TEST"`, so languages with different order can swap `"TEST{value}"`.

csillag: > I'm not sure if it needs to be translated 🤷🏻 

OK, asked on the slack channel, hopefully someone with a non-English background will chip in.

csillag: Preliminary feedback seems to indicate that we don't need to translate the ticker symbols. In any case I have removed the dummy ROSE and TEST values from the translations.json file, and created a stub function for future extension if we want to translate them somehow.

src/app/pages/TransactionDetailPage/index.tsx
csillag: OK, for now I'm adding the ticker to all accounts and transactions in the API layer. Opened discussion with backend developers about including the data; if they will add it, we can remove it..

src/app/pages/AccountDetailsPage/index.tsx
```diff
@@ -19,8 +18,6 @@ export const AccountDetailsPage: FC = () => {
   const address = useLoaderData() as string
   const { account, isLoading } = useAccount(scope, address)
 
-  const rosePriceQuery = useGetRosePrice()
```
csillag, line 22: @lukaw3d says:

> hm, keeping requests in Page components is becoming a bit annoying. I would still stick to it tho. Makes tests, storybooks, and reusing components clearer

csillag: Bit isn't the ticker price like a textbook example of data that should be passed down in context, instead of trickling it through like 3 layers of components? Especially since the price of ROSE is totally independent of the page we are at..

The tests, storybooks, etc is indeed clearer, but then again, those can also be handled by adding the right component wrapper / mock hook...

csillag: Anyway, I have put it back to page level now.

PR #464, opened 2023-05-28
Don't show ROSE price when not on mainnet

csillag:


csillag:
Please note that we are still waiting for the final OK from @donouwens on this one.

csillag:
I dont think it is relevant. Rose tokens only live on rhe Mainnet, and it is impossible to do any transactions betweek testnet and mainnet.

csillag:
> @csillag Sure, but isn't the Testnet used for anything that's aimed to be on the Mainnet eventually? Correct me if I'm wrong, but with that in mind, the ROSE token holds some relevancy, 

Yes, that is true. (This is the reason we even have to ask this question.)

> Test token don't have any value we can show there, no?

No, TEST tokens are free. (You get them for free by opening a [faucet](https://faucet.testnet.oasis.dev/).)

csillag:
> do the other widgets hold any value in that case (Active Accounts, Nodes, Transaction)? Do those widgets show the total of all networks or only the data from the Testnet?

All of these other widgets show data excusively from TestNet.

> If they only show the data from Testnet

Yes

> I agree to keep those and remove (or replace see below**) the ROSE token widget and keep the other widgets.

OK, here we go.

> However, if we show data from Mainnet or all networks in those widgets ...

They don't. 


> Would it have any value to show how many TEST Tokens are used over period x instead of the ROSE widget?

I don't know. Do we even have such data?

csillag:
In any case, we are in agreement that we _can_ remove this one, right? (And then we can think about maybe adding something else to the free space later?)

lukaw3d (approved):

PR #466, opened 2023-05-30
Sticky header

buberdds:


csillag:
I tested this and it works great, but I'm a bit concerned by the fact that the composition of the components is different on desktop and mobile. (Ie. the search is included in PageLayou on desktop, but in Header on mobile.)

While I see why this might have been the most natural way to implement this now from a technical point of view, I am concerned that maintaining this in the future might be difficult and error prone, because there are too many variations. 

Wouldn't it be possible to avoid this fundamental difference between desktop and mobile, somehow?

src/app/components/PageLayout/Logotype.tsx
```diff
@@ -9,7 +9,7 @@ import { useTranslation } from 'react-i18next'
 
 interface LogotypeProps {
   color?: string
-  showText: boolean
+  showText?: true | false | undefined // If undefined: shows text on desktop
```
csillag, line 12: Do we need both the question mark _and_ the undefined value? Seems a bit redundant...

lukaw3d: Nope

lukaw3d:
Rebased and split commits

lukaw3d (approved):

csillag:
> [lukaw3d](https://github.com/lukaw3d) approved these changes [2 minutes ago](https://github.com/oasisprotocol/explorer/pull/466#pullrequestreview-1470950128)

I think since @buberdds is not around ATM, we should just merge this, now that it's approved.

PR #467, opened 2023-05-30
Display tx value also in USD

csillag:
(This is based on #463)

For accounts, we are displaying the balance in ROSE _and_ USD, but for transactions, we only displayed the value in ROSE.

As discussed with @donouwens , we could also display fiat value here. This change adds the USD value for the transactions, too. Like this:

image

Please note that we are basing this calculation on ROSE's _current_ value... so the value at the time of transfer might have been completely different. We might want to make this clear in the description somehow...


csillag:
> Please note that we are basing this calculation on ROSE's _current_ value... so the value at the time of transfer might have been completely different. We might want to make this clear in the description somehow...

Updated UI to reflect this:

image

And while hovering on the question mark:

image

src/app/components/Account/index.tsx
```diff
@@ -88,22 +93,22 @@ export const Account: FC<AccountProps> = ({ account, isLoading, roseFiatValue, s
...
+                    value: parseFloat(balance) * tokenFiatValue,
                     formatParams: {
                       value: {
                         currency: 'USD',
                       } satisfies Intl.NumberFormatOptions,
                     },
                   })}
-                  <CoinGeckoReferral />
-                </StyledBox>
+                  {usedGecko && <CoinGeckoReferral />}
```
lukaw3d, line 110: I would rename "hasUsedCoinGecko"

src/coin-gecko/api.ts
```diff
@@ -40,3 +41,26 @@ export function useGetRosePrice() {
     },
   )
 }
+
+export type TokenPriceInfo = {
+  price?: number
+  usedGecko?: boolean
```
lukaw3d, line 47: I would make it a required field :shrug:

src/coin-gecko/api.ts
```diff
@@ -40,3 +41,26 @@ export function useGetRosePrice() {
...
+}
+
+export const useTokenPrice = (ticker: NativeTicker): TokenPriceInfo => {
+  const rosePrice = useGetRosePrice().data
```
lukaw3d, line 51: (small note: this makes a request even if picker isnt ROSE)

src/coin-gecko/api.ts
```diff
@@ -40,3 +41,26 @@ export function useGetRosePrice() {
     },
   )
 }
+
+export type TokenPriceInfo = {
+  price?: number
```
lukaw3d, line 46: I'd rather see an explicit loading state

src/app/components/Account/index.tsx
```diff
@@ -88,22 +93,22 @@ export const Account: FC<AccountProps> = ({ account, isLoading, roseFiatValue, s
           </dd>
 
           <dt>{t('common.balance')}</dt>
-          <dd>{t('common.valueInRose', { value: balance })}</dd>
-          {roseFiatValue && balance && (
+          <dd>{t('common.valueInToken', { value: balance, ticker: tickerName })}</dd>
+          {!!tokenFiatValue && balance && (
```
lukaw3d, line 97: !! also returns false when value is 0. We dont care to show 0 USD everywhere on testnet, but I think the abstraction should be clearer about that, instead of treating 0 and undefined the same.

src/app/pages/AccountDetailsPage/index.tsx
```diff
@@ -22,16 +21,11 @@ export const AccountDetailsPage: FC = () => {
   const address = useLoaderData() as string
   const accountQuery = useGetRuntimeAccountsAddress(network, layer, address)
   const account = accountQuery.data?.data
-  const rosePriceQuery = useGetRosePrice()
```
lukaw3d, line 25: hm, keeping requests in Page components is becoming a bit annoying. I would still stick to it tho. Makes tests, storybooks, and reusing components clearer

lukaw3d (approved):

src/app/components/Account/index.tsx
csillag: Done

src/coin-gecko/api.ts
csillag: Done

src/app/pages/AccountDetailsPage/index.tsx
csillag: This comment actually belongs to https://github.com/oasisprotocol/explorer/pull/463, where this change was made. (This PR is build on top of that PR.)

src/coin-gecko/api.ts
csillag: Added

src/app/components/Account/index.tsx
csillag: OK, so I have added an explicit `isFree` flag to the hook, and we only hide the value when this flag is on (or still loading, etc.) But in any case, a price of 0 is no longer sufficient reason to hide it. (And of course I have enabled the free flag for TEST tokens.)

src/coin-gecko/api.ts
csillag: > (small note: this makes a request even if picker isnt ROSE)

Yes, that's because of the stupid rules of hooks. Is there a good way to make the call conditional? Please note that (conditionally) setting `enabled: false` on the query is _not_ sufficient to skip it.

csillag:
> [lukaw3d](https://github.com/lukaw3d) approved these changes [8 hours ago](https://github.com/oasisprotocol/explorer/pull/467#pullrequestreview-1463001346)

So far, so good, but this actually depends on #463. We still need to review that, before merging this one, right?

csillag:
> So far, so good, but this actually depends on #463. We still need to review that, before merging this one, right?

Never mind, we have reviewed that, too, as part of this review.

src/app/pages/TransactionDetailPage/CurrentFiatValue.tsx
```diff
@@ -0,0 +1,34 @@
...
+  return price === undefined ? null : (
+    <FiatMoneyAmountBox>
+      <span>
+        {t('common.fiatValueInUSD', {
+          value: amount * price,
+          formatParams: {
+            value: {
+              currency: 'USD',
+            } satisfies Intl.NumberFormatOptions,
+          },
```
lukaw3d, line 24: Just noticed that this sometimes rounds up!
https://explorer.dev.oasis.io/mainnet/sapphire/transactions/0x0de151f3bc5bb385526933b075bd7063733809a815e4ce3fbd428b485f955e15

0.100000001686272452 ROSE * 0.05003 => $0.01

lukaw3d: And `roundingMode: 'trunc',` doesn't work in firefox

csillag: How about saying "**Approximate** fiat value", instead of "Fiat value"?

PR #472, opened 2023-06-06
Switch to production API

lukaw3d:


src/oasis-indexer/generated/api.ts
```diff
@@ -741,6 +811,10 @@ Absent if the event did not originate from a transaction.
 Absent if the event did not originate from a transaction.
  */
   tx_hash?: string | null;
+  /** Ethereum trasnsaction hash of this event's originating transaction.
+Absent if the event did not originate from an EVM transaction.
+ */
+  eth_tx_hash?: string;
```
csillag, line 817: I think this is something we have been waiting for, and we are ready to use if enabled.

csillag (approved):

PR #476, opened 2023-06-07
Adjust the behavior of ScopedSearch: completely ignore other paratimes

csillag:
- ScopedSearch should show results only from the wanted paratime
- Fix some Styling issues

### DIFF
--- a/src/app/pages/SearchResultsPage/useRedirectIfSingleResult.ts
+++ b/src/app/pages/SearchResultsPage/useRedirectIfSingleResult.ts
@@ -6,14 +6,10 @@ import { isItemInScope, SearchScope } from '../../../types/searchScope'
 import { Network } from '../../../types/network'
 
 /** If search only finds one result then redirect to it */
-export function useRedirectIfSingleResult(
-  scope: SearchScope | undefined,
-  isLoading: boolean,
-  results: SearchResults,
-) {
+export function useRedirectIfSingleResult(scope: SearchScope | undefined, results: SearchResults) {
   const navigate = useNavigate()
 
-  let shouldRedirect = !isLoading && results.length === 1
+  let shouldRedirect = results.length === 1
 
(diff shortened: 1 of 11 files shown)


src/app/pages/SearchResultsPage/useRedirectIfSingleResult.ts
```diff
@@ -6,14 +6,10 @@ import { isItemInScope, SearchScope } from '../../../types/searchScope'
...
-export function useRedirectIfSingleResult(
-  scope: SearchScope | undefined,
-  isLoading: boolean,
-  results: SearchResults,
-) {
+export function useRedirectIfSingleResult(scope: SearchScope | undefined, results: SearchResults) {
   const navigate = useNavigate()
 
-  let shouldRedirect = !isLoading && results.length === 1
+  let shouldRedirect = results.length === 1
```
lukaw3d, line 12: Maybe we shouldn't be combining results until all searches are done, so `results.length === 1` is safer
https://github.com/oasisprotocol/explorer/blob/6ffa4e77f7cb71c13d3054d83512d87608859eda/src/app/pages/SearchResultsPage/hooks.ts#L115-L119

useRedirectIfSingleResult currently relies on being included in components after isLoading:
https://github.com/oasisprotocol/explorer/blob/6ffa4e77f7cb71c13d3054d83512d87608859eda/src/app/pages/SearchResultsPage/SearchResultsView.tsx#L21-L33

csillag: > Maybe we shouldn't be combining results until all searches are done,

We don't. In the current code, the redirect thingy is only called after the whole loading is done.

csillag: Updated hook so that we always return [] while loading.

PR #477, opened 2023-06-08
Add strict type-checks that most switch statements cover all cases

lukaw3d:


src/types/errors.ts
```diff
@@ -25,3 +25,19 @@ export interface ErrorPayload {
   code: AppErrors
   message: string
 }
+
+// Adds strict type-check that a type was exhausted
+// https://www.typescriptlang.org/docs/handbook/2/narrowing.html#exhaustiveness-checking
+// https://stackoverflow.com/questions/41102060/typescript-extending-error-class
+export class ExhaustedTypeError extends Error {
```
lukaw3d, line 32: This is the same as https://github.com/oasisprotocol/oasis-wallet-web/blob/71c3967b865812d84faad783395c09c9aebfe3f6/src/types/errors.ts#L38

src/app/components/Transactions/LogEvent.tsx
```diff
@@ -111,13 +112,14 @@ const DecodedLogEvent: FC<{ scope: SearchScope; event: RuntimeEvent }> = ({ scop
...
     case RuntimeEventType.consensus_accountswithdraw:
-    default:
       return (
         <div>
           <div>{eventName}</div>
           <pre>{JSON.stringify(event, null, ' ')}</pre>
         </div>
       )
+    default:
+      throw new ExhaustedTypeError('Unexpected event type', event.type)
```
csillag, line 122: I'm not happy with throwing an error (and hence breaking the whole  page) just because we see something we don't recognize. I thing we should also have an `exhaustedTypeWarning()` function, which just uses `console.warn()`, and use that in the milder cases like this one.

lukaw3d: Converted to warnings in production and errors in development

src/app/pages/DashboardPage/LearningMaterials.tsx
```diff
@@ -49,7 +49,12 @@ const getContent = (t: TFunction, layer: Layer) => {
         link: docs.sapphire,
         title: t('common.sapphire'),
       }
+    case Layer.cipher:
+      throw new AppError(AppErrors.UnsupportedLayer)
+    case Layer.consensus:
+      throw new AppError(AppErrors.UnsupportedLayer)
     default:
+      exhaustedTypeWarning('Unexpected layer', layer)
       throw new AppError(AppErrors.UnsupportedLayer)
```
csillag, line 58: Here we are throwing two exceptions on quick succession. (At least in PROD mode.) Are they both necessary?

csillag (approved):
Reviewed, LGTM (except one comment)

src/app/pages/DashboardPage/LearningMaterials.tsx
lukaw3d: Yes, unless we introduce more fallbacks

PR #480, opened 2023-06-08
Refactor banners and show them based on domain so same build can be deployed to production

lukaw3d:
Mirrors how we handle banners in web wallet:
https://github.com/oasisprotocol/oasis-wallet-web/pull/1447
https://github.com/oasisprotocol/oasis-wallet-web/pull/1491/commits/7729c966e2105d9f6c2d17a0927983b0cc3f89f6
https://github.com/oasisprotocol/oasis-wallet-web/pull/1457


csillag:
Do we really need a banner when developing locally?

csillag (approved):
Generally I'm OK with this, except that I would treat localhost as an exception, and don't show any banners there.

lukaw3d:
> Generally I'm OK with this, except that I would treat localhost as an exception, and don't show any banners there.

That seems safe :+1:

PR #482, opened 2023-06-08
Mobile ParaTime picker

lubej:


csillag:
A few questions / suggestions, @donouwens please comment.

1. Currently, when opening the paratime picker on mobile, we start with the network selection page. I don't think we want that. The idea was that we don't really want to promote this whole idea of having different network, as of yet. That is why on the desktop, when opening the paratime picker, the network selection menu is hidden; it has to be opened using a separate button. Therefore, I think that when opening it on mobile, we should start with the layer selector, and it should require a dedicated action to get to the nework selection. Right?

2. Shall we us `Paratime` or `ParaTime` ? I think we should standardise on one, but now I see both versions.

3. After selecting a paratime from the list, we arrive to the confirmation screeen. The buttons at the bottom say cancel / select, and the way to get back to the list is on the top, saying "Select ParaTime". My problem is that we are using "select" twice, with different meaning here. On the top, "select Paratimes" means "go back to the list", and on the bottom, "select" means "confirm my selection and close the dialog". I think we should change the wording to be less confusing. Maybe say "Back to the list of Paratimes" on the top?
4.

csillag:
Tested this again. Two comments/questions, to @donouwens and then to @lubej :

1. On the desktop version, the button opening the network menu is nameless; there is no text, in order to de-emphasize it. However, in the mobile version, it ways "< View Networks". I think if we want to be consistent, we should also hide the text here, and just go with the "<", right?
2. For the disabled layers (like Consensus, Cipher, and even Emerald but it's not yet visible on this branch), there is no indicator whatsoever. On the desktop we have at least the tooltip, but on mobile there is no hover so no tooltip. There should be something, right? I think a design has a label saying "coming soon" or something like that....

src/app/components/PageLayout/Header.tsx
```diff
@@ -8,18 +8,35 @@ import { NetworkSelector } from './NetworkSelector'
...
+    ref: headerRef,
+  })
+
+  useEffect(() => {
+    if (!isTablet) {
+      return
+    }
+
+    if (headerRef.current !== null) {
+      document.body.style.setProperty('--app-header-height', `${headerHeight?.toFixed(2) || 0}px`)
```
csillag, line 33: Is this a generally valid, compatible and accepted CSS selector?

src/app/components/ParaTimePicker/LayerDetails.tsx
```diff
@@ -80,6 +81,7 @@ const contentMinHeight = '270px'
 export const LayerDetails: FC<LayerDetailsProps> = ({ network, selectedLayer }) => {
   const { t } = useTranslation()
   const theme = useTheme()
+  const isTablet = useMediaQuery(theme.breakpoints.down('md'))
```
csillag, line 84: Can't we use the newly available `useScreenSize()` for this? It already provides `isTable`. I  think it would be confusing to use different definitions at multiple points of the code... right?

src/app/components/ParaTimePicker/LayerMenu.tsx
```diff
@@ -12,13 +12,24 @@ import { getLayerLabels } from '../../utils/content'
...
+  <Typography
+    component="span"
+    sx={{ fontSize: '10px', fontStyle: 'italic', color: COLORS.grayMedium, ml: 2 }}
+  >
+    {children}
+  </Typography>
+)
+
 export const DisabledLayerMenuItem: FC<BaseLayerMenuItemProps> = ({ divider, layer }) => {
+  const { isTablet } = useScreenSize()
```
csillag, line 32: .. and here we are using `isTable` from `useScreenSize()`. I propose that we do it always the same way, like this.

src/styles/theme/defaultTheme.ts
```diff
@@ -36,6 +37,7 @@ declare module '@mui/material/styles' {
   }
 
   interface PaletteColor extends CustomLayoutPalette {}
+
```
csillag, line 40: Pls drop new empty line

src/styles/theme/defaultTheme.ts
```diff
@@ -70,6 +72,7 @@ declare module '@mui/material/Chip' {
   export interface ChipPropsColorOverrides {
     tertiary: true
   }
+
```
csillag, line 75: Pls drop new empty line

src/app/components/ParaTimePicker/index.tsx
```diff
@@ -18,80 +18,187 @@ import { NetworkMenu } from './NetworkMenu'
...
 
+enum ParaTimePickerTabletStep {
+  Network,
+  ParaTime,
+  ParaTimeDetails,
+}
+
 const ParaTimePickerContent: FC<ParaTimePickerContentProps> = ({ onClose, onConfirm }) => {
+  const theme = useTheme()
+  const isTablet = useMediaQuery(theme.breakpoints.down('md'))
```
csillag, line 91: Again, please let's stick to `useScreenSizes()`

csillag:
Reviewed the code, generally looks good (albeit there is lots of ugly CSS magic). I have left a bunch of small comments, and I have a bigger question which we will discuss in person

src/app/components/ParaTimePicker/LayerDetails.tsx
lubej: Right, leftover from the rebase. Nice catch 👍

buberdds:
on https://oasis-explorer.pages.dev/ it looks a bit weird 
Screenshot 2023-06-19 at 16 50 51

lubej:
> 



> on https://oasis-explorer.pages.dev/ it looks a bit weird Screenshot 2023-06-19 at 16 50 51

Ah right, the warning message hides when scrolled to the bottom. Will fix

PR #478, opened 2023-06-08
Adjust global search

csillag:
Depends on #476 

- Only show MainNet results by default; the rest only upon request.
- Group results by network and not scopes.
- Inside network sections, sort results by paratime.

PR #484, opened 2023-06-08
Clean up / simplify code

csillag:
No need to filter and regroup the results,
we can just use the whole input array,
since it has been filtered at the parent layer.


csillag:
Reaction to https://github.com/oasisprotocol/explorer/pull/478#discussion_r1223326004

PR #486, opened 2023-06-09
Revert "Move banners into sticky header"

lukaw3d:
This reverts commit 56fc8ccb72dc04ae3db3e9332bff8e8936ac8eee.

> @donouwens:  Actually, I would prefer the warning banners to not remain within the viewport, to reduce the screen estate the take up


lukaw3d:
| Before | After |
| --- | --- |
| before  |  after |

csillag (approved):
Reviewed. LGTM

PR #491, opened 2023-06-12
Don't show a second 'Transactions' in the transactions row

csillag:
As per https://app.clickup.com/t/865cgzwdk

### DIFF
--- a/src/app/components/Account/index.tsx
+++ b/src/app/components/Account/index.tsx
@@ -61,7 +61,7 @@ export const Account: FC<AccountProps> = ({ account, isLoading, tokenPriceInfo, 
   const balance = account?.balances[0]?.balance ?? '0'
   const address = account ? account.address_eth ?? account.address : undefined
 
-  const transactionsLabel = account ? t('common.transactionsNumber', { count: account.stats.num_txns }) : ''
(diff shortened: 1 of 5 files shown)


src/app/components/Account/index.tsx
```diff
@@ -61,7 +61,7 @@ export const Account: FC<AccountProps> = ({ account, isLoading, tokenPriceInfo,
   const balance = account?.balances[0]?.balance ?? '0'
   const address = account ? account.address_eth ?? account.address : undefined
 
-  const transactionsLabel = account ? t('common.transactionsNumber', { count: account.stats.num_txns }) : ''
+  const transactionsLabel = account ? account.stats.num_txns : ''
```
buberdds, line 64: We need to convert both `account.stats.num_txns` and `block.num_transactions` numbers to string using  `toLocaleString()`

csillag: The reason I didn't to that is that we are already changing it at the bignumber PR #333, and I wanted to avoid the conflict. But whatever, I will just do it as you suggested, for now

csillag: Done

buberdds (approved):

PR #492, opened 2023-06-12
TX details: rename "block" to "height"

csillag:
As per https://app.clickup.com/t/861mx1rme


csillag:
No, the screenshot was wrong, we shouldn't do this.

PR #494, opened 2023-06-12
Centralize and simplify checking the screen size

csillag:
This PR doesn't change any functionality, just simplifies how we calculate the `isMobile`, `isTabled` and `isDesktop` flags.

### DIFF
--- a/src/app/hooks/useScreensize.ts
+++ b/src/app/hooks/useScreensize.ts
@@ -1,0 +1,11 @@
+import useMediaQuery from '@mui/material/useMediaQuery'
+import { useTheme } from '@mui/material/styles'
+
+export const useScreenSize = () => {
+  const theme = useTheme()
+  return {
+    isMobile: useMediaQuery(theme.breakpoints.down('sm')),
+    isTablet: useMediaQuery(theme.breakpoints.down('md')),
+    isDesktop: useMediaQuery(theme.breakpoints.up('md')),
+  }
+}
--- a/src/app/components/Account/AccountLink.tsx
+++ b/src/app/components/Account/AccountLink.tsx
@@ -1,7 +1,6 @@
 import { FC } from 'react'
 import { Link as RouterLink } from 'react-router-dom'
-import { useTheme } from '@mui/material/styles'
-import useMediaQuery from '@mui/material/useMediaQuery'
+import { useScreenSize } from '../../hooks/useScreensize'
 import Link from '@mui/material/Link'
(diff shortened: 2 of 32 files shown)


src/app/hooks/useScreensize.ts
```diff
@@ -0,0 +1,11 @@
+import useMediaQuery from '@mui/material/useMediaQuery'
+import { useTheme } from '@mui/material/styles'
+import { Theme } from '@mui/material/styles/createTheme'
+
+export const useScreenSizeOnTheme = (theme: Theme) => ({
+  isMobile: useMediaQuery(theme.breakpoints.down('sm')),
+  isTablet: useMediaQuery(useTheme().breakpoints.down('md')),
```
buberdds, line 7: There is no consistency, `isMobile` relies on function arg and `itTablet`/`isDesktop` use MUI hook directly.

csillag: Ah sorry, fixed.

src/app/hooks/useScreensize.ts
```diff
@@ -0,0 +1,11 @@
+import useMediaQuery from '@mui/material/useMediaQuery'
+import { useTheme } from '@mui/material/styles'
+import { Theme } from '@mui/material/styles/createTheme'
+
+export const useScreenSizeOnTheme = (theme: Theme) => ({
```
buberdds, line 5: Do we need this additional hook? If this was created just to handle `ResultListFrame`. I would just replace `isMobile` in styled component with ` [theme.breakpoints.down('sm')]` syntax (no need for calling hook).

csillag: OK, done as suggested.

buberdds (approved):

src/app/hooks/useScreensize.ts
```diff
@@ -0,0 +1,11 @@
+import useMediaQuery from '@mui/material/useMediaQuery'
+import { useTheme } from '@mui/material/styles'
+
+export const useScreenSize = () => {
+  const theme = useTheme()
+  return {
+    isMobile: useMediaQuery(theme.breakpoints.down('sm')),
+    isTablet: useMediaQuery(theme.breakpoints.down('md')),
```
lukaw3d, line 8: This seems wrong. Should be between mobile and desktop

csillag: After discussing the same issue with @lubej, I have opened https://github.com/oasisprotocol/explorer/issues/555.

PR #500, opened 2023-06-12
Paratime Picker design feedback

buberdds:


csillag:
Please note that we also have #482 , waiting for review.  Maybe it would make sense to finish that first?

buberdds:
Validated by Don, ready for a review. I don't think it has to wait as PRs target different problems.

src/app/components/LayerStatus/index.tsx
```diff
@@ -0,0 +1,30 @@
+import { FC } from 'react'
+import { useTranslation } from 'react-i18next'
+import CheckCircleIcon from '@mui/icons-material/CheckCircle'
+import ErrorIcon from '@mui/icons-material/Error'
+import Typography from '@mui/material/Typography'
+import { COLORS } from '../../../styles/theme/colors'
+
+type LayerStatusProps = {
+  valid: boolean
```
csillag: I would prefer calling it `uptodate` (or `upToDate`) instead of `valid`, because invalid is too strong to describe a state that is not fully uptodate...

csillag: Ahh, I see it has been renamed in the next commit. Never mind then.

PR #505, opened 2023-06-12
Indicate encrypted transactions in TX lists

csillag:
As per https://app.clickup.com/t/861mx1bew and https://app.clickup.com/t/861mx1kwu

For testing, look for block 411186 on Sapphire testnet, or the account 0x6fA03516728d4a483578D63ea124C4Ec4A00e2bB

### DIFF
--- a/src/app/components/Transactions/index.tsx
+++ b/src/app/components/Transactions/index.tsx
@@ -4,6 +4,7 @@ import { useTranslation } from 'react-i18next'
 import formatDistanceStrict from 'date-fns/formatDistanceStrict'
 import Box from '@mui/material/Box'
 import ArrowForwardIcon from '@mui/icons-material/ArrowForward'
+import LockIcon from '@mui/icons-material/Lock'
 import { Table, TableCellAlign } from '../../components/Table'
 import { TransactionStatusIcon } from '../../components/TransactionStatusIcon'
 import { RuntimeTransactionLabel } from '../../components/RuntimeTransactionLabel'
@@ -16,6 +17,8 @@ import { AccountLink } from '../Account/AccountLink'
 import { TransactionLink } from './TransactionLink'
 import { trimLongString } from '../../utils/trimLongString'
 import Typography from '@mui/material/Typography'
+import { doesAnyOfTheseLayersSupportEncryptedTransactions } from '../../../types/layers'
+import { TransactionEncryptionStatus } from '../TransactionEncryptionStatus'
 
 const StyledCircle = styled(Box)(({ theme }) => ({
   position: 'absolute',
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -63,6 +63,7 @@
     "bignumber.js": "9.1.1",
     "date-fns": "2.30.0",
     "i18next": "22.5.1",
+    "lodash.uniq": "^4.5.0",
```
buberdds, line 66: Shouldn't we use lodash-es? Other than that, do we need 3rd party lib here? Can't we just `[...new Set(layers)]` instead of uniq ?

csillag: I don't know which is the preferred library, and I don't have a strong preference either way. OK, dropping the 3rd party...

src/app/components/Transactions/index.tsx
```diff
@@ -59,8 +62,14 @@ export const Transactions: FC<TransactionsProps> = ({
...
 }) => {
   const { t } = useTranslation()
+  // We only want to show encryption status of we are listing transactions
+  // from paratimes that actually support encrypting transactions
+  const canHaveEncryption = doesAnyOfTheseLayersSupportEncryptedTransactions(
+    transactions?.map(tx => tx.layer),
+  )
   const tableColumns = [
     { content: t('common.status') },
+    ...(canHaveEncryption ? [{ content: (<LockIcon />) as unknown as string }] : []), // The table does support widgets in the column headers, but the TS definition is unaware of that.
```
buberdds, line 72: LockIcon needs `COLORS.grayMedium`

csillag: Added

src/types/layers.ts
```diff
@@ -26,3 +28,11 @@ const layerOrder: Record<Layer, number> = {
 
 export const orderByLayer = (itemA: HasLayer, itemB: HasLayer): number =>
   layerOrder[itemA.layer] - layerOrder[itemB.layer]
+
+const layersWithEncryptedTransactions: Layer[] = [Layer.sapphire]
+
+export const doesLayerSupportEncryptedTransactions = (layer: Layer): boolean =>
+  layersWithEncryptedTransactions.indexOf(layer) !== -1
```
buberdds, line 35: optional
`.includes(layer)` is more readable

buberdds (approved):

src/types/layers.ts
csillag: Done

src/app/utils/helpers.ts
```diff
@@ -57,3 +57,7 @@ export function getEthAccountAddress(preimage: AddressPreimage | undefined): str
   // We need to convert from base64 to hex, add the prefix, and convert to checksum address
   return toChecksumAddress(`0x${Buffer.from(preimage.address_data, 'base64').toString('hex')}`)
 }
+
+export function uniq<T>(input: T[] | undefined): T[] {
+  return input === undefined ? [] : [...new Set(input)]
```
buberdds, line 62: not sure we need this function, but Ok. [...new Set(undefined)] will return an empty array anyway

src/app/components/Transactions/index.tsx
```diff
@@ -59,8 +62,16 @@ export const Transactions: FC<TransactionsProps> = ({
...
+  // We only want to show encryption status of we are listing transactions
+  // from paratimes that actually support encrypting transactions
+  const canHaveEncryption = doesAnyOfTheseLayersSupportEncryptedTransactions(
+    transactions?.map(tx => tx.layer),
+  )
   const tableColumns = [
     { content: t('common.status') },
+    ...(canHaveEncryption
+      ? [{ content: (<LockIcon htmlColor={COLORS.grayMedium} />) as unknown as string }]
+      : []), // The table does support widgets in the column headers, but the TS definition is unaware of that.
```
lukaw3d, line 74: Table does not support it too well
https://github.com/oasisprotocol/explorer/blob/a77aa3ac81aa59ba4dd532f2fb4db90161052ce9/src/app/components/Table/index.tsx#L108

csillag: Yes, that could be improved. (By generating the key using the index, or something similar.)

PR #511, opened 2023-06-13
Add mobile ticks

lubej:


lubej:
Task:
https://app.clickup.com/t/861mx1fuv

src/index.tsx
```diff
@@ -10,6 +10,9 @@ import './styles/index.css'
 import './locales/i18n'
 import Axios from 'axios'
 
+// ES2020 Intl.NumberFormatOptions Polyfill
+import '@formatjs/intl-numberformat/polyfill'
```
buberdds, line 14: why we need polyfill ?

src/app/components/charts/LineChart.tsx
```diff
@@ -57,6 +60,9 @@ const LineChartCmp = <T extends object>({
         tick={withLabels ? { fill: COLORS.brandDark, strokeWidth: 0 } : false}
         type="number"
         tickMargin={tickMargin}
+        tickFormatter={tick => {
+          return tickMark ? intlNumberFormat(tick) : tick
```
buberdds, line 64: can't we handle this via i18next?

csillag: I think we can handle this with bignumber. For now, let's just  use `number.toLocaleString()`, and add a TODO, and I will pulish it in #333.

buberdds:
```
          tickFormatter={tick =>
            t('foo', {
              value: tick,
              formatParams: {
                value: {
                  notation: 'compact',
                } satisfies Intl.NumberFormatOptions,
              },
            })
          }
          ```
          in lang file `"foo": "{{value, number}}",`

lubej: Right, thanks.  TypeScript was complaining for some reason, and I was thinking we need the polyfill. Removed.

src/index.tsx
lubej: Removed

buberdds (approved):

PR #514, opened 2023-06-13
Implement scroll to top on page navigate

lubej:


lubej:
Task:
https://app.clickup.com/t/861mx1hfj

lukaw3d:
My preferred UX would be: scroll to top on navigation, but keep position when refreshing and reopening tabs

lukaw3d:
and restore position when clicking back (or forward)

csillag:
> My preferred UX would be: scroll to top on navigation, but keep position when refreshing and reopening tabs, and restore position when clicking back (or forward)

👍🏻

lubej:
> > My preferred UX would be: scroll to top on navigation, but keep position when refreshing and reopening tabs, and restore position when clicking back (or forward)
> 
> 👍🏻

Implemented, ScrollRestoration did not quite cover all the cases, especially on forward and page refresh. As the case for reopening the tabs, it works as long there is an active tab open - since the scroll map is saved in sessionStorage.

lukaw3d:
@lubej can you check if https://github.com/oasisprotocol/explorer/pull/567 (https://2010762f.oasis-explorer.pages.dev/) satisfies all requirements? I checked the basics, and seems to behave the same/better, without a custom implementation

lubej:
> @lubej can you check if #567 (https://2010762f.oasis-explorer.pages.dev/) satisfies all requirements? I checked the basics, and seems to behave the same/better, without a custom implementation

One thing that does not work is reopening the tab, but given how much it simplifies the code, I would go with your implementation. Not sure why I was facing issues with ScrollRestoration -> forward particularly. When in your case it seems to work fine.

PR #522, opened 2023-06-13
Display a message when failing to load account details

csillag:
Before:
image

After:
image

### DIFF
--- a/.changelog/522.bugfix.md
+++ b/.changelog/522.bugfix.md
@@ -1,0 +1,1 @@
+Show a nice error message when can't load account details.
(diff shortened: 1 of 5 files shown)


csillag:
This was triggered by recent SQL problems on the backend, so this is not just a theoretical consideration, but actually relevant in the real world.

csillag:
> Could we change the copy to: 'Unfortunately we couldn't load the account details at this time. Please try again later.'

Done. (I won't regenerate the screenshot for this change..)

buberdds (approved):

PR #523, opened 2023-06-13
Improve handling of nameless ERC-20 tokens

csillag:
There are some ERC-20 tokens which don't have a name or a ticker, because these are (strictly speaking) not mandatory parts of the contract specification. We still need to display something on the UI.

This can be tested with account 0x0df7C50f085345a12A2d06cE0D5519607b6E9B47 (Sapphire, Testnet.)

Before:

image

After:

image

### DIFF
--- a/.changelog/523.trivial.md
+++ b/.changelog/523.trivial.md
@@ -1,0 +1,1 @@
+Improve handling of nameless ERC-20 tokens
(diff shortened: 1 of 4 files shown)


csillag:
OK, so now it looks like this: 
image

buberdds (approved):

PR #527, opened 2023-06-14
Use custom locale for formatting distance between dates

buberdds:
We need compact/custom labels when showing Age column labels. 

Another PR will change table cell paddings so we should be able to resolve issue from Slack.


csillag (approved):
Reviewed. LGTM.

PR #532, opened 2023-06-14
Tweak md breakpoint

buberdds:
Don's ideas from Slack, ~waiting for deploy URL~

### DIFF
--- a/.changelog/532.bugfix.2.md
+++ b/.changelog/532.bugfix.2.md
@@ -1,0 +1,1 @@
+Adjust layout to new breakpoint
--- a/src/styles/theme/defaultTheme.ts
+++ b/src/styles/theme/defaultTheme.ts
@@ -83,6 +83,12 @@ declare module '@mui/material/styles' {
 }
 
 export const defaultTheme = createTheme({
+  breakpoints: {
+    values: {
+      ...createTheme().breakpoints.values,
+      lg: 1024,
+    },
+  },
   palette: {
     background: {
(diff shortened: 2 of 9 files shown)


.changelog/532.bugfix.2.md
```diff
@@ -0,0 +1 @@
+Increase md breakpoint value
```
csillag, line 1: Why do we have two changelog fragments, instead one, with two lines? (Is this how it's supposed to work?)

src/styles/theme/defaultTheme.ts
```diff
@@ -82,6 +82,15 @@ declare module '@mui/material/styles' {
 }
 
 export const defaultTheme = createTheme({
+  breakpoints: {
+    values: {
+      xs: 0,
+      sm: 600,
+      md: 1000, // default is 900
```
csillag, line 89: Do we have to override all of them, in order to override this single one? Can't we do something like 
```
values: {
   ...defaultValues,
   md: 900,
}
```

?

.changelog/532.bugfix.2.md
buberdds: I am following one of the comment I've received in the past 
https://github.com/oasisprotocol/oasis-core-ledger/pull/261#discussion_r1174122546

tjanez (approved):

csillag:
Is the W.I.P in the title still relevant?

PR #542, opened 2023-06-15
Follow indexer api changes

csillag:
This is built on top of #541.

I updated the code to follow API changes.

The most notable change is removing support for all token types except ERC-20.

PR #544, opened 2023-06-15
First steps towards displaying contract info

csillag:
~This depends on #542.~

First steps towards supporting contracts:

- When displaying a contract, say so in the title
- Show link to TX that created the contract

For testing, search for address 0x301879c8677e050f0f7104fe6a3059806Abb41C6

For reviewing, I suggest going commit by commit.

### DIFF
--- a/src/app/components/SubPageCard/index.tsx
+++ b/src/app/components/SubPageCard/index.tsx
@@ -11,7 +11,7 @@ import Skeleton from '@mui/material/Skeleton'
 type StyledComponentProps = {
   featured?: boolean
   isLoadingTitle?: boolean
-  title?: string
+  title?: ReactNode
   subheader?: string
   action?: ReactNode
   noPadding?: boolean
--- a/src/locales/en/translation.json
+++ b/src/locales/en/translation.json
@@ -38,6 +38,7 @@
     "block": "Block",
     "bytes": "{{value, number}}",
     "cancel": "Cancel",
+    "createdAt": "Created at",
     "data": "Data",
     "emerald": "Emerald",
     "cipher": "Cipher",
@@ -89,6 +90,9 @@
     "mainnet": "Mainnet",
     "testnet": "Testnet",
     "valuePair": "{{value, number}}"
+  },
+  "contract": {
+    "title": "Contract"
   },
   "nodes": {
     "title": "Active nodes",
--- a/.changelog/544.feature.md
+++ b/.changelog/544.feature.md
@@ -1,0 +1,1 @@
+At address page, recognize contracts, and use appropriate title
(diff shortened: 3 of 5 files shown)


src/locales/en/translation.json
```diff
@@ -1,7 +1,6 @@
 {
   "account": {
-    "cantLoadDetails": "Unfortunately we couldn't load the account details at this time. Please try again later.",
-    "emptyTokenList": "This account holds no {{token}} tokens.",
+      "emptyTokenList": "This account holds no {{token}} tokens.",
```
buberdds, line 3: indent

src/app/pages/AddressDetailsPage/index.tsx
```diff
@@ -31,9 +34,17 @@ export const AddressDetailsPage: FC = () => {
...
+  ) : isContract ? (
+    t('contract.title')
+  ) : (
+    t('account.title')
+  )
+
   return (
     <PageLayout>
-      <SubPageCard featured title={t('account.title')}>
+      <SubPageCard featured title={title as string}>
```
buberdds, line 47: We pass here JSX.Element `<Skeleton`. Let's update SubPageCard prop title to `ReactNode`.

buberdds (commented):
For renaming part let's wait other folks input too.

src/locales/en/translation.json
csillag: Fixed

src/app/pages/AddressDetailsPage/index.tsx
csillag: Yeah, fixed that. (But it turns out we don't even need it here, since SubPageCard can handle displaying the skeleton on it's own, if we just tell it that the title is still loading, so now I'm doing that instead.

csillag:
> For renaming part let's wait other folks input too.

After discussion, I have undone the renaming.

src/app/components/SubPageCard/index.tsx
```diff
@@ -11,7 +11,7 @@ import Skeleton from '@mui/material/Skeleton'
 type StyledComponentProps = {
   featured?: boolean
   isLoadingTitle?: boolean
-  title?: string
+  title?: ReactNode
```
lukaw3d, line 14: Is this for next steps, or leftover?

lukaw3d (approved):

src/app/components/SubPageCard/index.tsx
csillag: Actually, in the end, unrelated but useful. For a while I thought that we need to implement skeleton titles (while loading) externally, and for that, passing in JSX components as title was necessary. But later I realized that `isTitleLoaded` is supported naturally, so no need to use JSX here. But it's still useful for other situations, so I didn't undo the change.

PR #546, opened 2023-06-16
Add Tokens page

csillag:


src/app/components/Tokens/TokenList.tsx
```diff
@@ -0,0 +1,82 @@
...
+import { TablePaginationProps } from '../Table/TablePagination'
+import { AccountLink } from '../Account/AccountLink'
+import { TokenLink } from './TokenLink'
+import { CopyToClipboard } from '../CopyToClipboard'
+
+type TokensProps = {
+  tokens?: EvmToken[]
+  isLoading: boolean
+  limit: number
+  verbose?: boolean
```
buberdds, line 13: It looks like we don't need verbose mode at all.

src/app/components/Tokens/TokenList.tsx
```diff
@@ -0,0 +1,82 @@
...
+    { content: t('tokens.supply'), align: TableCellAlign.Right },
+    { content: t('common.ticker'), align: TableCellAlign.Right },
+  ]
+
+  const tableRows = tokens?.map((token, index) => {
+    return {
+      key: token.contract_addr,
+      data: [
+        {
+          content: (index + 1).toLocaleString(),
```
buberdds, line 37: You need to take an offset into account.

csillag: Right. Fixed.

buberdds (approved):
This is nice start, but what's a plan for tokens feature? Should it be hidden for now so internal releases are "stable"/fully functional or we don't care about internal releases and we are aiming for public launch only?

csillag:
>  but what's a plan for tokens feature? Should it be hidden for now so internal releases are "stable"/fully functional or we don't care about internal releases and we are aiming for public launch only?

I think the indexer has already merged the PRs for extended support (including getting data about the individual tokens etc), so I think we will be flash out the support soon enough. I don't know about the plan for the releases and milestone.

PR #553, opened 2023-06-19
Fix displaying error messages on unsupported layers

csillag:
Before:

image

After:

image

### DIFF
--- a/src/app/hooks/useScopeParam.ts
+++ b/src/app/hooks/useScopeParam.ts
@@ -1,4 +1,4 @@
-import { useParams } from 'react-router-dom'
+import { useParams, useRouteError } from 'react-router-dom'
 import { Network } from '../../types/network'
 import { RouteUtils } from '../utils/route-utils'
 import { AppError, AppErrors } from '../../types/errors'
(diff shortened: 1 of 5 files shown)


buberdds:
when we render layout with unsupported layer, do we need to hide network/ParaTime picker?

csillag:
> when we render layout with unsupported layer, do we need to hide network/ParaTime picker?

In the current for, yes we should, because otherwise it might give us further errors.
However, showing it could be the easiest way back to a valid URL.... but in order to be able to show it, we need to teach it to handle launching with an invalid (or disabled) layer. I don't want to add code changes to it now, since there is another pending PR. 

So for now I will just hide it. User can still click on the Oasis logo and start overf.

csillag:
Updated the code accordingly.

src/app/hooks/useScopeParam.ts
```diff
@@ -10,38 +10,50 @@ export const useNetworkParam = (): Network | undefined => {
...
+    valid: true,
   }
 
-  if (!scope || !RouteUtils.getEnabledNetworks().includes(scope.network))
-    throw new AppError(AppErrors.UnsupportedNetwork)
+  if (network === undefined || layer === undefined) {
+    scope.valid = false
+    if (!error)
+      throw new Error(
+        'You must either specify both network and layer or none of them. You can not have one but not the other.',
```
lukaw3d, line 36: Why are we encoding this here? Isn't it a consequence of routes, not scope param, so add a 404 route instead

PR #554, opened 2023-06-19
Prevent showing horizontal scrollbar on home page

buberdds:
Reported via Internal Launch Feedback.

For me scrollbar is visible always on home page 
current `https://explorer.dev.oasis.io/`
Screenshot 2023-06-19 at 14 10 15

Issue here is that home page layout is using `100vw` width. Combo of `100vw` + vertical scroll + no max width  will cause horizontal scrolls to be visible.

### DIFF
--- a/.changelog/554.bugfix.md
+++ b/.changelog/554.bugfix.md
@@ -1,0 +1,1 @@
+Prevent showing horizontal scrollbar on home page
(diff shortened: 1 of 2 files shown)


csillag:
> Issue here is that home page layout is using 100vw width. Combo of 100vw + vertical scroll + no max width will cause horizontal scrolls to be visible.

Shouldn't we get rid of the vertical scroll here? The presence of the vertical scroll seems to be an error to me, since the main component is not taking the banner into account...

buberdds:
Ah yeah that's another issue I will have to look at.

csillag:
> Ah yeah that's another issue I will have to look at.

Do we need _both_ fixes? (Won't fixing that other issue automatically resolve this as well?)

csillag (approved):
Reviewed. LGTM.

buberdds:
yeah, we will need both probably

PR #562, opened 2023-06-19
Table column content: require key and support ReactNode content

csillag:
Related to https://github.com/oasisprotocol/explorer/pull/505#discussion_r1234493647

### DIFF
--- a/src/app/components/Table/index.tsx
+++ b/src/app/components/Table/index.tsx
@@ -61,7 +61,8 @@ export type TableRowProps = {
 }
 
 export type TableColProps = {
-  content: string
+  key: string
+  content: ReactNode
   align?: TableCellAlign
   width?: string
 }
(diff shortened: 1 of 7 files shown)


src/app/components/Table/index.tsx
```diff
@@ -105,7 +105,7 @@ export const Table: FC<TableProps> = ({
             <TableRow>
               {columns.map((column, index) => (
                 <TableCell
-                  key={column.content}
+                  key={`field-${index}`}
```
lukaw3d, line 108: From discussion: this might be risky

I would rather follow the pattern from table body where rows and cells have `key` field

csillag: > I would rather follow the pattern from table body where rows and cells have key field

Updated to do this.

lukaw3d (approved):

PR #563, opened 2023-06-20
Fix mobile ParaTime picker scroll issue

lubej:
# Problem

Drawer does not calculate the height correctly, when not at the top of the page -> https://github.com/oasisprotocol/explorer/pull/482#issuecomment-1597325615

# Links

- https://www.loom.com/share/1f11a36547984ea0a42739d2790d932b


src/app/hooks/useWindowScroll.ts
```diff
@@ -0,0 +1,17 @@
...
+
+export const useWindowScroll = (): number => {
+  const [windowScrollY, setWindowScrollY] = useState(0)
+
+  useEffect(() => {
+    const windowScrollListener = () => {
+      setWindowScrollY(window.scrollY)
+    }
+
+    window.addEventListener('scroll', windowScrollListener)
```
lukaw3d, line 11: Scroll should almost always be a passive listener

src/app/utils/dom-utils.ts
```diff
@@ -0,0 +1,7 @@
+export const getElementViewportOffset = (el: HTMLElement, elementHeight = el.clientHeight) => {
+  const windowHeight = window.outerHeight,
+    elRect = el.getBoundingClientRect(),
+    top = elRect.top,
+    bottom = elRect.bottom
+  return Math.max(0, top > 0 ? Math.min(elementHeight, windowHeight - top) : Math.min(bottom, windowHeight))
+}
```
lukaw3d, line 7: I overall don't think this feature is worth the added complexity of https://github.com/oasisprotocol/explorer/pull/482 + this.
Just show the picker over the topbar :shrug: -10% of project's complexity

src/app/hooks/useWindowScroll.ts
lubej: Right, it's required, because offset calculation depends on the scroll. So that scroll actually triggers it.

src/app/utils/dom-utils.ts
lubej: I agree on complexity. One simple solution would be to always scroll to the top, before showing the mobile drawer. Because banners are causing the Picker to look like it is misaligned.

src/app/hooks/useWindowScroll.ts
lukaw3d: I don't understand 
```suggestion
    window.addEventListener('scroll', windowScrollListener, { passive: true })
```

lubej: Ah, I thought you meant it should not trigger CD. Nevermind. Let's discuss on sync-up tomorrow.

buberdds:
I think we can close this one.

PR #570, opened 2023-06-20
Fix token supply formatting

lukaw3d:


.changelog/570.bugfix.md
```diff
@@ -0,0 +1 @@
+Add token overview page and dashboard component
```
csillag, line 1: This should be improved

csillag (approved):
LGTM (except the changelog)

.changelog/570.bugfix.md
lukaw3d: This is the same so it can be deduplicated by towncrier (when a new feature and subsequent improvements are within one release)

csillag: Ahh, I see. Btw there is also a `trivial` filename format, which is dropped by towncrier..

lukaw3d: yup. I'll change to that, it's just less confusing

PR #571, opened 2023-06-20
Wrap all long values in description list

lukaw3d:
| Before | After | 
| --- | --- |
| localhost_1234_testnet_sapphire_tx_0x12106a139c9eb799f25b2fd3879827c23c737debefa63da8569560ba556e163c (1) | localhost_1234_testnet_sapphire_tx_0x12106a139c9eb799f25b2fd3879827c23c737debefa63da8569560ba556e163c |

### DIFF
--- a/.changelog/571.bugfix.md
+++ b/.changelog/571.bugfix.md
@@ -1,0 +1,1 @@
+Wrap all long values in detail pages
--- a/src/app/components/StyledDescriptionList/index.tsx
+++ b/src/app/components/StyledDescriptionList/index.tsx
@@ -59,6 +59,7 @@ export const StyledDescriptionList = styled(InlineDescriptionList, {
   },
   dd: {
     color: COLORS.brandExtraDark,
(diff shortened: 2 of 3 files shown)


lukaw3d:
@donouwens thoughts? this seems like a good fallback if mobile+tablet designs don't capture everything

csillag:
Nice, but I don't get why `Encrypted data` looks like that (on the "before" image), since we already have wrapping enabled at https://github.com/oasisprotocol/explorer/blob/42e96f394b137ded663c53b84d798e930b476771/src/app/components/LongDataDisplay/index.tsx#L21

And it does work, down to a specific width, and then stops working. I think that would be worth a separate fix.

lukaw3d:
> @csillag And it does work, down to a specific width, and then stops working. I think that would be worth a separate fix.

Looks like it stops working at the width of tx hash (tx hash didn't have wrapping, and it extended the parent element width)

csillag:
Changelog fragment seems to be missing

PR #574, opened 2023-06-21
Fix BlockDetail verbose flag

lubej:


.changelog/574.bugfix.md
```diff
@@ -0,0 +1 @@
+Fix BlockDetailPage verbose flag
```
buberdds, line 1: We could group this by using the same change log fragment as in main PR.

buberdds (approved):

PR #575, opened 2023-06-21
Make blocks and txn clickable

lubej:
The Blocks and Txn should be clickable without having to first click the ParaTime button.


buberdds (approved):

lukaw3d (approved):

lukaw3d:
Yay

PR #577, opened 2023-06-21
Add tooltip with error message to TransactionStatusIcon if withText=false

lukaw3d:
https://ab5d2c0e.oasis-explorer.pages.dev/mainnet/emerald/block/1637064
image


csillag (approved):
Reviewed. LGTM.

PR #579, opened 2023-06-21
Differentiate user being offline and API being offline

lukaw3d:


src/locales/en/translation.json
```diff
@@ -277,6 +277,7 @@
   },
   "home": {
     "apiOffline": "Our {{ target }} API is offline. We’re trying to reconnect",
+    "userOffline": "You are offline. We’re trying to reconnect",
```
lukaw3d, line 280: @donouwens

src/app/components/OfflineBanner/hook.ts
```diff
@@ -5,9 +5,11 @@ import { AppError, AppErrors } from '../../../types/errors'
 import { useFormattedTimestampString } from '../../hooks/useFormattedTimestamp'
 import { paraTimesConfig } from '../../../config'
 
-export const useIsApiOffline = (network: Network): boolean => {
+export const useIsApiOffline = (network: Network): false | 'userOffline' | 'apiOffline' => {
```
csillag, line 8: At this point I think we should rename this to `useNetworkProblem()`.

csillag (approved):
Reviewed, LGTM. (See comment)

src/app/components/OfflineBanner/hook.ts
lukaw3d: I think that's too confusing: user's "network" && mainnet "network"?

csillag: > isApiReachable?

No, that would make the "false" return value confusing.

> `useConnectionProblem()` ?

lukaw3d: but it's not necessarily a connection problem if API is unresponsive :D

csillag: OK, I am fine with `useIsApiReachable`, but then we need a different result structure. `{ reacheble: true }` when all is fine, and  `{ reachable: false, reason: "whatever" }` in case of issues.

src/app/components/OfflineBanner/index.tsx
```diff
@@ -26,12 +27,19 @@ export const NetworkOfflineBanner: FC<{ wantedNetwork?: Network }> = ({ wantedNe
...
   const targetNetwork = wantedNetwork || scope?.network || Network.mainnet
-  const isNetworkOffline = useIsApiOffline(targetNetwork)
+  const isNetworkReachable = useIsApiReachable(targetNetwork)
   const networkNames = getNetworkNames(t)
   const target = networkNames[targetNetwork]
-  return isNetworkOffline ? (
-    <StyledAlert severity="warning">{t('home.apiOffline', { target })}</StyledAlert>
-  ) : null
+  if (!isNetworkReachable.reachable) {
+    if (isNetworkReachable.reason === 'userOffline') {
```
csillag, line 34: I would prefer adding the full text of reason right inside the hook (we have access to translation there, too); no need to have a layer of compilation here...

src/app/pages/HomePage/index.tsx
```diff
@@ -139,7 +139,7 @@ export const HomePage: FC = () => {
           </LogotypeBox>
           <SearchInputContainer>
             <SearchInputBox>
-              <Search disabled={isApiOffline} variant={searchVariant} onFocusChange={onFocusChange} />
+              <Search disabled={!isApiReachable} variant={searchVariant} onFocusChange={onFocusChange} />
```
csillag, line 142: I think we need to update this to get the inside from within the structure returned by the hook.

lukaw3d: :scream:

PR #587, opened 2023-06-23
Chart buffer

buberdds:
~waiting for Don's feedback~

### DIFF
--- a/src/app/components/charts/BarChart.tsx
+++ b/src/app/components/charts/BarChart.tsx
@@ -21,7 +21,6 @@ interface BarChartProps<T extends object> extends Formatters {
   rounded?: boolean
   withBarBackground?: boolean
   withLabels?: boolean
-  tickMark?: boolean
   margin?: Margin
 }
 
@@ -35,7 +34,6 @@ const BarChartCmp = <T extends object>({
   rounded,
   withLabels,
   withBarBackground,
-  tickMark,
   margin,
 }: BarChartProps<T>) => {
   const { t } = useTranslation()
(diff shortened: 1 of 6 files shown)


src/app/components/charts/BarChart.tsx
```diff
@@ -46,6 +46,7 @@ const BarChartCmp = <T extends object>({
         {cartesianGrid && <CartesianGrid vertical={false} stroke={COLORS.antiFlashWhite3} />}
         {withLabels && (
           <YAxis
+            domain={[0, (dataMax: number) => Math.ceil((dataMax + dataMax * 0.2) / 100) * 100]}
```
lukaw3d, line 49: Do we need extra space on Active Accounts too? yaxis doesn't apply to it

buberdds: From Don: "All, please so TPS, Active Accounts, and the Holders (on the Token page)."

src/app/components/charts/BarChart.tsx
```diff
@@ -21,10 +21,24 @@ interface BarChartProps<T extends object> extends Formatters {
   rounded?: boolean
   withBarBackground?: boolean
   withLabels?: boolean
-  tickMark?: boolean
   margin?: Margin
 }
 
+const calculateCustomTicks = <T extends object>(data: T[], dataKey: keyof T) => {
```
lukaw3d, line 27: How come we need custom tick positions?

buberdds: Still experimenting here and my plan is now to try to avoid custom ticks somehow.

Why I was testing custom ticks: with just a "domain" prop, `YAxis` can be rendered with gaps.  Might be related to `https://github.com/recharts/recharts/issues/2019`. 
Screenshot 2023-06-26 at 13 16 50

lukaw3d: :/

lukaw3d: Reproduced in https://github.com/oasisprotocol/explorer/pull/591 even with `interval="preserveStartEnd"`

lukaw3d (approved):

PR #593, opened 2023-06-26
Save Get Started button state

lubej:


lubej:
Should we merge the states? 🤔

lukaw3d:
Wait, was this planned? I thought "Explore" step was intentionally added to bring focus to search, and so that mobile users can normally scroll the screen, instead of drag event being intercepted by the graph area.

lubej:
> Wait, was this planned? I thought "Explore" step was intentionally added to bring focus to search, and so that mobile users can normally scroll the screen, instead of drag event being intercepted by the graph area.

So the feedback was that "Get Started" button was intrusive. It does make sense on desktop and mobile portrait mode. But you are right, it might look weird on landscape mode on mobile.

lukaw3d (approved):

PR #597, opened 2023-06-27
Fix zoom out link not visible on Testnet

lubej:


src/app/pages/HomePage/Graph/ParaTimeSelector/index.tsx
```diff
@@ -210,12 +216,7 @@ const ParaTimeSelectorCmp: FC<ParaTimeSelectorProps> = ({ disabled, step, setSte
               <ZoomOutBtn
                 onClick={onZoomOutClick}
                 disabled={disabled}
-                sx={{
-                  position: 'absolute',
-                  top: theme.spacing(4),
-                  left: '50%',
-                  transform: 'translateX(-50%)',
-                }}
+                color={network === Network.mainnet ? 'secondary' : 'primary'}
```
buberdds, line 219: We control css colors using themes.

```
export const ZoomOutBtn = styled(Button)(({ theme }) => ({
  color: theme.palette.layout.main,

buberdds (approved):

src/app/pages/HomePage/Graph/ParaTimeSelector/index.tsx
lubej: Right. was looking at it. But it seemed like it was not worth adding another variable just for one case. Added.

buberdds: oops I thought `brandExtraDark` is what we need sorry for a bad example

PR #598, opened 2023-06-27
Update learning materials for Sapphire testnet

buberdds:


src/app/pages/DashboardPage/LearningMaterials.tsx
```diff
@@ -32,30 +33,71 @@ const StyledLink = styled(Link)(() => ({
...
-        header: t('learningMaterials.emerald.header'),
-        link: docs.emerald,
-        title: t('common.emerald'),
-      }
+type LayerContent = {
+  description: string
+  header: string
+  url: string
+}
+type NetworkContent = Partial<Record<Layer, LayerContent[]>>
```
lubej, line 41: Personally I would go with some kind of object, instead of array. This implies that it is possible to have 0,1,2, ... N possible combinations. Can we al least make it `[LayerContent, LayerContent, LayerContent]` type.

lubej (approved):

src/app/pages/DashboardPage/LearningMaterials.tsx
buberdds: OK, switching to an object.

PR #600, opened 2023-06-27
Add Testnet Faucet to snapshot cards

buberdds:


lukaw3d (approved):
(I find the button style a bit odd, but I didn't check mocks)

buberdds:
I will go through a design review now just in case.

buberdds:
Button styles are fine. I am updating ParaTime Snapshot grid columns to match mocks.

PR #606, opened 2023-06-27
Ensure learning materials are specified for every enabled layer

lukaw3d:


lukaw3d:
meh, fails in other lines

PR #607, opened 2023-06-27
Ensure learning materials and search suggestions are specified for every enabled layer

lukaw3d:


src/app/utils/route-utils.ts
```diff
@@ -6,14 +6,23 @@ import { EvmTokenType, Layer } from '../../oasis-indexer/api'
...
+export type SpecifiedPerEnabledLayer<T = any> = {
+  [N in keyof (typeof RouteUtils)['ENABLED_LAYERS_FOR_NETWORK']]: {
+    [L in keyof (typeof RouteUtils)['ENABLED_LAYERS_FOR_NETWORK'][N]]: T
   }
+}
+
+export abstract class RouteUtils {
+  private static ENABLED_LAYERS_FOR_NETWORK = {
+    [Network.mainnet]: {
+      [Layer.emerald]: true,
```
csillag, line 18: Technically, this should be disabled for now, too. (So commented out.)
(It was an oversight that I have accidentally re-enabled it at one point, by leaving in a debug change while merging.)

lukaw3d: Okay ill try to find a sapphire verified contract example first

lukaw3d: I have not found it :shrug:

csillag (approved):
Reviewed, LGTM.

(There or some things that we need to discuss, but not strictly related to this PR, so whatever.)

PR #612, opened 2023-06-28
Remove desktop tooltips on Graph

lubej:


buberdds (approved):
Not sure it's related to these changes, but why I cannot open Cipher (mobile) tooltip?

lubej:
> Not sure it's related to these changes, but why I cannot open Cipher (mobile) tooltip?

Fixed.

PR #618, opened 2023-06-28
Hardcode values of active nodes

buberdds:
Temporary workaround for active nodes issue.
When hardcoded values are merged, we have to create gh issue to revert this once BE is fixed.

~Waiting for Don's feedback, not sure what to show in a tooltip.~

### DIFF
--- a/src/app/pages/DashboardPage/Nodes.tsx
+++ b/src/app/pages/DashboardPage/Nodes.tsx
@@ -11,6 +11,7 @@ import { AppErrors } from '../../../types/errors'
 import { useRequiredScopeParam } from '../../hooks/useScopeParam'
 import Tooltip from '@mui/material/Tooltip'
 import { tooltipDelay } from '../../../styles/theme'
+import { paraTimesConfig } from '../../../config'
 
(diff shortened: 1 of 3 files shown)


src/app/pages/DashboardPage/Nodes.tsx
```diff
@@ -36,7 +35,7 @@ export const Nodes: FC = () => {
       <Tooltip
         arrow
         placement="top"
-        title={hasActiveNodes ? t('nodes.tooltip') : t('nodes.unknown')}
+        title={t('nodes.tooltip')}
```
csillag, line 38: Why not simply set `hasActiveNodes` to true?

csillag: If we did that, it would be easier to revert later. (Less change.)

csillag (approved):
Reviewed, LGTM.

src/app/pages/DashboardPage/Nodes.tsx
buberdds: Good point, I will bring it back.

PR #619, opened 2023-06-28
Handle situations when a Paratime is unavailable

csillag:
Before it was a silent failure

### DIFF
--- a/src/app/components/OfflineBanner/index.tsx
+++ b/src/app/components/OfflineBanner/index.tsx
@@ -30,12 +30,14 @@ export const RuntimeOfflineBanner: FC = () => {
   const scope = useRequiredScopeParam()
   const { t } = useTranslation()
 
-  const { outOfDate, lastUpdate } = useRuntimeFreshness(scope)
-  if (!outOfDate) return null
+  const { outOfDate, lastUpdate, unavailable } = useRuntimeFreshness(scope)
+  if (!outOfDate && !unavailable) return null
   const target = getNameForScope(t, scope)
   return (
     <StickyAlert severity="warning">
-      {lastUpdate
+      {unavailable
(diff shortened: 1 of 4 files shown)


src/app/components/OfflineBanner/index.tsx
```diff
@@ -30,12 +30,14 @@ export const RuntimeOfflineBanner: FC = () => {
...
-  const { outOfDate, lastUpdate } = useRuntimeFreshness(scope)
-  if (!outOfDate) return null
+  const { outOfDate, lastUpdate, unavailable } = useRuntimeFreshness(scope)
+  if (!outOfDate && !unavailable) return null
   const target = getNameForScope(t, scope)
   return (
     <StickyAlert severity="warning">
-      {lastUpdate
+      {unavailable
+        ? t('home.runtimeUnavailable', { target })
```
lukaw3d, line 39: I would reuse `home.apiOffline`

but also, this should not be possible on backend. Can we get backend not to do this, instead?

csillag: > I would reuse `home.apiOffline`

I considered that, but the scope of that warning is too wide to correctly describe this situation. (After all, we _can_ get info about other runtimes...)

> but also, this should not be possible on backend. Can we get backend not to do this, instead?

You are certainly welcome to try to pursue them! I have already brought up the topic on the relevant slack channel, and got the answer that this is more or less considered to be normal... in any case, I think it doesn't have to have multiple lines of defense here.

csillag: Sorry, I miss-typed my message above. I meant to say that it doesn't _hurt_ to have multiple lines of defense here.

lukaw3d (approved):

PR #627, opened 2023-06-29
Update graph graphics

lubej:


src/app/pages/HomePage/Graph/Graph/index.tsx
```diff
@@ -168,216 +172,184 @@ const GraphCmp: ForwardRefRenderFunction<SVGSVGElement, GraphProps> = (
...
+            fill={graphTheme.text}
           />
-          <text x="150.5" y="39" fill={graphTheme.text} fontSize="4px">
```
buberdds, line 265: All these svg paths that renders labels look blurry. Should we keep text nodes?

buberdds (commented):
1) ParaTimes circles are semi transparent and connecting lines are visible. 

2) Why Sapphire and Cipher look like they are enabled, but Emerald don't. This is more question to Don.

src/app/pages/HomePage/Graph/Graph/index.tsx
lubej: Wanted to do this in next PR, as there are new animations, which affect the fonts as well. To keep PRs manageable.

buberdds: ah Ok wasn't aware that we will have follow up updates 👍

buberdds (approved):

PR #628, opened 2023-06-29
Tweak browser back button behaviour

buberdds:
Prevent history.pushState in a few cases.

### DIFF
--- a/src/app/components/Account/ShowMoreTokensLink.tsx
+++ b/src/app/components/Account/ShowMoreTokensLink.tsx
@@ -1,11 +1,11 @@
 import { FC } from 'react'
-import { Link as RouterLink } from 'react-router-dom'
 import { useTranslation } from 'react-i18next'
 import { styled } from '@mui/material/styles'
 import { COLORS } from '../../../styles/theme/colors'
 import { EvmTokenType, RuntimeAccount, type Token } from '../../../oasis-nexus/api'
 import { RouteUtils } from '../../utils/route-utils'
 import { accountTokenContainerId } from '../../pages/AccountDetailsPage/AccountTokensCard'
+import { RouterLink } from '../../components/RouterLink'
 
 export const StyledLink = styled(RouterLink)(({ theme }) => ({
   color: COLORS.brandDark,
@@ -37,7 +37,7 @@ export const ShowMoreTokensLink: FC<ShowMoreTokensLinkProps> = ({ account, token
   }
 
   return (
-    <StyledLink to={erc20link} color="inherit">
+    <StyledLink to={erc20link} color="inherit" replace>
       {t('account.showMore', { counter: additionalTokensCounter })}
     </StyledLink>
   )
--- a/src/app/components/Account/index.tsx
+++ b/src/app/components/Account/index.tsx
@@ -1,6 +1,5 @@
 import { FC } from 'react'
 import { useTranslation } from 'react-i18next'
-import { Link as RouterLink } from 'react-router-dom'
 import { staking } from '@oasisprotocol/client'
 import Box from '@mui/material/Box'
 import { useScreenSize } from '../../hooks/useScreensize'
@@ -22,6 +21,7 @@ import { TokenPriceInfo } from '../../../coin-gecko/api'
 import { ContractCreatorInfo } from './ContractCreatorInfo'
 import { ContractVerificationIcon } from '../ContractVerificationIcon'
 import { TokenLink } from '../Tokens/TokenLink'
+import { RouterLink } from '../RouterLink'
 
 export const StyledAvatarContainer = styled('dt')(({ theme }) => ({
   '&&': {
--- a/.changelog/628.trivial.md
+++ b/.changelog/628.trivial.md
@@ -1,0 +1,1 @@
+Tweak browser back button behaviour
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -40,6 +40,7 @@ const config = {
     ],
     'prefer-template': 'error',
 
+    'react/prop-types': 'off', // Not needed with TS
     'react/jsx-no-target-blank': 'off', // Not needed with modern browsers
     'react/react-in-jsx-scope': 'off', // Not needed after React v17
     'react/display-name': 'off', // TODO: Maybe enable
--- a/src/app/components/Account/TokenPills.tsx
+++ b/src/app/components/Account/TokenPills.tsx
@@ -1,6 +1,5 @@
 import { FC } from 'react'
 import { useTranslation } from 'react-i18next'
-import Link from '@mui/material/Link'
 import Chip from '@mui/material/Chip'
 import Typography from '@mui/material/Typography'
 import { useScreenSize } from '../../hooks/useScreensize'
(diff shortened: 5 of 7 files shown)


csillag (approved):
Reviewed. LGTM.

src/app/components/Account/index.tsx
```diff
@@ -150,7 +150,7 @@ export const Account: FC<AccountProps> = ({ account, isLoading, tokenPriceInfo,
           <dt>{t('common.transactions')}</dt>
           <dd>
             {account.stats.num_txns ? (
-              <Link component={RouterLink} to={transactionsAnchor!}>
+              <Link component={RouterLink} to={transactionsAnchor!} replace>
```
lukaw3d, line 153: Missing same thing on [Block](https://c4126abe.oasis-explorer.pages.dev/mainnet/sapphire/block/575757) > Transactions: 1

lukaw3d, line 153: Bug on search page:
- open https://c4126abe.oasis-explorer.pages.dev/search?q=0xC3ecf872F643C6238Aa20673798eed6F7dA199e9
- click Transactions: 5
- go back

lukaw3d (changes requested):

src/app/components/Account/index.tsx
buberdds: Does `showLayer` prop meant to be used for such comparisons?

lukaw3d: That seems unreliable. I think a maintainable solution would be to automatically `replace` everytime location.pathname+location.search stays the same and only location.hash changes

buberdds: 1) Why controlling this via props passed from `SearchResultsList` is not reliable? 
2) `location.pathname+location.search stays the same and only location.hash changes`. 
but location.pathname can change. user can nav through tabs and click transactions link

csillag: Please note that since merging #634, tab navigation [no longer](https://github.com/oasisprotocol/explorer/pull/634/commits/4befa9eaaa7ffca50dda63072a0c975b386ca494) participates [in auto-scrolling](https://github.com/oasisprotocol/explorer/pull/634/commits/4322a1b7969cbc6f6e726eac6e40324d1687af4d).

lukaw3d: - Not reliable: showLayer has the wrong responsibility, so another prop would be better. But I think it isn't reliable because reusing the component in a different path can introduce a hardly detectable bug. I would rather have a path convention to define this behaviour, instead of each component call site.
- Path: oh, we don't even want to save tabs? I don't see an easily maintainable solution; can we just not do it?

buberdds:
use history.replaceState (replace)

- when user opens account details page and clicks Transactions link 
http://localhost:1234/mainnet/emerald/address/0xCB2412a993f406eFf10a74011410a4F35e3549E3
http://localhost:1234/mainnet/emerald/address/0xCB2412a993f406eFf10a74011410a4F35e3549E3#transactions

- when user opens block details page and clicks Transactions link 
http://localhost:1234/mainnet/emerald/block/6236663
http://localhost:1234/mainnet/emerald/block/6236663#transactions

Needs manual control or ReactLink replace cannot be 1 to 1 match of pathname and search
- when user opens account details page and clicks Show More tokens 
http://localhost:1234/mainnet/emerald/address/0xba504818fdd8d3dba2ef8fd9b4f4d5c71ad1d1d3
http://localhost:1234/mainnet/emerald/address/0xBA504818FdD8D3dBA2Ef8fD9B4F4D5c71aD1d1D3/tokens/erc-20#tokens

- when user opens account details page and clicks a token pill in Tokens row
http://localhost:1234/mainnet/emerald/address/0xba504818fdd8d3dba2ef8fd9b4f4d5c71ad1d1d3
http://localhost:1234/mainnet/emerald/address/0xBA504818FdD8D3dBA2Ef8fD9B4F4D5c71aD1d1D3/tokens/erc-20#oasis1qzk2y3kvp7000qck8zjsmwadhy3lumnmdys20x3z

history.pushState (default behaviour)
TBA

buberdds:
Options: 
1) RouterLink with full match + manual control of replace prop
2) RouterLink with partial match
3) Manual control of replace prop (pass prop through SearchResult), remove RouterLink

lukaw3d:
Is the writeup "current behaviour" or is it the goal. And:

> Needs manual control or ReactLink replace cannot be 1 to 1 match of pathname and search
> - when user opens account details page and clicks Show More tokens
> http://localhost:1234/mainnet/emerald/address/0xba504818fdd8d3dba2ef8fd9b4f4d5c71ad1d1d3
> http://localhost:1234/mainnet/emerald/address/0xBA504818FdD8D3dBA2Ef8fD9B4F4D5c71aD1d1D3/tokens/erc-20#tokens

> history.pushState (default behaviour)
> - when user clicks tabs in Account Details
> http://localhost:1234/mainnet/emerald/address/0xCB2412a993f406eFf10a74011410a4F35e3549E3
> http://localhost:1234/mainnet/emerald/address/0xCB2412a993f406eFf10a74011410a4F35e3549E3/token-transfers#transfers
> http://localhost:1234/mainnet/emerald/address/0xCB2412a993f406eFf10a74011410a4F35e3549E3/tokens/erc-20#tokens

Show More tokens does the same thing as clicking tabs. Is the difference because of lowecase address in first URL?

lukaw3d:
Option 4: RouterLink with full match + convert tab paths to hash

src/app/components/Account/ShowMoreTokensLink.tsx
```diff
@@ -37,7 +37,7 @@ export const ShowMoreTokensLink: FC<ShowMoreTokensLinkProps> = ({ account, token
   }
 
   return (
-    <StyledLink to={erc20link} color="inherit">
+    <StyledLink to={erc20link} color="inherit" replace>
```
lukaw3d, line 40: Why manual replace?
- https://cd976a58.oasis-explorer.pages.dev/search?q=0x404E70A162487c9Af8982a89a5453f389d5257b1
- Click "+ 1 more"
- Click back

buberdds:
> Option 4: RouterLink with full match + convert tab paths to hash

how this can work with `tokens/erc-20#oasis1qpyyglghqwzfqpwuvcmvehkxc24r8f5leq3cp3mk`  ?

lukaw3d:
> > Option 4: RouterLink with full match + convert tab paths to hash
> 
> how this can work with `tokens/erc-20#oasis1qpyyglghqwzfqpwuvcmvehkxc24r8f5leq3cp3mk` ?

we would need a custom hash syntax :/ And it becomes odd when pagination is supposed to be kept in history, but tabbing isn't
`/token-transfers?page=2` -> `?page=2#/token-transfers`

src/app/components/Account/ShowMoreTokensLink.tsx
buberdds: ahh yeah that works only in Account details context. 
Moreover, option 3 `Manual control of replace prop` would be much more messy. I did not think about tokens.

PR #629, opened 2023-06-29
Improve how tabs are displayed on small screens

csillag:
Our Cards have rounded corners, which is nice, but when used under a tab menu, on mobile, sometimes it causes a problem. Look at just below the tabs, at the edges:

| Before | After |
| ------ | ------
| image | image  |

### DIFF
--- a/src/styles/theme/defaultTheme.ts
+++ b/src/styles/theme/defaultTheme.ts
@@ -738,12 +738,15 @@ export const defaultTheme = createTheme({
     },
     MuiTabs: {
       styleOverrides: {
-        root: {
+        root: ({ theme }) => ({
           // neglect the default border radius of sibling element (Card component in most cases)
           '&& + *': {
             borderTopLeftRadius: 0,
+            [theme.breakpoints.down('sm')]: {
+              borderTopRightRadius: 0,
+            },
           },
-        },
+        }),
         indicator: {
           display: 'none',
         },
--- a/.changelog/629.trivial.md
+++ b/.changelog/629.trivial.md
@@ -1,0 +1,1 @@
+Imrove how tabs are displayed on small screens


lukaw3d:
typo in PR name and commit

src/styles/theme/defaultTheme.ts
```diff
@@ -738,12 +738,15 @@ export const defaultTheme = createTheme({
     },
     MuiTabs: {
       styleOverrides: {
-        root: {
+        root: ({ theme }) => ({
           // neglect the default border radius of sibling element (Card component in most cases)
           '&& + *': {
             borderTopLeftRadius: 0,
+            [theme.breakpoints.down('sm')]: {
+              borderTopRightRadius: 0,
```
lukaw3d, line 746: this is quite hacky/unreliable. It's a heuristic that tab names will be longer than tab content on small displays

@donouwens is it okay when it fails like this:
| Before | After |
| --- | --- |
| 7fd5532d oasis-explorer pages dev_mainnet_emerald_address_0xCB2412a993f406eFf10a74011410a4F35e3549E3 (1)  |  7fd5532d oasis-explorer pages dev_mainnet_emerald_address_0xCB2412a993f406eFf10a74011410a4F35e3549E3 |

csillag: > this is quite hacky/unreliable. It's a heuristic that tab names will be longer than tab content on small displays

My intention was simply to get rid of rounded corners on the top of the cars under the menu, when on mobile. This is independent from the length of the names. I think on a small screen there is just too much stuff going on near the top of the tab menu; even if we don't see the anomaly (blue/white mixing), I don't consider it a loss that there is no rounded corner there. 

So I don't think it "fails" in the above test case; I think it's a welcome simplification.

lukaw3d (approved):

PR #632, opened 2023-06-30
Rename indexer to nexus

csillag:
Now that the Oasis Indexer has officially been renamed to Oasis Nexus, it's time to update our code and documentation to reflect that.

Apparently the renaming is a gradual process; the self-documenting API is still referring to itself as indexer at lots of places, I assume this will be sorted out during the next few Nexus releases.

Also, the domain names have not been changed yet, so we are keeping the old ones, for now.


src/oasis-indexer/orval.config.js
```diff
@@ -3,10 +3,10 @@ import removeNetworkFromName from './removeNetworkFromName.js'
...
 /** @type {import('@orval/core').Config} */
 const config = {
-  indexer: {
+  nexus: {
     input: {
       // target: './v1.yaml',
-      // target: 'https://raw.githubusercontent.com/oasisprotocol/oasis-indexer/main/api/spec/v1.yaml',
+      // target: 'https://raw.githubusercontent.com/oasisprotocol/nexus/main/api/spec/v1.yaml',
       // target: 'https://index-staging.oasislabs.com/v1/spec/v1.yaml',
       target: 'https://index.oasislabs.com/v1/spec/v1.yaml',
```
csillag, line 11: I assume the domain will be renamed at some point too, but for now, we need to stick with the old name.

csillag:
I have updated the CI scripts, too, but it hasn't been tested on GitHub yet. I guess if the action for codegen is broken, we will just have to fix it after merging.

lukaw3d (approved):

PR #634, opened 2023-06-30
Token dashboard: add "Token Transfers" and "Code" tabs

csillag:
This PR adds 2 of the 3 tabs required for the token dashboard:
 - Token transfers
 - Code

Designs: 
 - [Token transfers](https://www.figma.com/file/ifCrok8cP5ymEYjMa2PIi9/Block-Explorer?type=design&node-id=5840-215517)
 - [Code](https://www.figma.com/file/ifCrok8cP5ymEYjMa2PIi9/Block-Explorer?type=design&node-id=5967-211710)

The last tab, token holders, comes later.

### DIFF
--- a/src/app/components/NonScrollingRouterLink/index.tsx
+++ b/src/app/components/NonScrollingRouterLink/index.tsx
@@ -1,0 +1,6 @@
+import React, { ComponentPropsWithRef, FC } from 'react'
+import { Link } from 'react-router-dom'
+
+export const NonScrollingRouterLink: FC<ComponentPropsWithRef<typeof Link>> = React.forwardRef(
+  (props, ref) => <Link {...props} preventScrollReset ref={ref} />,
+)
--- a/.changelog/634.feature.md
+++ b/.changelog/634.feature.md
@@ -1,0 +1,1 @@
+Token dashboard: add "Token Transfers" and "Code" tabs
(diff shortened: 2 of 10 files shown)


src/app/components/NonScrollingRouterLink/index.tsx
```diff
@@ -0,0 +1,6 @@
+import React, { FC } from 'react'
+import { Link } from 'react-router-dom'
+
+export const NonScrollingRouterLink: FC<Parameters<typeof Link>[0]> = React.forwardRef((props, ref) => (
```
buberdds, line 4: You can make it cleaner by using utility type `FC<ComponentPropsWithRef<typeof Link>>`.

buberdds (approved):

src/app/components/NonScrollingRouterLink/index.tsx
csillag: Thanks for the tip, done.

lukaw3d: Do we have further plans for this abstraction, or is it just a longer way of adding "preventScrollReset"?

csillag: As far as I am concerned, it serves a dual purpose:
 - Avoid having to spell out `preventScrollReset={true}`
 - Oftentimes we want to import ReactRouter's `Link` as `RouterLink`, since we also want to work with Mui's `Link` in the same namespace. This is a bit tedious, and requires manual editing (no automati imports.) However, using `NoScrollingRouterLink` avoids this problem. (Along the same lines, I think it would also make sense to create an alias to React Router's `Link` called `RouterLink`. It would make it easier to import.)

PR #623, opened 2023-06-29
Implement top part of token dashboard

csillag:
The token dashboard is what we find when visiting `/:network:/:runtime:/token/:address:`, for example [like this](https://86e7b442.oasis-explorer.pages.dev/mainnet/sapphire/token/0xA64b0ED9789824CBdA2E0a971190EC28347BBC4e).

This is different from the normal account / contract page. (Although they link to each other.)

The PR implements the top and middle part of the token dashboard. The bottom part (with multiple tabs) is coming later.

Designs:
 - [Desktop](https://www.figma.com/file/ifCrok8cP5ymEYjMa2PIi9/MVP?type=design&node-id=5815%3A213056&t=7Scpdn0TgvO6uvMa-1)
 - [Mobile](https://www.figma.com/file/ifCrok8cP5ymEYjMa2PIi9/MVP?type=design&node-id=5959%3A226256&t=7Scpdn0TgvO6uvMa-1)

Also, when displaying info about a contract
 * Properly display creator info (see the design [here](https://www.figma.com/file/ifCrok8cP5ymEYjMa2PIi9/Block-Explorer?type=design&node-id=6092-228146&mode=design&t=xsoK0xQ7yBBphvVx-0) ) -- test with `0x301879c8677e050f0f7104fe6a3059806Abb41C6`
 * If the contract is about a token, link to the relevant token dashboard (see on the above visual) -- test with `0x6Cb9750a92643382e020eA9a170AbB83Df05F30B`

PR #635, opened 2023-06-30
Add "Token Holders" tab to Token Dashboard

csillag:
This ~is built on top of #634, and~ implements the last missing tab, "Token Holders".

Design is [here](https://www.figma.com/file/ifCrok8cP5ymEYjMa2PIi9/Block-Explorer?type=design&node-id=5967-211045).

### DIFF
--- a/src/app/components/Tokens/TokenHolders.tsx
+++ b/src/app/components/Tokens/TokenHolders.tsx
@@ -13,7 +13,7 @@ type TableTokenHolder = BareTokenHolder & {
 type TokenHoldersProps = {
   holders: TableTokenHolder[] | undefined
   isLoading: boolean
-  decimals: number | undefined
+  decimals: number
   totalSupply: string | undefined
   limit: number
   pagination: false | TablePaginationProps
@@ -24,7 +24,7 @@ export const TokenHolders: FC<TokenHoldersProps> = ({
   limit,
   pagination,
   holders,
-  decimals = 18,
+  decimals,
   totalSupply,
 }) => {
   const { t } = useTranslation()
--- a/src/app/components/Tokens/TokenHolders.tsx
+++ b/src/app/components/Tokens/TokenHolders.tsx
@@ -1,0 +1,108 @@
+import { FC } from 'react'
+import { useTranslation } from 'react-i18next'
+import { Table, TableCellAlign, TableColProps } from '../Table'
+import { BareTokenHolder } from '../../../oasis-nexus/api'
+import { TablePaginationProps } from '../Table/TablePagination'
+import { AccountLink } from '../Account/AccountLink'
+import { fromBaseUnits } from '../../utils/helpers'
+import { TokenPriceInfo } from '../../../coin-gecko/api'
+
+type TableTokenHolder = BareTokenHolder & {
+  markAsNew?: boolean
+}
+
+type TokenHoldersProps = {
+  holders: TableTokenHolder[] | undefined
+  isLoading: boolean
+  decimals: number | undefined
+  totalSupply: string | undefined
+  limit: number
+  tokenPrice?: TokenPriceInfo
+  pagination: false | TablePaginationProps
+}
+
+export const TokenHolders: FC<TokenHoldersProps> = ({
+  isLoading,
+  limit,
+  pagination,
+  holders,
+  decimals = 18,
+  totalSupply,
+  tokenPrice,
+}) => {
+  const { t } = useTranslation()
+  const tableColumns: TableColProps[] = [
+    { key: 'rank', content: t('common.rank'), align: TableCellAlign.Center },
+    { key: 'address', content: t('common.address') },
+    { key: 'quantity', content: t('common.quantity'), align: TableCellAlign.Right },
+    { key: 'percentage', content: t('common.percentage'), align: TableCellAlign.Right },
+    ...(tokenPrice
+      ? [{ key: 'value', align: TableCellAlign.Right, content: t('common.value'), width: '250px' }]
+      : []),
+  ]
+
+  const calculateRatio = (balance: string): string => {
+    return totalSupply === undefined
+      ? t('common.missing')
+      : `${((100 * parseFloat(fromBaseUnits(balance, decimals))) / parseFloat(totalSupply)).toFixed(4)}%`
+  }
+
+  const tableRows = holders?.map((holder, index) => {
+    return {
+      key: holder.holder_address,
+      data: [
+        {
+          key: 'rank',
+          content: holder.rank.toLocaleString(),
+          align: TableCellAlign.Center,
+        },
+        {
+          key: 'address',
+          content: (
+            <AccountLink scope={holder} address={holder.eth_holder_address || holder.holder_address} />
+          ),
+        },
+        {
+          key: 'quantity',
+          content: t('tokens.totalSupplyValue', { value: fromBaseUnits(holder.balance, decimals) }),
+          align: TableCellAlign.Right,
+        },
+        {
+          key: 'percentage',
+          content: calculateRatio(holder.balance),
+          align: TableCellAlign.Right,
+        },
+        ...(tokenPrice?.price
+          ? [
+              {
+                key: 'value',
+                content: t('common.fiatValueInUSD', {
+                  value: parseFloat(fromBaseUnits(holder.balance, decimals)) * tokenPrice.price,
+                  formatParams: {
+                    value: {
+                      currency: 'USD',
+                    } satisfies Intl.NumberFormatOptions,
+                  },
+                }),
+
+                align: TableCellAlign.Right,
+              },
+            ]
+          : []),
+      ],
+      highlight: holder.markAsNew,
+    }
+  })
+
+  return (
+    <Table
+      columns={tableColumns}
+      rows={tableRows}
+      rowsNumber={limit}
+      name={t('transactions.latest')}
+      isLoading={isLoading}
+      pagination={pagination}
+      extraHorizontalSpaceOnMobile
+    />
+  )
+}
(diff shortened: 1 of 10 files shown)


src/app/components/Tokens/TokenHolders.tsx
```diff
@@ -0,0 +1,113 @@
...
+        },
+        {
+          key: 'address',
+          content: (
+            <AccountLink scope={holder} address={holder.eth_holder_address || holder.holder_address} />
+          ),
+        },
+        {
+          key: 'quantity',
+          content: t('tokens.totalSupplyValue', { value: fromBaseUnits(holder.balance, decimals) }),
```
buberdds, line 72: Shouldn't we use RoundedBalance here for example 
`http://localhost:1234/mainnet/sapphire/token/0x4344919960B0196c512807eE2f3922302fFFE542/holders#holders`

```
-- | --
1 | 0x3BA9F711C9808902f6a200d3715ba8a7e4D8b63f | 100 | 100.0000%
2 | 0x00000000000000000000000000000000000000ff | 0 | 0.0000%
```

we show 0 in the second row for example

src/app/components/Tokens/TokenHolders.tsx
```diff
@@ -0,0 +1,113 @@
...
+
+  const calculateRatio = (balance: string): string => {
+    return totalSupply === undefined
+      ? t('common.missing')
+      : `${((100 * parseFloat(fromBaseUnits(balance, decimals))) / parseFloat(totalSupply)).toFixed(4)}%`
+  }
+
+  const tableRows = holders?.map((holder, index) => {
+    return {
+      key: `event-${index + (pagination ? (pagination.selectedPage - 1) * pagination.rowsPerPage : 0)}`,
```
buberdds, line 53: isn't  address unique in token holders. Can we use it in keys instead of trying to make our own unique values?

src/app/components/Tokens/TokenHolders.tsx
```diff
@@ -0,0 +1,113 @@
...
+  isLoading,
+  limit,
+  pagination,
+  holders,
+  decimals = 18,
+  totalSupply,
+  tokenPrice,
+}) => {
+  const { t } = useTranslation()
+  // const { isMobile } = useScreenSize()
```
buberdds, line 34: leftover

src/app/components/Tokens/TokenHolders.tsx
```diff
@@ -0,0 +1,113 @@
...
+  tokenPrice?: TokenPriceInfo
+  pagination: false | TablePaginationProps
+}
+
+export const TokenHolders: FC<TokenHoldersProps> = ({
+  isLoading,
+  limit,
+  pagination,
+  holders,
+  decimals = 18,
```
buberdds, line 29: It does not match all ParaTimes. Can we remove this default?

src/app/components/Tokens/TokenHolders.tsx
```diff
@@ -0,0 +1,113 @@
...
+          key: 'percentage',
+          content: calculateRatio(holder.balance),
+          align: TableCellAlign.Right,
+        },
+        ...(tokenPrice?.price
+          ? [
+              {
+                key: 'value',
+                content: t('common.fiatValueInUSD', {
+                  value: parseFloat(fromBaseUnits(holder.balance, decimals)) * tokenPrice.price,
```
buberdds, line 85: imo this was discussed and we won't support token price at all. Can we kill all `tokenPrice` pieces? I think Don said he removed price related stuff in "Public Launch" designs.

src/app/components/Tokens/TokenHolders.tsx
csillag: Yes of course you are right. Fixed.

src/app/components/Tokens/TokenHolders.tsx
csillag: Removed

src/app/components/Tokens/TokenHolders.tsx
csillag: It was fully conditional code, not active ATM, but sure, we can remove it.

csillag: Removed

src/app/components/Tokens/TokenHolders.tsx
csillag: The problem is that the decimals field is [optional](https://github.com/oasisprotocol/nexus/blob/main/api/spec/v1.yaml#L2468) on tokens. So I guess in case it's missing, we need to look at the paratime config....

csillag: Done

src/app/components/Tokens/TokenHolders.tsx
csillag: > Shouldn't we use RoundedBalance here for Quantity? 

OK done

csillag: > Not sure about Percentage. for example `http://localhost:1234/mainnet/sapphire/token/0x4344919960B0196c512807eE2f3922302fFFE542/holders#holders`
> 
> ```
> -- | --
> 1 | 0x3BA9F711C9808902f6a200d3715ba8a7e4D8b63f | 100 | 100.0000%
> 2 | 0x00000000000000000000000000000000000000ff | 0 | 0.0000%
> ```
> 
> we show 0 in the second row

I don't see the issue here. The percentage must _obviously_ be a rounded value, so what else would anyone expect?

csillag: OK, so I think we are good here.

lukaw3d: I would expect
da222769 oasis-explorer pages dev_mainnet_sapphire_token_0x4344919960B0196c512807eE2f3922302fFFE542_holders (1)
not
da222769 oasis-explorer pages dev_mainnet_sapphire_token_0x4344919960B0196c512807eE2f3922302fFFE542_holders

src/app/pages/TokenDashboardPage/TokenHoldersCard.tsx
```diff
@@ -1,40 +1,64 @@
...
+    <Card>
+      <LinkableDiv id={tokenHoldersContainerId}>
+        <CardHeader disableTypography component="h3" title={t('tokens.holders')} />
+      </LinkableDiv>
+      <CardContent>
+        <ErrorBoundary light={true}>
+          {isFetched && !totalCount && <CardEmptyState label={t('tokens.emptyTokenHolderList')} />}
+          <TokenHolders
+            holders={holders}
+            decimals={token?.decimals ?? paratimeDefaultDecimals}
```
lukaw3d, line 48: https://github.com/oasisprotocol/explorer/pull/623#discussion_r1247196333

src/app/components/Tokens/TokenHolders.tsx
```diff
@@ -0,0 +1,90 @@
...
+        },
+        {
+          key: 'address',
+          content: (
+            <AccountLink scope={holder} address={holder.eth_holder_address || holder.holder_address} />
+          ),
+        },
+        {
+          key: 'quantity',
+          content: <RoundedBalance value={fromBaseUnits(holder.balance, decimals)} />,
```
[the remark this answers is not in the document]
csillag: Well it's problematic, because in api.is, when parsing the results of the get token holders call, we don't have access to the definition of the token, which defines the number of decimals. _Maybe_ we could make it into a required argument of the wrapper function call, but then it would change the signature, which is not nice, and also semantically doesn't make all that much sense.

src/app/components/Tokens/TokenHolders.tsx
```diff
@@ -0,0 +1,90 @@
...
+  const tableRows = holders?.map((holder, index) => {
+    return {
+      key: holder.holder_address,
+      data: [
+        {
+          key: 'rank',
+          content: (
+            index +
+            (pagination ? (pagination.selectedPage - 1) * pagination.rowsPerPage : 0) +
+            1
```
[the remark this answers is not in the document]
csillag: We could add this in api, which returns the holders ordered by decreasing ratio.

src/app/components/Tokens/TokenHolders.tsx
```diff
@@ -0,0 +1,90 @@
...
+import { AccountLink } from '../Account/AccountLink'
+import { fromBaseUnits } from '../../utils/helpers'
+import { RoundedBalance } from '../RoundedBalance'
+
+type TableTokenHolder = BareTokenHolder & {
+  markAsNew?: boolean
+}
+
+type TokenHoldersProps = {
+  holders?: TableTokenHolder[]
```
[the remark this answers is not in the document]
csillag: OK fixed

src/app/pages/TokenDashboardPage/TokenHoldersCard.tsx
csillag: OK fixed

src/app/components/Tokens/TokenHolders.tsx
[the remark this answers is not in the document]
csillag: OK, moved rank calculation into API code

src/app/components/Tokens/TokenHolders.tsx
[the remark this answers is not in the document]
csillag: @lukaw3d , would you prefer if I added decimals as a required parameter to this call, so that I can use it to normalize the data in the API layer?

src/oasis-nexus/api.ts
```diff
@@ -532,3 +537,36 @@ export const useGetRuntimeEvents: typeof generated.useGetRuntimeEvents = (
...
+      transformResponse: [
+        ...arrayify(axios.defaults.transformResponse),
+        (data: generated.TokenHolderList, headers, status) => {
+          if (status !== 200) return data
+          return {
+            ...data,
+            holders: data.holders.map((holder, index) => {
+              return {
+                ...holder,
+                rank: index + (params?.offset || 0) + 1,
```
lukaw3d, line 561: `?? 0` is slightly more accurate

src/app/components/Tokens/TokenHolders.tsx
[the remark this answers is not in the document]
lukaw3d: no, rather keep it as is

lukaw3d (approved):

src/oasis-nexus/api.ts
csillag: OK updated

PR #643, opened 2023-07-03
Show precise tooltip for "<0.00000" values too

lukaw3d:


buberdds (approved):

csillag (approved):
All good, except the missing freakin' changelog fragment. 
(Which could easily be generated from the commit message. Where is the AI when we need it?)

PR #654, opened 2023-07-04
Update ParaTime snapshot charts 

buberdds:
~waiting for Don's feedback~

- show last data point value in a snapshot card
- add new label for chart durations
- update dropdown labels
- replace volume_per_second with tx_volume in txs chart

### DIFF
--- a/src/app/components/Snapshots/SnapshotCardDurationLabel.tsx
+++ b/src/app/components/Snapshots/SnapshotCardDurationLabel.tsx
@@ -1,0 +1,34 @@
+import { FC } from 'react'
+import { TFunction } from 'i18next'
+import { useTranslation } from 'react-i18next'
+import Box from '@mui/material/Box'
+import Typography from '@mui/material/Typography'
+import { COLORS } from '../../../styles/theme/colors'
+import { ChartDuration } from '../../utils/chart-utils'
+
+const getLabels = (t: TFunction): Record<ChartDuration, string> => ({
+  [ChartDuration.TODAY]: t('chartDuration.lastHour'),
+  [ChartDuration.WEEK]: t('chartDuration.lastDay'),
+  [ChartDuration.MONTH]: t('chartDuration.lastDay'),
+  [ChartDuration.ALL_TIME]: t('chartDuration.lastDay'),
+})
+
+type SnapshotCardLabelProps = {
+  duration: ChartDuration
+  value: number | undefined
+}
+export const SnapshotCardDurationLabel: FC<SnapshotCardLabelProps> = ({ duration, value }) => {
+  const { t } = useTranslation()
+  const labels = getLabels(t)
+
+  if (!value) {
+    return null
+  }
+
+  return (
+    <Box sx={{ display: 'flex', alignItems: 'center', gap: 3 }}>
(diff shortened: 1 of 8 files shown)


src/locales/en/translation.json
```diff
@@ -284,21 +284,21 @@
...
   "select": {
     "placeholder": "Select"
   },
   "chartDuration": {
-    "today": "Today",
-    "thisWeek": "This week",
-    "thisMonth": "This month",
+    "today": "Past day",
+    "thisWeek": "Past week",
+    "thisMonth": "Past month",
```
lubej, line 296: Should we change the keys as well?

lubej (approved):
LGTM

src/app/components/Snapshots/SnapshotCardDurationLabel.tsx
```diff
@@ -0,0 +1,34 @@
...
+import Typography from '@mui/material/Typography'
+import { COLORS } from '../../../styles/theme/colors'
+import { ChartDuration } from '../../utils/chart-utils'
+
+const getLabels = (t: TFunction): Record<ChartDuration, string> => ({
+  [ChartDuration.TODAY]: t('chartDuration.lastHour'),
+  [ChartDuration.WEEK]: t('chartDuration.lastDay'),
+  [ChartDuration.MONTH]: t('chartDuration.lastDay'),
+  [ChartDuration.ALL_TIME]: t('chartDuration.lastDay'),
+})
```
lukaw3d, line 14: This looks wrong, so it should have an explanatory name or comment

src/app/components/Snapshots/SnapshotCardDurationLabel.tsx
```diff
@@ -0,0 +1,34 @@
...
+import Box from '@mui/material/Box'
+import Typography from '@mui/material/Typography'
+import { COLORS } from '../../../styles/theme/colors'
+import { ChartDuration } from '../../utils/chart-utils'
+
+const getLabels = (t: TFunction): Record<ChartDuration, string> => ({
+  [ChartDuration.TODAY]: t('chartDuration.lastHour'),
+  [ChartDuration.WEEK]: t('chartDuration.lastDay'),
+  [ChartDuration.MONTH]: t('chartDuration.lastDay'),
+  [ChartDuration.ALL_TIME]: t('chartDuration.lastDay'),
```
lukaw3d, line 13: Hm, looks like Active Accounts shows value for last month:
image

buberdds: Oh yeah, will ask Don what he wants here. It's weird that we have different behaviours for "all time" filter.

buberdds: Ok, charts behave correctly according to Don. 
"Exactly, they don’t have to be the same because we are showing different increments in the two charts"
We want two diff labels.

lukaw3d: I would rather switch to showing both daily, since we also can't correctly aggreggate monthly active accounts :shrug:

PR #656, opened 2023-07-04
Help to avoid sending over the mnemonic to nexus for search

csillag:
Up to now, we could only search for strings in a few very strictly specified format. (Ie. block number, tx hash, etc.) But now that we can do free full-text search for token names, it has become a bit of a security vulnerability that it's possible to accidentally copy and send sensitive information, like your wallet private key mnemonics, if you are working with the wallet and the explorer at the same time, and accidentally copy the wrong string to the wrong field.

Although would probably not be an everyday event, it's worth to add a simple check to warn about this, before doing it.

If it happens, this is what we do:

image

### DIFF
--- a/src/app/components/Search/index.tsx
+++ b/src/app/components/Search/index.tsx
@@ -19,6 +19,7 @@ import { SearchScope } from '../../../types/searchScope'
 import { textSearchMininumLength } from './search-utils'
 import Typography from '@mui/material/Typography'
 import { isValidBlockHeight } from '../../utils/helpers'
+import { isValidMnemonic } from '../../utils/helpers'
 
 export type SearchVariant = 'button' | 'icon' | 'expandable'
 
@@ -105,14 +106,21 @@ const SearchCmp: FC<SearchProps> = ({ scope, variant, disabled, onFocusChange: o
   const [isFocused, setIsFocused] = useState(false)
   const valueInSearchParams = useSearchParams()[0].get('q') ?? ''
 
-  const valueWithoutPrefix = value.trim()
+  const wordsOfPower = t('search.wordsOfPower')
+  const hasWordsOfPower = value.trim().toLowerCase().startsWith(wordsOfPower.toLowerCase())
+  const valueWithoutPrefix = hasWordsOfPower
+    ? value.trim().substring(wordsOfPower.length).trim()
+    : value.trim()
 
   const isTooShort =
     !!value && valueWithoutPrefix.length < textSearchMininumLength && !isValidBlockHeight(valueWithoutPrefix)
 
+  const hasPrivacyProblem = !hasWordsOfPower && isValidMnemonic(valueWithoutPrefix)
+
   useEffect(() => {
-    setValue(valueInSearchParams)
-  }, [valueInSearchParams])
+    setValue(hasWordsOfPower ? `${wordsOfPower} ${valueInSearchParams}` : valueInSearchParams)
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [valueInSearchParams]) // We only want to update the value from code when the URL changes
 
   const onChange = (newValue: string) => {
     setValue(newValue)
@@ -146,7 +154,13 @@ const SearchCmp: FC<SearchProps> = ({ scope, variant, disabled, onFocusChange: o
   const searchButtonContent =
     variant !== 'button' ? <SearchIcon sx={{ color: COLORS.grayMediumLight }} /> : t('search.searchBtnText')
 
-  const hasError = isTooShort
+  const errorMessage = isTooShort
+    ? t('search.error.tooShort')
+    : hasPrivacyProblem
+    ? t('search.error.privacy', { appName: t('appName'), wordsOfPower })
+    : undefined
+
+  const hasError = !!errorMessage
 
   return (
     <SearchForm
@@ -195,7 +209,7 @@ const SearchCmp: FC<SearchProps> = ({ scope, variant, disabled, onFocusChange: o
           value &&
           value !== valueInSearchParams && (
             <div>
-              {isTooShort && (
+              {hasError && (
                 <>
                   <Typography
                     component="span"
@@ -211,7 +225,7 @@ const SearchCmp: FC<SearchProps> = ({ scope, variant, disabled, onFocusChange: o
                   >
                     <ErrorIcon />
                     &nbsp;
-                    {t('search.error.tooShort')}
+                    {errorMessage}
                   </Typography>
                   <br />
                 </>
--- a/src/locales/en/translation.json
+++ b/src/locales/en/translation.json
@@ -1,4 +1,5 @@
 {
+  "appName": "Oasis Explorer",
   "account": {
     "cantLoadDetails": "Unfortunately we couldn't load the account details at this time. Please try again later.",
     "emptyTokenList": "This account holds no {{token}} tokens.",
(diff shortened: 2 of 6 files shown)


nhynes:
Wait, why don't you instead detect the mnemonic and send it to your personal computer? You're just leaving money on the table!

csillag:
> Wait, why don't you instead detect the mnemonic and send it to your personal computer? You're just leaving money on the table!

That will come in a subsequent change, when the buzz around explorer has already faded... :grin:

lukaw3d (commented):
I don't think most of this is worth the maintenance burden. I definitely don't want TFunction inside routes :/ and sending slightly invalid mnemonic isn't safe either

csillag:
> I don't think most of this is worth the maintenance burden. I definitely don't want TFunction inside routes :/

If we are OK with having to say "I command thee" in English, independently of i18n, then we can get rid of most of the extra complexity, including the one impacting the routes.

> sending slightly invalid mnemonic isn't safe either

My guees would be that this typically happens using copy-paste. Copy-paste doesnt usually introduce slight errors...

csillag:
> If we are OK with having to say "I command thee" in English, independently of i18n, then we can get rid of most of the extra complexity, including the one impacting the routes.

I simplified the implementation.
| before | after |
|----- | ---- |
|  image | image   |

The routing code is no longer impacted in any way.

csillag:
> I don't think most of this is worth the maintenance burden. 

@lukaw3d how about the new, simplified implementation? (With no TFunction creep)

csillag:
Wait, I have another idea...

csillag:
> Wait, I have another idea...

OK, we are now down to 
image

csillag:
I have now refactored this so that we catch it sooner: right at the search form itself. This has several benefits:
- We can stop the sensitive data from being added to the URL, and hence the browser history.
- Architecturally, it's must more simple. (We don't have to channel the status or error message anywhere, because we display it right where we detect it)
- Since we are in the UI layer, we can again i18n the Words of Power properly.

tjanez (commented):
I think this is a nice safety measure on our side, :+1:

csillag:
> I like this a lot, but I do believe the text is too long - I doubt users would read the whole thing.

Well they really should, if they really want to do something this stupid...

> Maybe we can create an integration for the second one that works in the same way as the search suggestions under the search bar ?

Do you suggest adding it to the beginning the the present message (linked on Figma), or display the message _instead of the current one_ when trying to search for something with only 1 or 2 characters?

csillag:
Meanwhile we are missing the 0.2.0 🚄 train  😿

csillag:
> > I like this a lot, but I do believe the text is too long - I doubt users would read the whole thing.

> Well they really should, if they really want to do something this stupid...

Would adding paragraphs help with that?

csillag:
> If a user is doing this (which arguably is stupid) it would even give us more reason to make the message shorter and clearer IMO. A stupid user definitely won't be reading this.

OK, rephreasing in idiocracy lingo.

src/app/components/Search/index.tsx
```diff
@@ -99,11 +102,22 @@ const SearchCmp: FC<SearchProps> = ({ scope, variant, disabled, onFocusChange: o
   const searchPlaceholderTranslated = isMobile ? t('search.mobilePlaceholder') : t('search.placeholder')
   const [value, setValue] = useState('')
   const [isFocused, setIsFocused] = useState(false)
+  const [isTooShort, setIsTooShort] = useState(false)
```
lukaw3d, line 105: Can you separate "isTooShort" into a small PR

src/app/components/Search/index.tsx
```diff
@@ -113,7 +127,20 @@ const SearchCmp: FC<SearchProps> = ({ scope, variant, disabled, onFocusChange: o
   const onFormSubmit = (e?: FormEvent) => {
     e?.preventDefault()
     if (value) {
-      navigate(RouteUtils.getSearchRoute(scope, value))
+      let wantedValue = value.trim()
+      let underSpell = false
```
lukaw3d, line 131: I don't love the variable names. Something like `isUnderSpellNext` would better differentiate it from `isUnderSpell`

src/app/components/Search/index.tsx
```diff
@@ -113,7 +127,20 @@ const SearchCmp: FC<SearchProps> = ({ scope, variant, disabled, onFocusChange: o
   const onFormSubmit = (e?: FormEvent) => {
     e?.preventDefault()
     if (value) {
```
lukaw3d, line 129: (This was previously the only validation step, now there's three. It doesn't look easily mergeable into `if (value && !tooShort && !privacyProblem) {`, so I would usually do: `if (!value) return` and ` if (tooShort) return` and `if (privacyProblem) return` )

src/locales/en/translation.json
```diff
@@ -364,6 +364,10 @@
   },
   "search": {
     "placeholder": "Address, Block, Contract, Txn Hash, Transaction ID, Token name, etc",
+    "error": {
+      "tooShort": "Please enter at least 3 characters for searching",
+      "privacy": "This thing that you are trying to search for looks an awful lot like a mnemonic for an Oasis wallet. Please note that this is super-secret data that should never ever be shared with anyone; not even with such excellent services as our {{ appName }}. That being said, we are not here to tell you what you can or can not do with your own data, so if you insist, we WILL search for it. So, if you really think that there is a token with a name that contains this, then in order to signify that you understand and accept the terrible risk of sending this data to our servers, please insert this to the beginning of your search: '{{ wordsOfPower }}'! (Without quotation marks, of course.) If you do so, we will comply with your command."
```
lukaw3d, line 369: I don't think the message matters much; this is fine. Bit awkward for future translators

src/app/components/Search/index.tsx
csillag: OK, cleaned up

src/app/components/Search/index.tsx
csillag: Up, refactored that way.

src/app/components/Search/index.tsx
csillag: Yes but first I want to change the UX according to @donouwens's suggestion about the search suggestions.

csillag:
> Maybe we can create an integration for the second one that works in the same way as the search suggestions under the search bar (https://www.figma.com/file/ifCrok8cP5ymEYjMa2PIi9/Block-Explorer?type=design&node-id=6092%3A227275&mode=design&t=3K5SlcjhCb4mkoJv-1)?

Did that now:

image
image
image

Also, when using the Words of Power:

image
image
image

src/app/components/Search/index.tsx
csillag: See #671

csillag:
Continuing the discussion about the "too short" case in #671

csillag:
> Could this design and text work @csillag?

Sure, but now that we have split this to two, let's finish and merge #671 first, and then I will return to this.

csillag:
> Could this design and text work @csillag?
> 
> https://www.figma.com/file/ifCrok8cP5ymEYjMa2PIi9/Block-Explorer?type=design&node-id=6626%3A247268&mode=design&t=Gk19GONKNbhNZF6X-1

Implemented like this:

image

src/locales/en/translation.json
```diff
@@ -375,7 +375,8 @@
   "search": {
     "placeholder": "Address, Block, Contract, Txn Hash, Transaction ID, Token name, etc",
     "error": {
-      "tooShort": "Please enter either at least 3 characters or a number in order to perform a search."
+      "tooShort": "Please enter either at least 3 characters or a number in order to perform a search.",
+      "privacy": "It seems like you might accidentally entered a keyphrase for a wallet. Please note that your mnemonic is a secret key that should never be shared, even not with our {{ pagetitle }}.\nExecuting this search is highly unlikely to return any results. If you want to proceed nonetheless, please add “{{ wordsOfPower }}” in front of your search query to perform the search at your own risk."
```
lukaw3d, line 379: pagetitle => appName

lukaw3d (approved):

src/locales/en/translation.json
csillag: done

csillag:
See follow up in https://github.com/oasisprotocol/explorer/pull/689

PR #659, opened 2023-07-05
Show different labels for ParaTime snapshot charts

buberdds:
Based on Don's feedback we want to keep current charts behaviour, but we need to show different label for ALL_TIME filter.

### DIFF
--- a/src/app/pages/ParatimeDashboardPage/ActiveAccounts.tsx
+++ b/src/app/pages/ParatimeDashboardPage/ActiveAccounts.tsx
@@ -1,4 +1,5 @@
 import { FC } from 'react'
+import { TFunction } from 'i18next'
 import { useTranslation } from 'react-i18next'
 import startOfMonth from 'date-fns/startOfMonth'
 import { SnapshotCard } from '../../components/Snapshots/SnapshotCard'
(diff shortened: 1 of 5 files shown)


src/app/pages/ParatimeDashboardPage/ActiveAccounts.tsx
```diff
@@ -57,8 +58,16 @@ type ActiveAccountsProps = {
   chartDuration: ChartDuration
 }
 
+const getLabels = (t: TFunction): Record<ChartDuration, string> => ({
+  [ChartDuration.TODAY]: t('chartDuration.lastHour'),
+  [ChartDuration.WEEK]: t('chartDuration.lastDay'),
+  [ChartDuration.MONTH]: t('chartDuration.lastDay'),
+  [ChartDuration.ALL_TIME]: t('chartDuration.lastMonth'),
```
csillag, line 65: This mapping looks weird? (Smells like, off by one?)

src/app/pages/ParatimeDashboardPage/TransactionsChartCard.tsx
```diff
@@ -15,12 +16,20 @@ import { PercentageGain } from '../../components/PercentageGain'
 import startOfHour from 'date-fns/startOfHour'
 import { useRequiredScopeParam } from '../../hooks/useScopeParam'
 
+const getLabels = (t: TFunction): Record<ChartDuration, string> => ({
+  [ChartDuration.TODAY]: t('chartDuration.lastHour'),
+  [ChartDuration.WEEK]: t('chartDuration.lastDay'),
+  [ChartDuration.MONTH]: t('chartDuration.lastDay'),
+  [ChartDuration.ALL_TIME]: t('chartDuration.lastDay'),
+})
+
```
csillag, line 25: Why do we have this mapping repeated by card?

buberdds: We want to have different label for each chart.

csillag:
So, as far as I can tell, two things happened here together:
 - There was a small refactoring, changing where the chart duration labels are calculated
 - There was (probably) some change in the calculation

But since the two changes are conflated in one commit, it's hard for me to figure out what was the actual behavior change....

src/app/pages/ParatimeDashboardPage/ActiveAccounts.tsx
csillag: OK I get it now after the explanation

src/app/pages/ParatimeDashboardPage/TransactionsChartCard.tsx
csillag: Ahh I see, the very last item is different. (I would probably still prefer to have a common function with a parameter, but whatever.)

csillag (approved):
Reviewed. LGTM

PR #661, opened 2023-07-05
Switch to production API

buberdds:


csillag (approved):
Lgtm.

Is the indexing at a stage when we have correct Sapphire data?

buberdds:
It's not up to date, but it looks like staging database was wiped out.

csillag:
Ok merged

PR #669, opened 2023-07-05
Link from unverified contracts to Sourcify to encourage verifying

lukaw3d:
image

### DIFF
--- a/.changelog/669.feature.md
+++ b/.changelog/669.feature.md
@@ -1,0 +1,1 @@
+Link from unverified contracts to Sourcify to encourage verifying
--- a/src/app/components/ContractVerificationIcon/index.tsx
+++ b/src/app/components/ContractVerificationIcon/index.tsx
@@ -85,11 +85,13 @@ export const ContractVerificationIcon: FC<ContractVerificationIconProps> = ({ ac
         {statusIcon[status]}
       </StyledBox>
       &nbsp; &nbsp;
-      {verified && !noLink && (
+      {!noLink && (
         <Typography component="span" sx={{ fontSize: '12px', color: COLORS.brandExtraDark }}>
           <Trans
             t={t}
-            i18nKey="contract.verification.openInSourcify"
+            i18nKey={
+              verified ? 'contract.verification.openInSourcify' : 'contract.verification.verifyInSourcify'
+            }
             components={{
               SourcifyLink: (
                 <Link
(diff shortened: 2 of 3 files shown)


csillag:
I like the general idea, but how can a random visitor start the verification process, when it needs to have the source code and the metadata manually inputted? At sourcify, there _is_ an option called "import from contract", but it doesn't seem to work in the test case. (Also the address of the contract must be entered manually again, which is poor UI.) Would it be possible to provide a closer integrration? At least avoid having to re-enter the address, but ideally also automatically provide the alleged sources? Or is this information not available anywhere?

Also, what happens is the verification fails? What does it mean? Does it say something about the contract itself, or does it just mean that the person who started the verification process provided the wrong data? Do we even get notified about failed verification attempts?

lukaw3d:
Sometimes a random visitor can do the verification, since our chains are [not monitored](https://docs.sourcify.dev/docs/chains/); I verified a few contracts just using "Import from Contract" + input address and chain. Other than that, it's intended for contract owners.

I don't think we can avoid re-entering the address :shrug: 

> but ideally also automatically provide the alleged sources? Or is this information not available anywhere?

not sure what you mean here

> Verification fails

Pretty sure it just means the compilation result has a different hash, so the source code or the compilation parameters differed. And oasis is not notified about such mismatches.

csillag:
> I verified a few contracts just using "Import from Contract" + input address and chain. 

Is there a way for us to determine in advance whether or not this will be possible, just by looking at our own data? Btw, how does Sourcify pull the data? What kind of service to they use? 

I think we should show a different message (or at least tooltip) depending on whether or not this is possible for the given contract. Since if it's not, I think we should probably say something like "Verify this contract (assuming you have the sources)" or something like that...

> I don't think we can avoid re-entering the address shrug

Can we talk to the Sourcify team and ask them to provide a route which accepts the address as part of the path? (And hopefully also the "import from contract" setting...)

lukaw3d:
I don't think we can pre-determine it https://docs.sourcify.dev/docs/monitoring/

I guess the longer label "Verify this contract (assuming you have the sources)" is also a substitute for explaining what verified contracts even mean.

> Address in path

TODO: check https://github.com/ethereum/sourcify

lukaw3d:
Don said
> I don’t think it’s particularly inviting to have that text link, but I guess we could reword it to be less actionable; `Verify through Sourcify` - let’s go with that, please.

csillag (approved):
LGTM.

PR #671, opened 2023-07-05
Show warning when search text is too short

csillag:
image
image
image

### DIFF
--- a/.changelog/671.feature.md
+++ b/.changelog/671.feature.md
@@ -1,0 +1,1 @@
+Show a warning when search text string it too short
--- a/src/app/components/Search/index.tsx
+++ b/src/app/components/Search/index.tsx
@@ -10,11 +10,15 @@ import { COLORS } from '../../../styles/theme/colors'
 import { RouteUtils } from '../../utils/route-utils'
 import { useScreenSize } from '../../hooks/useScreensize'
 import HighlightOffIcon from '@mui/icons-material/HighlightOff'
+import ErrorIcon from '@mui/icons-material/Error'
 import IconButton from '@mui/material/IconButton'
 import { SearchSuggestionsButtons } from './SearchSuggestionsButtons'
 import { formHelperTextClasses } from '@mui/material/FormHelperText'
 import { outlinedInputClasses } from '@mui/material/OutlinedInput'
 import { SearchScope } from '../../../types/searchScope'
+import { textSearchMininumLength } from './search-utils'
+import Typography from '@mui/material/Typography'
+import { isValidBlockHeight } from '../../utils/helpers'
 
 export type SearchVariant = 'button' | 'icon' | 'expandable'
 
@@ -101,9 +105,18 @@ const SearchCmp: FC<SearchProps> = ({ scope, variant, disabled, onFocusChange: o
   const [isFocused, setIsFocused] = useState(false)
   const valueInSearchParams = useSearchParams()[0].get('q') ?? ''
 
+  const valueWithoutPrefix = value.trim()
+
+  const isTooShort =
+    !!value && valueWithoutPrefix.length < textSearchMininumLength && !isValidBlockHeight(valueWithoutPrefix)
+
   useEffect(() => {
     setValue(valueInSearchParams)
   }, [valueInSearchParams])
+
+  const onChange = (newValue: string) => {
+    setValue(newValue)
+  }
 
   const onFocusChange = (value: boolean) => {
     setIsFocused(value)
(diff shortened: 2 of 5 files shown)


csillag:
> I would suggest this for the 'too little characters' notification: https://www.figma.com/file/ifCrok8cP5ymEYjMa2PIi9/Block-Explorer?type=design&node-id=6624%3A246974&mode=design&t=Gk19GONKNbhNZF6X-1

OOps, I didn't see that.

csillag:
I have updated the screenshort after making it more similar to the visuals.

csillag:
> Could we outline the input field with a red stroke as well, please?

Like this? 
image

csillag:
> the stroke feels thicker than in the visuals. I assume it's the 3px active stroke, but maybe we could try the inactive 1px if it's not too much work?

I just triggered the built-in "error" flag on the input field, anything beyond that was automatic. Will have to dig deeper... 🧤

lukaw3d:
I don't like that it is so hard to get to search suggestions

csillag:
> I don't like that it is so hard to get to search suggestions

Please elaborate

lukaw3d:
If user doesn't know what to search for, they shouldn't need to make up 3 chars

lukaw3d:
also, hold on, why are we not allowing searching for block "10" :/

csillag:
> also, hold on, why are we not allowing searching for block "10" :/

That we can fix easily.

image
image

csillag:
> If user doesn't know what to search for, they shouldn't need to make up 3 chars

Fair point. Here is some potential solutions, and the drawbacks I see

|**Option** | **Problem** |
|---|---|
|**0.** **(Current behavior)** for short search string, the default suggestions are obscured. | If user doesn't know what to search for, they shouldn't need to make up 3 chars just to get the search suggestion. | 
| **1.** Leave the default suggestions in place, don't show the "too short" message, just disable the search   | It won't be obvious why the search button is disabled.  |
| **2.** Leave the default suggestions in place, don't show the "too short" message, enable the search and execute it normally.   | The users might never figure out why they can't find their "XYZ" token when they search for "XY".  |
| **3.** Leave the default suggestions in place, don't show the "too short" message, enable the search, but instead of executing it, show a separate error message (at a different location, not obscuring the search suggestion) upon searching. **(This was my previous implementation)**   | Visually might not be that pleasant (?) |
| **4.** Introduce a second line of search suggestions. The first line would always be the default suggestions, and the second line would be the "too short" message, when applicable. | This is tricky because currently the "too short" message is themed as an error, while the other one is not. So we either have to down-grade it from an error to a normal message (no ping bg), or come up with a design of the two lines being in different colors, which might be confusing.|

@donouwens @lukaw3d please state your preferences.

csillag:
I have now implemented option 4. It can be tested [here](https://f9b897a5.oasis-explorer.pages.dev/).

csillag:
The table with options has been updated, see online

csillag:
>  the stroke feels thicker than in the visuals. I assume it's the 3px active stroke, but maybe we could try the inactive 1px. [...]  Ideally I'd say the error would be less thick, but I understand if that's not within the scope of this task.

OK @donouwens , but what do you mean exactly?

A) Shall I reduce the 3px active stroke at the outline of input fields to 1px everywhere?
B) Shall I reduce the 3px active stroke at the outline of input fields _in case of errors_ to 1px everywhere?
C) Shall I reduce the 3px active stroke at the outline, in case of errors, just this once, at this field?

csillag:
> > the stroke feels thicker than in the visuals. I assume it's the 3px active stroke, but maybe we could try the inactive 1px. [...]  Ideally I'd say the error would be less thick, but I understand if that's not within the scope of this task.
> 
> OK @donouwens , but what do you mean exactly?
> 
> A) Shall I reduce the 3px active stroke at the outline of input fields to 1px everywhere? 
> B) Shall I reduce the 3px active stroke at the outline of input fields _in case of errors_ to 1px everywhere?
> C) Shall I reduce the 3px active stroke at the outline, in case of errors, just this once, at this field?

Until further notice, I have done option B)

lukaw3d:
Option 5: show suggestions on focus, show "min 3 chars" when anything is entered
(that would require quite some refactoring of search using ClickAwayListener, so separate PR?)

lukaw3d (approved):

csillag:
Sorry @donouwens we want to move on, so we will just merge this now. We can always update the styling if required.

csillag:
>  visually it needs updating - could we show it like this:

Sorry could you please help me by detecting and enumerating the points of difference?

| Design | Implementation |
| ---- | --- |
| image | image |

csillag:
See follow up in https://github.com/oasisprotocol/explorer/pull/689

PR #651, opened 2023-07-03
Improve displaying events

lukaw3d:
| Before | After |
| --- | --- |
| [link with all event types](https://50733e59.oasis-explorer.pages.dev/mainnet/emerald/tx/0x481380a29028d51ee19a06ef0cae56d5719b0f9465e2adba8a26fce9961f2f4c) | [link with all event types](https://9acdaf47.oasis-explorer.pages.dev/mainnet/emerald/tx/0x481380a29028d51ee19a06ef0cae56d5719b0f9465e2adba8a26fce9961f2f4c) |
| 50733e59 oasis-explorer pages dev_mainnet_emerald_tx_0x481380a29028d51ee19a06ef0cae56d5719b0f9465e2adba8a26fce9961f2f4c | 9acdaf47 oasis-explorer pages dev_mainnet_emerald_tx_0x481380a29028d51ee19a06ef0cae56d5719b0f9465e2adba8a26fce9961f2f4c |

PR #672, opened 2023-07-05
Remove "Decoded" column from events to improve mobile layout

lukaw3d:
Related to https://github.com/oasisprotocol/explorer/pull/651#issuecomment-1620638881
| Before | After |
| --- | --- |
| localhost_1234_mainnet_emerald_tx_5277bcc0db638ec88bced8be20fb7828aed2b97fa9893470fa5ee124e3344cca (1) | localhost_1234_mainnet_emerald_tx_5277bcc0db638ec88bced8be20fb7828aed2b97fa9893470fa5ee124e3344cca |

### DIFF
--- a/src/app/components/Transactions/LogEvent.tsx
+++ b/src/app/components/Transactions/LogEvent.tsx
@@ -175,8 +175,11 @@ const LogEvent: FC<{
               <AccountLink
                 address={event.body.owner}
                 scope={scope}
-                plain={addressSwitchOption === AddressSwitchOption.ETH}
+                plain={addressSwitchOption !== AddressSwitchOption.Oasis}
               />
+              {addressSwitchOption === AddressSwitchOption.Oasis && (
+                <CopyToClipboard value={event.body.owner} />
+              )}
             </dd>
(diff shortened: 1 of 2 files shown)


src/app/components/Transactions/LogEvent.tsx
```diff
@@ -179,8 +175,11 @@ const DecodedLogEvent: FC<{
               <AccountLink
                 address={event.body.owner}
                 scope={scope}
-                plain={addressSwitchOption === AddressSwitchOption.ETH}
+                plain={addressSwitchOption !== AddressSwitchOption.Oasis}
```
csillag, line 178: What's the point of this change?

csillag (approved):
Reviewed. LGTM.

src/app/components/Transactions/LogEvent.tsx
lukaw3d: Just more accurate: "We have oasis address, so `plain` when addressSwitchOption isn't what we have"

instead of "We have oasis address, so `plain` when addressSwitchOption is ETH"

PR #675, opened 2023-07-06
Skip hour/day in order to avoid charts drop

buberdds:


buberdds:
There is one more issue I need to address for daily chart

PR #676, opened 2023-07-06
Reposition help screen button on mobile

lubej:


lubej:
> fyi: not related to this change, but we lost footer on laptop/mobile (per design?) maybe we should add some padding bottom.

Right per design. Will add a bit of padding 👍

PR #680, opened 2023-07-06
Clean up theme.spacing usage and React imports

buberdds:


src/styles/theme/defaultTheme.ts
```diff
@@ -499,7 +499,7 @@ export const defaultTheme = createTheme({
       styleOverrides: {
         paperAnchorTop: ({ theme }) => ({
           borderRadius: '0 0 12px 12px',
-          padding: `${theme.spacing(4)} 5%`,
+          padding: theme.spacing(4, '5%'),
```
lukaw3d, line 502: it doesn't have negative numbers, but has percentages? :O

lukaw3d (approved):

src/styles/theme/defaultTheme.ts
buberdds: yeah it looks like it's not supported

PR #685, opened 2023-07-06
Display type in tokens table

csillag:
Now that we have two types of tokens, ERC-20 and ERC-721, it helps to show which is which in the tables.

|Before | After|
|---|---|
|Token list horizontal | l|
| image  |image |
|Token list vertical | |
| image |  image|
|Token  dashboard | |
| image |  image  |
| Account details | |
| image | image|

### DIFF
--- a/src/app/components/Tokens/TokenList.tsx
+++ b/src/app/components/Tokens/TokenList.tsx
@@ -7,6 +7,7 @@ import { TokenLink } from './TokenLink'
 import { CopyToClipboard } from '../CopyToClipboard'
 import { DelayedContractVerificationIcon } from '../ContractVerificationIcon'
 import Box from '@mui/material/Box'
+import { getTokenTypeName } from '../../pages/TokenDashboardPage/TokenTypeCard'
 
 type TokensProps = {
   tokens?: EvmToken[]
@@ -21,6 +22,7 @@ export const TokenList = (props: TokensProps) => {
   const tableColumns: TableColProps[] = [
     { key: 'index', content: '' },
     { key: 'name', content: t('common.name') },
+    { key: 'type', content: t('common.type') },
     { key: 'contract', content: t('common.smartContract') },
     { key: 'verification', content: t('contract.verification.title') },
     {
@@ -53,6 +55,10 @@ export const TokenList = (props: TokensProps) => {
             />
           ),
           key: 'name',
+        },
+        {
+          key: 'type',
+          content: getTokenTypeName(t, token.type),
         },
         {
           content: (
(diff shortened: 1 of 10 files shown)


csillag:
@donouwens what do you think?

.changelog/681.feature.md
```diff
@@ -1 +1 @@
-Fully bring back support for ERC-721 tokens
```
lukaw3d, line 1: intentional change to old fragment?

csillag: No, mistyped number. Fixed.

buberdds (approved):

csillag:
@donouwens To be honest, I am not sure about or nomenclature here. Looking at Etherscan, they simply call the [ERC-20 stuff](https://etherscan.io/tokens) "Tokens" and the [ERC-721 and similar stuff](https://etherscan.io/nft-top-contracts) "NFTs". (But then for the second category, they do show "type" in the table. Maybe what they are doing is not a bad idea?

But then this brings up the question, do we even _want_ to have these two categories (tokens and NFTs) in the same table? If yes, let's just say "Tokens" and "NFTs" as the type? Or have two different tables? Does this also mean new routes and new pages?

So many possibilities.

In any case, I'm ok with colored tags, but I also would also consider adding "NFT" to the tags like this: `NFT (ERC-721)`

lukaw3d:
My suggestion would be: same table for both, no type column, name = "Wrapped Rose" | "AI ROSE (NFT)". But then vertical view becomes "AI ROSE (AI ROSE) (NFT)"

csillag:
> My suggestion would be: same table for both, no type column, name = "Wrapped Rose" | "AI ROSE (NFT)". But then vertical view becomes "AI ROSE (AI ROSE) (NFT)"

Could the (NFT) added to the name be a tag like this?

image

And can we put the ERC-721 into the tooltip?

csillag:
Tweaked the look according to the visuals Updated screenshots above.

csillag:
Updated styling. I replaced the first 2 screenshot, but not the others.

src/app/components/Tokens/TokenList.tsx
```diff
@@ -15,12 +25,40 @@ type TokensProps = {
   pagination: false | TablePaginationProps
 }
 
+export const TokenTypeTag: FC<{ tokenType: EvmTokenType; sx?: SxProps }> = ({ tokenType, sx = {} }) => {
```
lukaw3d, line 28: This is similar to `TokenPills` that wraps around mui's `Chip`.

We should choose more consistent names for "the background rounded label thing"

csillag: How about `DogTag` ?

csillag: (I mean for a common ancestor, with common sizing and fond settings, and then create or descendants from that, with consistent names, either Tag or Pill)

PR #687, opened 2023-07-06
Properly display balance on ERC-721 token transfers

csillag:


src/app/components/Account/ContractCreatorInfo.tsx
```diff
@@ -60,7 +60,7 @@ export const ContractCreatorInfo: FC<{
       &nbsp;
       <Trans
         t={t}
-        i18nKey={'contract.createdAt'}
+        i18nKey="contract.createdAt"
```
csillag, line 63: Small independent fix sneaked in

src/locales/en/translation.json
```diff
@@ -62,7 +62,7 @@
     "height": "Height",
     "hide": "Hide",
     "loadMore": "Load more",
-    "lessThanAmount": "< {{value}} {{ticker}}",
+    "lessThanAmount": "< {{value}} <TickerLink>",
```
lukaw3d, line 65: <TickerLink />

src/app/components/Tokens/TokenTransfers.tsx
```diff
@@ -82,6 +129,7 @@ type TokenTransfersProps = {
   isLoading: boolean
   limit: number
   pagination: false | TablePaginationProps
+  tickersAsLink?: boolean | undefined
```
buberdds, line 132: `tickersAsLink:`

src/app/components/Tokens/TokenTransfers.tsx
```diff
@@ -63,17 +65,62 @@ const DelayedEventBalance: FC<
...
     },
-  })
+  )
+  const token = data?.data
+
+  if (isLoading) {
+    return <Skeleton variant="text" />
+  }
+  if (isError || !token) {
+    return t('common.missing')
```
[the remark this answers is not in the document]
lukaw3d: Hm, in case of a different type of transfer we dont know where amount field is, so maybe throw in devmode and "n/a" in production

src/locales/en/translation.json
csillag: Right. Fixed.

src/app/components/Tokens/TokenTransfers.tsx
csillag: I would prefer not doing that, because I want to able to use this component without specifying this optional prop. Removing the `?` make that impossible.

src/app/components/Tokens/TokenTransfers.tsx
```diff
@@ -63,17 +65,62 @@ const DelayedEventBalance: FC<
...
+
+  const ticker = token.symbol || t('common.missing')
 
-  const token = tokenQuery.data?.data
-  const realValue = value === undefined ? undefined : fromBaseUnits(value, token?.decimals || 0)
-  const ticker = token?.symbol || t('common.missing')
+  if (token.type === EvmTokenType.ERC20) {
+    const token = data?.data
+    const value = event.evm_log_params?.find(param => param.name === 'value')?.value as string | undefined
+    const realValue = value === undefined ? undefined : fromBaseUnits(value, token?.decimals || 0)
```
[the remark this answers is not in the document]
csillag: Renamed this pair from (value, realValue) -> (rawValue, value), plus added comment.

src/app/components/Tokens/TokenTransfers.tsx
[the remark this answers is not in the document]
csillag: OK, throwing in dev mode, "n/a" in production.

src/app/components/RoundedBalance/__tests__/index.test.tsx
```diff
@@ -1,41 +1,66 @@
...
+  (pattern: string): Matcher =>
+  (content, element) =>
+    !!element && extractTextFromElement(element).startsWith(content) && hasOnlyOneLevelOfChildren(element)
+
 describe('RoundedBalance', () => {
   it('should keep original value', () => {
     render(<RoundedBalance value="0.00075" ticker="USDC" />)
 
-    expect(screen.getByText('0.00075 USDC')).toBeInTheDocument()
+    expect(screen.getByText(getFlexMatcherFor('0.00075 USDC'))).toBeInTheDocument()
```
lukaw3d, line 33: this doesn't work.

src/app/components/Tokens/TokenTransfers.tsx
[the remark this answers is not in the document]
lukaw3d: Can you add snapshot tests for all these edge cases

src/app/pages/AccountDetailsPage/AccountTokenTransfersCard.tsx
```diff
@@ -44,6 +44,7 @@ export const AccountTokenTransfersCard: FC = () => {
               isTotalCountClipped,
               rowsPerPage: NUMBER_OF_ITEMS_ON_SEPARATE_PAGE,
             }}
+            differentTokens
```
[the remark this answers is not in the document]
lukaw3d: Oh, it's also about tickerAsLink :/

src/app/utils/placeholderLabel.tsx
```diff
@@ -0,0 +1,10 @@
+import { FC } from 'react'
```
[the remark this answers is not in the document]
csillag: Right. Fixed.

src/app/pages/AccountDetailsPage/AccountTokenTransfersCard.tsx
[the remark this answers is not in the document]
csillag: > I would rename to `hasDifferentTokensTypes` or `differentiateTokenTypes`
> [...]
> Oh, it's also about tickerAsLink :/

Yes it's dual purpose. The intended semantic is that this means that in this table we are going to have data related to potentially various multiple tokens, so we need more data related to the tokens. (Token type, link to dashboard)

src/app/components/Tokens/hooks.ts
```diff
@@ -0,0 +1,32 @@
...
+    throw AppErrors.UnsupportedLayer
+  }
+  useEffect(() => {
+    if (ethAddress) {
+      getEvmBech32Address(ethAddress).then(setOasisAddress)
+    }
+  }, [ethAddress])
+  const { isLoading, isError, data } = useGetRuntimeEvmTokensAddress(network, layer, oasisAddress, {
+    query: {
+      enabled: !!oasisAddress,
```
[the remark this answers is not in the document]
csillag: Yes! According to [the docs](https://tanstack.com/query/latest/docs/react/reference/useQuery), it's true when:

> if there's no cached data and no query attempt was finished yet.

I find it kinda misleading, compared to the natural interpretation of the English naming. That is why I do [this](https://github.com/oasisprotocol/explorer/blob/3a7f7fdddb65f41ceadca0aa4e465dc67a7314e0/src/app/pages/TokenDashboardPage/hook.ts#L25) sometimes.

In the current case, the only case when it would return a false positive is when there is no valid base64 address input, which I have now fixed.

src/app/components/RoundedBalance/__tests__/index.test.tsx
csillag: OK, text matching fixed. Fixing the tests revealed a problem, which is now also fixed.

lukaw3d: Works. But I think this is how it natively works too. Is this needed?

csillag: When you have two text elements side by side, like this: 
image

... and you search for the concatenated text, no match is found. There is an error message that explicitly explains that this might be because the text is divided into multiple elements, and in this case you should provide a more flexible matcher function. 

I think they could just as well provide it themselves, but apparently they don't, or at lest I didn't find it. The default matcher was definitely inadequate.

lukaw3d: If I remove getFlexMatcherFor tests still pass

csillag: OK I have no idea what happened here, then. For they definitely didn't pass before. Maybe it was an independent problem which has been resolved since, and the crafty error message mislead me about the reason? I don't know.

In any case I have no removed the unnecessary changes.

lukaw3d (approved):

src/app/components/Tokens/TokenTransfers.tsx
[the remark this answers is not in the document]
lukaw3d: Resurrecting with an example:
https://explorer.dev.oasis.io/mainnet/emerald/address/0xceF2f95f185D49bcd1c10DE7f23BEaCBaae6eD0f/token-transfers#transfers

image

instead of n/a we could show the raw transfer amounts
```js
"value": "1004350000000000000"
"value": "24224120000000000000"
"value": "895423197350738460735"
"value": "107180000000000000"
```

PR #689, opened 2023-07-07
Tweak the search field based on design input

csillag:
- [x] Update the placeholder
- [x] Add missing clear button
- [x] Add time delay between typing and showing the error message
- [x] ~Add animation when showing/hiding search suggestions~
- [x] ~Add animation when showing/hiding the error message~
- [x] Different colors and icons for warning and error messages
- [x] Update borders
- [x] Other updates?
- [x] ~Move some if the customizations to a higher level, into a dedicated (styled) component or into the theme itself~

### DIFF
--- a/src/app/components/Search/index.tsx
+++ b/src/app/components/Search/index.tsx
@@ -20,6 +20,7 @@ import { SearchScope } from '../../../types/searchScope'
 import { textSearchMininumLength } from './search-utils'
 import Typography from '@mui/material/Typography'
 import { isValidBlockHeight } from '../../utils/helpers'
+import { typingDelay } from '../../../styles/theme'
 import { isValidMnemonic } from '../../utils/helpers'
 
 export type SearchVariant = 'button' | 'icon' | 'expandable'
@@ -106,6 +107,7 @@ const SearchCmp: FC<SearchProps> = ({ scope, variant, disabled, onFocusChange: o
   const [value, setValue] = useState('')
   const [isFocused, setIsFocused] = useState(false)
   const valueInSearchParams = useSearchParams()[0].get('q') ?? ''
+  const [isTyping, setIsTyping] = useState(false)
 
   const wordsOfPower = t('search.wordsOfPower')
   const hasWordsOfPower = value.trim().toLowerCase().startsWith(wordsOfPower.toLowerCase())
@@ -125,6 +127,7 @@ const SearchCmp: FC<SearchProps> = ({ scope, variant, disabled, onFocusChange: o
 
   const onChange = (newValue: string) => {
     setValue(newValue)
+    setIsTyping(true)
   }
 
   const onFocusChange = (value: boolean) => {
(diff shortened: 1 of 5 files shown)


src/app/components/Search/index.tsx
```diff
@@ -154,13 +156,26 @@ const SearchCmp: FC<SearchProps> = ({ scope, variant, disabled, onFocusChange: o
...
+  const hasWarning = !!warningMessage
 
-  const hasError = !!errorMessage
+  const hasProblem = hasError || hasWarning
+
+  useEffect(() => {
+    if (hasProblem) {
+      const timeout = setTimeout(() => {
+        setItsTimeToAct(true)
+      }, 500)
```
lukaw3d, line 173: (I dont like this pattern much. I'd validate either instantly, or on submit)

csillag: Actually we validate instantly in the sense that the search button is disabled right away; we just don't want to shove the error message to the face of the user. (Hopefully he will keep typing and won't even have to see it at all.)

src/app/components/Search/index.tsx
```diff
@@ -105,6 +107,7 @@ const SearchCmp: FC<SearchProps> = ({ scope, variant, disabled, onFocusChange: o
   const [value, setValue] = useState('')
   const [isFocused, setIsFocused] = useState(false)
   const valueInSearchParams = useSearchParams()[0].get('q') ?? ''
+  const [itsTimeToAct, setItsTimeToAct] = useState(false)
```
lukaw3d, line 110: I dont think this variable name is clear

src/app/components/Search/index.tsx
```diff
@@ -154,13 +157,26 @@ const SearchCmp: FC<SearchProps> = ({ scope, variant, disabled, onFocusChange: o
...
+  const hasWarning = !!warningMessage
 
-  const hasError = !!errorMessage
+  const hasProblem = hasError || hasWarning
+
+  useEffect(() => {
+    if (hasProblem) {
+      const timeout = setTimeout(() => {
+        setItsTimeToAct(true)
+      }, tooltipDelay * 2)
```
lukaw3d, line 174: Is there a relation to tooltip?

src/app/components/Search/index.tsx
lukaw3d: Perhaps a better structure is: a delayed "has user stopped typing", instead of delayed "has a problem"

lukaw3d (approved):

src/app/components/Search/index.tsx
csillag: No. I have added a separate time constant now.

src/app/components/Search/index.tsx
csillag: OK, as suggested, now we have `isTyping`, which is always set to true when typing, and then set to false in a delayed manner. (And we display messages when !isTyping.)

PR #690, opened 2023-07-07
Introduce animations for the search input

csillag:
This makes it so that the help text (under the search input text field) collapses in animated way.

### DIFF
--- a/src/app/components/Search/index.tsx
+++ b/src/app/components/Search/index.tsx
@@ -4,6 +4,7 @@ import TextField, { textFieldClasses } from '@mui/material/TextField'
 import InputAdornment, { inputAdornmentClasses } from '@mui/material/InputAdornment'
 import { styled } from '@mui/material/styles'
 import Button from '@mui/material/Button'
+import Box from '@mui/material/Box'
 import SearchIcon from '@mui/icons-material/Search'
 import { useTranslation } from 'react-i18next'
 import { COLORS } from '../../../styles/theme/colors'
@@ -176,8 +177,6 @@ const SearchCmp: FC<SearchProps> = ({ scope, variant, disabled, onFocusChange: o
     return () => clearTimeout(timeout)
   }, [value])
 
-  const needsHelp = !!value && value !== valueInSearchParams
-
   return (
     <SearchForm
       searchVariant={variant}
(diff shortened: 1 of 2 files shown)


src/app/components/Search/index.tsx
```diff
@@ -228,11 +231,10 @@ const SearchCmp: FC<SearchProps> = ({ scope, variant, disabled, onFocusChange: o
           },
         }}
         helperText={
-          value &&
-          value !== valueInSearchParams && (
-            <>
-              {!isTyping && hasError && (
-                <>
+          needsHelp && (
+            <Collapse in={needsHelp}>
```
csillag, line 235: There is a reason why we are using _both_ as Collapse _and_ a condition. The Collapse only closes half-way, so to make it properly disappear, we also need the condition. Yes, it won't be visible all the time, but it still kinda works.

buberdds: Doesn't make sense to use Collapse here as animation will never work.

buberdds:
```
        FormHelperTextProps={{
        // turn off TS temporary for next line 
        // issue https://github.com/mui/material-ui/issues/33339
        // fix https://github.com/mui/material-ui/pull/35924
          component: 'div', // replace p with div tag
          sx: {
            marginTop: 0, // set to zero  so we can remove `needsHelp &&`
            marginBottom: 0, // set to zero  so we can remove `needsHelp &&`
            marginLeft: variant === 'button' ? '48px' : '17px',
            marginRight: variant === 'button' ? '48px' : '17px',
          },
        }}
```

removed margin needs to be moved under the first collapse

```
        helperText={
          <Collapse in={needsHelp}>
            <Box sx={{ py: '10px' }}>
              <Collapse in={!isTyping && hasError}>
              ....
```

csillag: That all works perfectly, thanks!

src/app/components/Search/index.tsx
```diff
@@ -220,67 +225,66 @@ const SearchCmp: FC<SearchProps> = ({ scope, variant, disabled, onFocusChange: o
         placeholder={searchPlaceholderTranslated}
         fullWidth
         FormHelperTextProps={{
+          // turn off TS temporary for next line
+          // issue https://github.com/mui/material-ui/issues/33339
+          // fix https://github.com/mui/material-ui/pull/35924
+          // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+          // @ts-ignore
```
buberdds, line 232: ts-expect-error does not need lint override

and from docs
```
Pick ts-expect-error if:
...
- you expect a fix to be coming in fairly quickly and you just need a quick workaround
```

csillag: Indeed. Changed accordingly.

src/app/components/Search/index.tsx
```diff
@@ -169,11 +170,15 @@ const SearchCmp: FC<SearchProps> = ({ scope, variant, disabled, onFocusChange: o
...
-    const timeout = setTimeout(() => {
-      setIsTyping(false)
-    }, typingDelay)
-    return () => clearTimeout(timeout)
-  }, [value])
+    if (hasProblem) {
+      const timeout = setTimeout(() => {
+        setIsProblemFresh(false)
+      }, typingDelay)
+      return () => clearTimeout(timeout)
```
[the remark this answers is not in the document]
csillag: No, we only need to clear it if we have set it up. (And we only set it up inside the if block.) Outside the IF, there is no `timeout variable available...

buberdds (approved):

PR #691, opened 2023-07-07
Improve token dashboard

csillag:
This PR improves two small things around the token dashboard.

###  Make sure there are loading states for all the cards on the dashboard

Now, while loading the token info:

image

### Handle undefined token total supply

When a token doesn't have the token supply defined, say "undefined" instead of "n/a", as suggested by @donouwens 

| before | after |
|----- | ----|
|  image  |  image |

### DIFF
--- a/src/app/components/Tokens/TokenTransfers.tsx
+++ b/src/app/components/Tokens/TokenTransfers.tsx
@@ -210,6 +210,7 @@ export const TokenTransfers: FC<TokenTransfersProps> = ({
       isLoading={isLoading}
       pagination={pagination}
       extraHorizontalSpaceOnMobile
+      alwaysWaitWhileLoading
     />
   )
 }
--- a/.changelog/691.feature.md
+++ b/.changelog/691.feature.md
@@ -1,0 +1,1 @@
+Token dashboard: make sure all cards have loading state
--- a/src/app/components/Table/index.tsx
+++ b/src/app/components/Table/index.tsx
@@ -74,7 +74,8 @@ type TableProps = {
   rows?: TableRowProps[]
   rowsNumber?: number
   stickyColumn?: boolean
-  extraHorizontalSpaceOnMobile?: boolean
+  extraHorizontalSpaceOnMobile?: boolean | undefined
+  alwaysWaitWhileLoading?: boolean | undefined
 }
 
 const stickyColumnStyles = {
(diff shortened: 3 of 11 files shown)


src/app/components/Tokens/TokenTransfers.tsx
```diff
@@ -204,7 +204,7 @@ export const TokenTransfers: FC<TokenTransfersProps> = ({
   return (
     <Table
       columns={tableColumns}
-      rows={tableRows}
+      rows={isLoading ? undefined : tableRows}
```
lukaw3d, line 207: How come this is needed? we already pass `isLoading={isLoading}`

lukaw3d: Looks like wrong level of responsibility

lukaw3d (approved):

src/app/components/Tokens/TokenTransfers.tsx
csillag: > How come this is needed? we already pass `isLoading={isLoading}`

Yes, but the [current implementation](https://github.com/oasisprotocol/explorer/blob/4fa627013e05c55f9a96036fba742d00332d2eed/src/app/components/Table/index.tsx#L128) of `Table` only considers `isLoadin` if there are no rows:

``` TypeScript
{!rows && isLoading && (
  <SkeletonTableRows rowsNumber={rowsNumber} columnsNumber={columns.length} />
)}
```

csillag: > Looks like wrong level of responsibility

Fair point. The problem is that if we change the behavior of Table (to inverse the priorities, and make `isLoading` trump the passed rows), that might change the behavior of other pages in unpredictable ways.

However we could make this a configurable, opt-in behavior of Table, and only enable it for this page (for now), and then the code would be at the correct level, and we wouldn't break other pages.

I think I will do this now.

csillag: Done.

lukaw3d: We should untangle that. Does any table ever rely on `!rows`? Origin seems to be https://github.com/oasisprotocol/explorer/pull/15

csillag: I think the idea seems to be that if we already have content, and are currently reloading, there is no reason to hide the existing data; that would just cause flickering. Instead, keep the old data while loading, and then update when we have the new data. It's a valid use-case, and I think we are actually using it for updating tables, like the latest transactions and latest blocks. (And then we also have the code to highlight new rows.)

So yes, I think we _are_ using that.

PR #695, opened 2023-07-09
Use eth addresses in account token balances

csillag:
Since the required ETH address data is currently not present in the balances provided by the API, this requires pulling up the contracts for each balance.

Before:

image

After:

image

### DIFF
--- a/.changelog/696.trivial.md
+++ b/.changelog/696.trivial.md
@@ -1,0 +1,1 @@
+Eliminate some useless token loading requests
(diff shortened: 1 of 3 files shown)


.changelog/696.trivial.md
```diff
@@ -0,0 +1 @@
+Eliminate some useless token loading requests
```
csillag, line 1: Adding back the old changelong entry, this time with the correct filename.

lukaw3d (approved):

PR #696, opened 2023-07-10
Eliminate some useless token loading requests

csillag:
When we are looking at account details, and the account is a token contract, then we also need to load the token data to get the name of the token, which we want to display.

Because of this, whenever looking at account details, we were always requesting token data at the same time.

This commit restricts sending this request to cases when we already know that we are looking at contract.

### DIFF
--- a/src/app/pages/TokenDashboardPage/hook.ts
+++ b/src/app/pages/TokenDashboardPage/hook.ts
@@ -9,16 +9,23 @@ import { SearchScope } from '../../../types/searchScope'
 import { useSearchParamsPagination } from '../../components/Table/useSearchParamsPagination'
 import { NUMBER_OF_ITEMS_ON_SEPARATE_PAGE } from '../../config'
 
-export const useTokenInfo = (scope: SearchScope, address: string) => {
+export const useTokenInfo = (scope: SearchScope, address: string, enabled = true) => {
   const { network, layer } = scope
   if (layer === Layer.consensus) {
     // There can be no ERC-20 or ERC-721 tokens on consensus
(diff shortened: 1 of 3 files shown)


src/app/pages/TokenDashboardPage/hook.ts
```diff
@@ -9,16 +9,18 @@ import { SearchScope } from '../../../types/searchScope'
...
     throw AppErrors.UnsupportedLayer
   }
-  const query = useGetRuntimeEvmTokensAddress(network, layer, address!)
+  const query = useGetRuntimeEvmTokensAddress(network, layer, address!, {
+    query: { enabled },
+  })
   const token = query.data?.data
   const { isLoading, isError, isFetched } = query
-  return { token, isLoading, isError, isFetched }
+  return { token, isLoading: isLoading && enabled, isError, isFetched }
```
buberdds, line 23: This is weird that we are to change what react-query returns (and it's weird that it returns true :D ). Maybe there are other props like `isInitialLoading` that we need here?

buberdds (approved):

src/app/pages/TokenDashboardPage/hook.ts
csillag: I looked at the docs, and no, they don't seem to have a trivial switch for this behavior. We _could_ insert some placeholder that that would be returned until the query actually runs, and then isLoading would be false, but I don't think it would make our core any more easier to read. (Ie. injecting fake data at one locations, and recognizing fake data at another.)

I will just add some code comments explaining what's going on.

PR #697, opened 2023-07-10
Reposition Token Overview on Dashboard

csillag:
Move Token Overview above Learning Materials on mobile dashboard page.

Task is here: https://app.clickup.com/t/861n1n25y

### DIFF
--- a/src/app/pages/ParatimeDashboardPage/index.tsx
+++ b/src/app/pages/ParatimeDashboardPage/index.tsx
@@ -27,8 +27,13 @@ export const ParatimeDashboardPage: FC = () => {
         <Grid item xs={12} md={6}>
           <LatestBlocks />
         </Grid>
+        {isMobile && (
+          <Grid item xs={12}>
+            <TopTokens />
+          </Grid>
+        )}
       </Grid>
-      <TopTokens />
+      {!isMobile && <TopTokens />}
       <TransactionsStats />
       <TotalTransactions />
       <Social />
--- a/.changelog/697.trivial.md
+++ b/.changelog/697.trivial.md
@@ -1,0 +1,1 @@
+Reposition Token Overview on Dashboard


buberdds (approved):

src/app/pages/ParatimeDashboardPage/index.tsx
```diff
@@ -21,14 +21,19 @@ export const ParatimeDashboardPage: FC = () => {
       <Divider variant="layout" sx={{ mt: isMobile ? 4 : 0 }} />
       <LatestTransactions />
       <Grid container spacing={4}>
-        <Grid item xs={12} md={6} sx={{ display: 'flex', order: isMobile ? 1 : 0 }}>
+        <Grid item xs={12} md={6} sx={{ display: 'flex', order: isMobile ? 2 : 0 }}>
```
buberdds, line 24: This change is not needed, right?

csillag: I'm not sure. I assumed that now (on mobile) we want it at the 3rd spot, after Blocks and Tokens. Isn't that the case?

src/app/pages/ParatimeDashboardPage/index.tsx
```diff
@@ -21,14 +21,19 @@ export const ParatimeDashboardPage: FC = () => {
...
           <LearningMaterials />
         </Grid>
         <Grid item xs={12} md={6}>
           <LatestBlocks />
         </Grid>
+        {isMobile && (
+          <Grid item xs={12} md={6}>
```
buberdds, line 31: we don't need md prop here

csillag: Indeed. Removed.

src/app/pages/ParatimeDashboardPage/index.tsx
csillag: Right, undone.

PR #703, opened 2023-07-10
Round token supply number in token list

buberdds:
Token supply has too many decimal points

### DIFF
--- a/src/app/components/Tokens/TokenList.tsx
+++ b/src/app/components/Tokens/TokenList.tsx
@@ -17,6 +17,7 @@ import { FC } from 'react'
 import Typography from '@mui/material/Typography'
 import { COLORS } from '../../../styles/theme/colors'
 import { SxProps } from '@mui/material/styles'
+import { RoundedBalance } from '../RoundedBalance'
 
 type TokensProps = {
   tokens?: EvmToken[]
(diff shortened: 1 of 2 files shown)


.changelog/702.trivial.md
```diff
@@ -0,0 +1 @@
+Round token supply number in token list
```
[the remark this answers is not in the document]
csillag: I see it's solved now.

src/app/components/Tokens/TokenList.tsx
```diff
@@ -129,7 +130,7 @@ export const TokenList = (props: TokensProps) => {
           align: TableCellAlign.Right,
         },
         {
-          content: token.total_supply,
+          content: <RoundedBalance value={token.total_supply} ticker={token.symbol} />,
```
csillag, line 133: Actually I prefer this way of formatting the supply: https://github.com/oasisprotocol/explorer/blob/master/src/locales/en/translation.json#L215

With rounded balance:
image


Via i18next: 
image

csillag: But I also understand that it is not great with decimals.
I think we need a way that satisfies both criteria.

Can we have the same grouping in RoundedBalance?

(Also, maybe we could consider to resurrect #333?)

buberdds: Ok, I think we are good here. As discussed during meeting, RoundedBalance logic will be adjusted in another PR.

PR #715, opened 2023-07-11
Format RoundedBalance using thousands separators and fix sparse values

lukaw3d:


src/app/components/RoundedBalance/__tests__/index.test.tsx
```diff
@@ -4,45 +4,69 @@ import { RoundedBalance } from './../'
...
+
+    it('with decimals', () => {
+      render(
+        <RoundedBalance
+          value="1111111111111111111111111111111111111111111111111111.00000000000002231"
+          ticker="ROSE"
+        />,
+      )
+      expect(
+        screen.getByText('1,111,111,111,111,111,111,111,111,111,111,111,111,111,111,111,111,111.00000… ROSE'),
```
buberdds, line 61: This is improvement, but (using more readable numbers) shouldn't we have something like: 
```
    it('whole number', () => {
      render(<RoundedBalance value="15000000.00000000000002231" ticker="ROSE" />)
      expect(screen.getByText('15M ROSE')).toBeInTheDocument()
    })
```

lukaw3d: yep; but I don't think we have designs for that anywhere yet. Not sure which components we would want compact notation and which wouldn't

buberdds: if it's not scoped for launch then this looks good

buberdds (approved):

PR #716, opened 2023-07-11
Graph feedback

lubej:
1. (Desktop) ParaTime should directly navigate when zoomed out.
2. When zoomed in the user should be able to see that status of the ParaTime.
3. When zoomed into a specific ParaTime any unrelated ParaTimes should have 50% opacity.
4. The outer stroke should be display on top of the graph content.
5. Reduce opacity on mobile tooltip.
6. Fix ‘Coming soon’ label displayed on mobile.
7. Shadows cut off on testnet graph.
8. Add selected network label.


buberdds:
1) Graph position if wrong. It's moved to right and bottom and it overlaps Selected Network
2) Emerald is broken on mobile (it should behave like Cipher)

src/app/pages/HomePage/Graph/Graph/index.tsx
```diff
@@ -218,24 +247,46 @@ const GraphCmp: ForwardRefRenderFunction<SVGSVGElement, GraphProps> = (
...
+    setOutOfDateMap({})
+  }, [network])
 
-  const disabledMap: Record<Layer, boolean> = {
-    [Layer.emerald]: isLayerDisabled(Layer.emerald),
-    [Layer.consensus]: isLayerDisabled(Layer.consensus),
-    [Layer.cipher]: isLayerDisabled(Layer.cipher),
-    [Layer.sapphire]: isLayerDisabled(Layer.sapphire),
-  }
+  const isLayerDisabled = useCallback(
```
buberdds, line 266: Why we need to wrap `getEnabledLayersForNetwork` with useCallback when disabledMap is going to be memoized?

src/app/pages/HomePage/Graph/ParaTimeSelector/index.tsx
```diff
@@ -212,7 +219,13 @@ const ParaTimeSelectorCmp: FC<ParaTimeSelectorProps> = ({
...
   }
 
-  const isZoomedIn = scale > 1.005
+  const clearSelectedLayer = () => {
+    if (selectedLayer?.current) {
+      setSelectedLayer(undefined)
+    }
+  }
+
+  const isZoomedIn = scale > 1.07
```
buberdds, line 228: If it's not a problem can you add a comment what 1.07 means and why it's correct? just no idea why 1.005 is wrong and 1.07 fixes something.

src/app/pages/HomePage/Graph/ParaTimeSelector/index.tsx
```diff
@@ -126,10 +126,17 @@ const ZoomOutBtnFade = styled(Fade)(() => ({
...
+const QuickPinchZoomOuter = styled('div')(({ theme }) => ({
   '> div': {
     position: 'absolute',
     inset: 0,
+    border: '10px solid transparent',
+    borderRadius: '50%',
+  },
+  [theme.breakpoints.up('sm')]: {
+    '> div': {
+      border: '1.7cqmin solid transparent',
```
buberdds, line 138: How this value is computed, where we defined a container?

src/app/pages/HomePage/Graph/Graph/index.tsx
```diff
@@ -218,24 +247,46 @@ const GraphCmp: ForwardRefRenderFunction<SVGSVGElement, GraphProps> = (
...
+    () => ({
+      [Layer.emerald]: isLayerDisabled(Layer.emerald),
+      [Layer.consensus]: isLayerDisabled(Layer.consensus),
+      [Layer.cipher]: isLayerDisabled(Layer.cipher),
+      [Layer.sapphire]: isLayerDisabled(Layer.sapphire),
+    }),
+    [isLayerDisabled],
+  )
+
+  const enabledLayers: Layer[] = useMemo(
```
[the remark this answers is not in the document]
lubej: It is used on multiple places in the graph, and it just an abstraction, to not repeat the function call.

src/app/pages/HomePage/Graph/Graph/index.tsx
lubej: Due to network dependency. Status call depends on disabledMap, which I want to trigger only when network changes.

src/app/pages/HomePage/Graph/Graph/index.tsx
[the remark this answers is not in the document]
buberdds: but why  we need `Object.keys(disabledMap).filter(layer => !disabledMap[layer as Layer]).map(layer => layer as Layer),` to get `enabledLayers` when `RouteUtils.getEnabledLayersForNetwork(network)`  returns it?

buberdds (approved):

PR #723, opened 2023-07-12
Handle transaction evm.Call method that is likely a transfer

buberdds:
BE is adding an optional boolean when runtime tx method is evm.Call. In this case we want to swap type so it looks like transfer in UI.


csillag (approved):
Reviewed. LGTM.

PR #727, opened 2023-07-12
Disable automatically setting AddressSwitch based on URL

lukaw3d:
Default to ETH instead. This shows more consistent information when seeing consensus withdrawals in transactions list (oasis tx hash, from eth, to oasis1) and then opening it:
- before: oasis tx hash, from oasis1, to oasis1
- after: oasis tx hash, from eth, to oasis1


csillag (approved):
LGTM.

csillag:
Changelog fragment!

lubej (approved):

PR #728, opened 2023-07-12
Add missing 0x prefix to eth hashes in runtime events

csillag:
| Before | After |
| --- | --- |
|  image   |  image |


src/oasis-nexus/api.ts
```diff
@@ -562,6 +563,7 @@ export const useGetRuntimeEvents: typeof generated.useGetRuntimeEvents = (
               }
               return {
                 ...event,
+                eth_tx_hash: `0x${event.eth_tx_hash}`,
```
lukaw3d, line 566: This often results in `eth_tx_hash: "0xundefined"`

lukaw3d (approved):

PR #731, opened 2023-07-13
Format large numbers in token total supply

buberdds:
Based on https://github.com/oasisprotocol/explorer/pull/715#discussion_r1260005955 it's not clear how/where/what we want to round. I added optional `roundLargeNumbers` prop so for now only two places are affected.

### DIFF
--- a/src/app/components/RoundedBalance/index.tsx
+++ b/src/app/components/RoundedBalance/index.tsx
@@ -14,6 +14,7 @@ type RoundedBalanceProps = {
   scope?: SearchScope
   tokenAddress?: string
   tickerAsLink?: boolean | undefined
+  compactLargeNumbers?: boolean
 }
 
 const numberOfDecimals = 5
@@ -24,6 +25,7 @@ export const RoundedBalance: FC<RoundedBalanceProps> = ({
   scope,
   tokenAddress,
   tickerAsLink,
+  compactLargeNumbers,
 }) => {
   const { t } = useTranslation()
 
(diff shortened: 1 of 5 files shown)


src/app/components/RoundedBalance/index.tsx
```diff
@@ -24,6 +25,7 @@ export const RoundedBalance: FC<RoundedBalanceProps> = ({
   scope,
   tokenAddress,
   tickerAsLink,
+  roundLargeNumbers,
```
lukaw3d, line 28: Needs a different name (this isnt rounding) (although the whole RoundedBalance isnt rounding either, it's truncating)

src/app/components/RoundedBalance/index.tsx
```diff
@@ -43,6 +45,30 @@ export const RoundedBalance: FC<RoundedBalanceProps> = ({
...
+      <Tooltip
+        arrow
+        placement="top"
+        title={t('common.valueInToken', { value: number.toFormat(), ticker: tickerName })}
+        enterDelay={tooltipDelay}
+        enterNextDelay={tooltipDelay}
+      >
+        <span>
+          {t('common.valuePair', {
+            value: number.toFixed(),
```
lukaw3d, line 59: Hm. How does this work :O this makes a string. I guess it silently does parseFloat?

buberdds: 🤷  Output seems Ok when I look at token pages

lukaw3d (approved):

src/app/components/RoundedBalance/index.tsx
lukaw3d: it can take strings :O https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/format#using_format_with_a_string

PR #738, opened 2023-07-14
Improve token transfer listing for tokens

csillag:
~**This depends on #743.**~

Improve pagination experience by proving rudimentary client-side pagination (limitation: looking only at the first 1000 records)

This impacts two pages:
- Tokens's token transfer listing (for example  `/mainnet/sapphire/token/0x998633BDF6eE32A9CcA6c9A247F428596e8e65d8` )
- Account's token transfer listing (for example `/mainnet/sapphire/address/0x998633BDF6eE32A9CcA6c9A247F428596e8e65d8/token-transfers#transfers`

~I think there is some problem with the pagination widget sometimes displaying the wrong number of total pages, but that is totally unrelated.~ **Update:** workaround added for anomalous behavior.

Technical note: great emphasis has been placed in developing this in separate commits so that it's easier to follow what is going on, so I suggest reviewing commit by commit.

### DIFF
--- a/src/app/components/Table/useClientSidePagination.ts
+++ b/src/app/components/Table/useClientSidePagination.ts
@@ -83,7 +83,12 @@ export function useClientSizePagination<Item, QueryResult extends List>({
       const tableProps: TablePaginationProps = {
         selectedPage: selectedClientPage,
         linkToPage,
-        totalCount: filteredData?.length,
+        totalCount: filteredData
+          ? // This correction is here to simulate the bogus behavior of server-side pagination
+            // Remove this when server-side pagination is fixed,
+            // and so the work-around in the pagination widget is fixed.
+            filteredData.length - clientPageSize * (selectedClientPage - 1)
+          : undefined,
         isTotalCountClipped: queryResult?.is_total_count_clipped, // TODO
         rowsPerPage: clientPageSize,
       }
--- a/src/app/components/Table/useClientSidePagination.ts
+++ b/src/app/components/Table/useClientSidePagination.ts
@@ -1,0 +1,91 @@
+import { To, useSearchParams } from 'react-router-dom'
+import { AppErrors } from '../../../types/errors'
+import { ComprehensivePaginationEngine } from './PaginationEngine'
+import { List } from '../../../oasis-nexus/api'
+import { TablePaginationProps } from './TablePagination'
+
+type ClientSizePaginationParams<Item> = {
+  paramName: string
+  pageSize: number
+  /**
+   * @deprecated this will mess up page size.
+   *
+   * Consider using client-side pagination instead.
+   */
+  filter?: (item: Item) => boolean
+}
+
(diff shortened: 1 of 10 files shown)


src/app/pages/TokenDashboardPage/hook.ts
```diff
@@ -41,10 +49,9 @@ export const useTokenTransfers = (scope: SearchScope, address: string) => {
     network,
     layer, // This is OK since consensus has been handled separately
     {
-      limit: NUMBER_OF_ITEMS_ON_SEPARATE_PAGE,
-      offset: offset,
+      ...pagination.paramsForQuery,
       rel: address,
-      // type: 'evm.log',
+      type: 'evm.log',
```
lukaw3d, line 54: First commit should be a separate PR.
Also, can we combine useTokenTransfers and useAccountTokenTransfers now?

csillag:
> First commit should be a separate PR.

OK, see #743

csillag:
>  can we combine useTokenTransfers and useAccountTokenTransfers now?

Done.

src/app/components/Table/useClientSidePagination.ts
```diff
@@ -0,0 +1,102 @@
...
+const knownListKeys: string[] = ['total_count', 'is_total_count_clipped']
+
+function findListIn<T extends List, Item>(data: T): Item[] {
+  const candidates = Object.keys(data).filter(
+    key => !knownListKeys.includes(key) && Array.isArray(data[key as keyof T]),
+  ) as (keyof T)[]
+  switch (candidates.length) {
+    case 0:
+      throw new Error('Data not found!')
+    case 1:
```
lukaw3d, line 23: With strict types this could be ensured at compile time, but not worth inventing it for this temporary workaround until backend adds filtering by signature

src/app/components/Table/useClientSidePagination.ts
```diff
@@ -0,0 +1,102 @@
...
+    paramsForQuery,
+    getResults: (queryResult, key) => {
+      const data = queryResult
+        ? key
+          ? (queryResult[key] as Item[])
+          : findListIn<QueryResult, Item>(queryResult)
+        : undefined
+      const filteredData = !!data && !!filter ? data.filter(filter) : data
+
+      const wantedStart = (selectedClientPage - 1) * clientPageSize
```
lukaw3d, line 79: I would use offset and limit for these calculations

lukaw3d (approved):
Looks easily reversible when we stop needing it

src/app/components/Table/useClientSidePagination.ts
csillag: Done.

csillag:
> Looks easily reversible when we stop needing it

Yes, there is `useComprehensiveSearchParamsPagination()`, which implements the exact same interface as the client-side pagination engine, so it's a drop-in replacement.

Issue #744, opened 2023-07-16
[Bug]: misleading error messages when asking for non-existent pages on paginated lists


csillag:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### SUMMARY

In several paginated lists, when someone asks for a page that is beyond the last page, we give an error message that says there is no data whatsoever, which is false. In other cases, there is no error message or way to get back to existing pages.

### STEPS TO REPRODUCE

No error message, where there should be something:

https://explorer.dev.oasis.io/mainnet/sapphire/block/657081?page=2
https://explorer.dev.oasis.io/mainnet/sapphire/token?page=4

Wrong/misleading error message:

https://explorer.dev.oasis.io/mainnet/sapphire/token/0x998633BDF6eE32A9CcA6c9A247F428596e8e65d8?page=22
https://explorer.dev.oasis.io/mainnet/sapphire/token/0x998633BDF6eE32A9CcA6c9A247F428596e8e65d8/holders?page=5
https://explorer.dev.oasis.io/mainnet/sapphire/address/0x46b983CF8f03db357Cf10ADE79a65eC76DE0BF49?page=2
https://explorer.dev.oasis.io/mainnet/sapphire/address/0x46b983CF8f03db357Cf10ADE79a65eC76DE0BF49/token-transfers?page=2#transfers

### EXPECTED BEHAVIOR

There are several possibilities here:

- Show a dedicated error message (saying that the requested page is beyond the end)
- Automatically jump back to the first page
- Automatically jump back to the last page (which is the closest existing one in relative to what was asked)

We should choose a way and apply in consistently.

csillag:
> Automatically jump back to the last page (which is the closest existing one in relative to what was asked)

Actually, this is hard to implement without getting some info from the backend about the number of results, which is currently not guaranteed to be present (in case of clipped results when we go beyond 1000)

csillag:
As far as I can tell, currently we only show this when we get an invalid page number (ie. something that is not a valid number), but not when it's beyond the end. (And even then, we don't show the link as a link.)

But now that I know about this visual, the requirements are clear. (We just need to implement this everywhere.)

csillag:
Fixes are coming in https://github.com/oasisprotocol/explorer/pull/746

csillag:
Closed by https://github.com/oasisprotocol/explorer/pull/746

PR #746, opened 2023-07-17
Clean up pagination-related error pages everywhere

csillag:
Un-break pagination-related error pages.

See https://github.com/oasisprotocol/explorer/issues/744 for more context.

| Before | After |
| --- | --- |
| [test 1](http://localhost:1234/mainnet/sapphire/address/0x46b983CF8f03db357Cf10ADE79a65eC76DE0BF49?page=100x) | |
| image | image  |
| [test 2](http://localhost:1234/mainnet/sapphire/address/0x46b983CF8f03db357Cf10ADE79a65eC76DE0BF49?page=100) | |
|  image  | image  |
| [test 3](http://localhost:1234/mainnet/sapphire/address/0x46b983CF8f03db357Cf10ADE79a65eC76DE0BF49/token-transfers?page=1x) | |
| image  | image  |
| [test4](http://localhost:1234/mainnet/sapphire/address/0x46b983CF8f03db357Cf10ADE79a65eC76DE0BF49/token-transfers?page=2) |  |
|  image   | image  |
| [test5](http://localhost:1234/mainnet/sapphire/block?page=2x) | |
|  image  | image  |
| [test6](http://localhost:1234/mainnet/sapphire/block?page=2000000) | |
|  image  | image  |
| [test7](http://localhost:1234/mainnet/sapphire/block/158695?page=2) | |
|  image  | image  |
| [test8](http://localhost:1234/mainnet/sapphire/token?page=30) | |
|   image |  image  |
| [test9](http://localhost:1234/mainnet/sapphire/tx?page=5000000000) | |
|  image  |  image  |
| [test10](http://localhost:1234/mainnet/sapphire/token/0x998633BDF6eE32A9CcA6c9A247F428596e8e65d8?page=2x) | |
| image | image |
| [test11](http://localhost:1234/mainnet/sapphire/token/0x998633BDF6eE32A9CcA6c9A247F428596e8e65d8?page=200) | |
| image | image | 
| [test12](http://localhost:1234/mainnet/sapphire/token/0x998633BDF6eE32A9CcA6c9A247F428596e8e65d8/holders?page=4x) | |
| image  |     image   |
| [test13](http://localhost:1234/mainnet/sapphire/token/0x998633BDF6eE32A9CcA6c9A247F428596e8e65d8/holders?page=20) | |
| image  |  image  |

### DIFF
--- a/src/app/components/Table/GoToFirstPageLink.tsx
+++ b/src/app/components/Table/GoToFirstPageLink.tsx
@@ -1,0 +1,46 @@
+import { FC } from 'react'
+import { useTranslation } from 'react-i18next'
+import { useSearchParams, useHref } from 'react-router-dom'
+import Link from '@mui/material/Link'
+
+/**
+ * Link to attempt to jump to the first page (in pagination)
+ *
+ * This is only a best-effort attempts, because we can't really know the name of the 'page' parameter
+ * used for pagination. The best we can do is to test the trivial option ("page"), but if it's not there,
+ * then we can't really do anything.
+ *
+ * Please note that to get to an invalid / non-existent page, the user probably had to do some manual URL hacking,
+ * so hopefully it's not terrible that he gets a full reload as the result of that.
+ */
+export const GoToFirstPageLink: FC = () => {
+  const { t } = useTranslation()
+  const [searchParams] = useSearchParams()
+  const wantedParamName = searchParams.has('page') ? 'page' : undefined
+  const label = t('errors.loadFirstPage')
+  const newSearchParams = new URLSearchParams(searchParams)
+  if (wantedParamName) {
+    newSearchParams.delete(wantedParamName)
+  }
+  const href = useHref({
+    search: newSearchParams.toString(),
+  })
+  // Ideally, we should use a router link here, but even if we change the URL without a reload,
+  // I can't get the error boundary to recover from the error state.
+  // when jumping to the first page, the ErrorBoundary component stays in place, only the internals change.
+  // But to show the first page, the Error Boundary component would need to switch back from the "error" state
+  // to the "normal" state, which doesn't happen automatically.
+  // In needs a manual `setState()` for recovering, which is hard to channel through,
+  // but even if we do it (which I did as an experiment), we find that
+  // the data around the ErrorBoundary is somewhat weird, and the updated URL is not reflected right away,
+  // so we just end up running into the same exception again...
+  // Apparently the solution is to umount and rerender the component outside the error boundary.
+  // But that's hard to do selectively, so it's easier to just do a full refresh.
+  return wantedParamName ? (
+    <Link href={href} sx={{ cursor: 'pointer' }}>
+      {label}
+    </Link>
+  ) : (
+    <span>{label}</span>
+  )
+}
--- a/src/app/pages/AccountDetailsPage/index.tsx
+++ b/src/app/pages/AccountDetailsPage/index.tsx
@@ -12,10 +12,9 @@ import { EvmToken, EvmTokenType, RuntimeAccount } from '../../../oasis-nexus/api
 import { accountTokenContainerId } from './AccountTokensCard'
 import { useAccount } from './hook'
 import { useRequiredScopeParam } from '../../hooks/useScopeParam'
-import { showEmptyAccountDetails } from '../../../config'
 import { CardEmptyState } from './CardEmptyState'
 import { contractCodeContainerId } from './ContractCodeCard'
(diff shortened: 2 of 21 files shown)


src/app/components/Table/GoToFirstPageLink.tsx
```diff
@@ -0,0 +1,33 @@
...
+import { useTranslation } from 'react-i18next'
+import { useSearchParams, useHref } from 'react-router-dom'
+import Link from '@mui/material/Link'
+
+export const GoToFirstPageLink: FC = () => {
+  const { t } = useTranslation()
+  const [searchParams] = useSearchParams()
+  // We have no sure way to know the number of the page parameter to remove.
+  // The best we can do is to test the trivial option.
+  // If it's not there, then we can't really do anything.
```
lukaw3d, line 11: this should be responsibility of the page components

csillag: The problem is that this component is rendered by the ErrorDisplayer, which is used by the ErrorBoundary, which is placed _outside_ the part that is doing the pagination. So it's impossible to pass the information down.

Theoretically it could be passed _up_ (inside the Error object), but we for various reasons we can't throw proper Errors (because that would trigger unwanted react behavior), just these strings, which can't carry information.

We could extend the string data structure somehow, but it's a pain...

src/app/components/Table/GoToFirstPageLink.tsx
```diff
@@ -0,0 +1,33 @@
...
+    newSearchParams.delete(wantedParamName)
+  }
+  const href = useHref({
+    search: newSearchParams.toString(),
+  })
+
+  // Ideally, we should use a router link here, but even if we change the URL without a reload,
+  // I can't get the error boundary to recover from the error state.
+  // I tried to implement recovery, but it added too much complexity,
+  // so it's easier to just reload in these cases.
```
lukaw3d, line 25: I don't understand

- https://7ffc492f.oasis-explorer.pages.dev/mainnet/sapphire/address/0x46b983CF8f03db357Cf10ADE79a65eC76DE0BF49?page=100x
- click Token Transfers
- click Transactions
- working state

is this no-reload recovered state you are trying to get?

csillag: The above process works by switching tabs, which means unmounting and re-rendering the whole DOM of the tab. So the `ErrorBoundary` component is fully destroyed and re-rendered. 

However, what we need here would be that the `ErrorBoundary` component stays in place, we just navigate _inside_ the same tab (to a different page). In order to actually need the first page, the Error Boundary component would need to switch back from the "error" state to the "normal" state, which doesn't happen automatically.

In needs a manual `setState()` for recovering,  which is hard to channel through, but even if we do it (which I did as an experiment), we find that the data around the Error Boundary is somewhat weird, and the updated URL is not reflected right away, so we just end up running into the same exception again...

So apparently the solution is to umount and rerender the same component outside the error boundary.

But that's hard to do selectively, so it's easier to just to a full refresh.

Please note that to get to an invalid / non-existent page, the user probably needs to do some manual URL hacking. I don't think it's unacceptable behavior on our part that he gets a full reload as a result.

lukaw3d: hm, alright

Can you add more of this context to the comment, and add a component-level comment this component is only "best effort", and is only rarely triggered by user changing URL

src/app/pages/AccountDetailsPage/index.tsx
```diff
@@ -71,26 +64,22 @@ export const AccountDetailsPage: FC = () => {
           tokenPriceInfo={tokenPriceInfo}
         />
       </SubPageCard>
-      {showDetails && (
-        <RouterTabs
-          tabs={[
-            { label: t('common.transactions'), to: txLink, visible: showTxs },
```
lukaw3d, line 77: Simplify RouterTabs too?

src/app/components/Table/GoToFirstPageLink.tsx
csillag: OK done

src/app/pages/AccountDetailsPage/index.tsx
csillag: I think being able to hide tabs is a generally useful feature (for the tab menu), so I would't remove it unless it becomes uncomfortable to maintain for some reason.

(Also, please note that this feature is better than simply _not_ listing an item as part of the menu, because as implemented now, an "invisible" tab can still be validly selected and displayed when accessing via a link, just not offered up for selection by default.)

PR #753, opened 2023-07-18
Fix alignment of fiat value tooltip icon in Account details

buberdds:


buberdds:
I will wait for Don's review in this case.

PR #758, opened 2023-07-19
Fix hovered layer on Firefox

lubej:


src/app/pages/HomePage/Graph/Graph/index.tsx
```diff
@@ -367,6 +366,7 @@ const GraphCmp: ForwardRefRenderFunction<SVGSVGElement, GraphProps> = (
         ref={ref}
         className={network}
         selectedLayer={selectedLayer}
+        hoveredLayer={hoveredLayer}
```
buberdds, line 369: GraphStyledProps are missing  `hoveredLayer: Layer | null`

buberdds (approved):
Can we revert `https://github.com/oasisprotocol/explorer/pull/752/files` if it's not needed anymore?

lubej:
> Can we revert `https://github.com/oasisprotocol/explorer/pull/752/files` if it's not needed anymore?

I think it's better to leave it.

Issue #754, opened 2023-07-18
[Bug]: Homepage isZoomedIn/scale issue


buberdds:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### STEPS TO REPRODUCE

Is hardcoded scale factor reliable?
https://github.com/oasisprotocol/explorer/blob/master/src/app/pages/HomePage/Graph/ParaTimeSelector/index.tsx#L231

- open homepage
- make browser height window smaller 
- at some point "zoom in" state will be triggered. I can see "Zoom out" button, transparent Search and ParaTimes status without interacting with Graph. Depends on browser size a default Graph state can be rendered with `isZoomedIn` set to true.

Is this expected behaviour or bug?

PR #759, opened 2023-07-19
Fix Graph zoomed in value

lubej:
Closes https://github.com/oasisprotocol/explorer/issues/754

Due to viewport changes, scale can differ. `1.5` scale, looks safe enough.

### DIFF
--- a/src/app/pages/HomePage/Graph/ParaTimeSelector/index.tsx
+++ b/src/app/pages/HomePage/Graph/ParaTimeSelector/index.tsx
@@ -126,6 +126,7 @@ const ZoomOutBtnFade = styled(Fade)(() => ({
   transitionDelay: '500ms !important',
 }))
 
+// border affecting scale (quick-pinch-zoom)
 const QuickPinchZoomOuter = styled('div')(({ theme }) => ({
   '> div': {
     position: 'absolute',
@@ -152,6 +153,7 @@ interface ParaTimeSelectorProps extends ParaTimeSelectorBaseProps {
   step: ParaTimeSelectorStep
   setStep: (value: ParaTimeSelectorStep) => void
   showInfoScreen: boolean
+  graphZoomedIn: boolean
   onGraphZoomedIn: (isGraphZoomedIn: boolean) => void
 }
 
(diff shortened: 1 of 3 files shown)


src/app/pages/HomePage/Graph/ParaTimeSelector/index.tsx
```diff
@@ -227,8 +227,8 @@ const ParaTimeSelectorCmp: FC<ParaTimeSelectorProps> = ({
     }
   }
 
-  // true when scale larger than initial zoom
-  const isZoomedIn = scale > 1.07
+  // true when scale larger than "safe" zoom, due to viewport
+  const isZoomedIn = scale > 1.5
```
buberdds, line 231: It's fine as hotfix, but looking at evolution here from `parseFloat(scale.toFixed(2)) > 1` to `scale > 1.005` to `scale > 1.07` to `scale > 1.5`. Why are setting zoom flag in render cycle. Maybe putting it in something like `useLayoutEffect` will fix that (no idea, just throwing ideas).

buberdds: Ok I think it stopped working, because we changed styles lately in `QuickPinchZoomOuter > div`. We apply styles to that third party container and it sets initial zoom factor based on container and child dimensions. Ideally we should not apply custom styled `QuickPinchZoom`, but whatever 

If above is true
1. Can you add a comment to `QuickPinchZoomOuter` that any changes affect `isZoomedIn` scale factor?
2. We could limit scale issue by moving most styles to outer component (limit `  '> div'`)
```
const QuickPinchZoomOuter = styled('div')(({ theme }) => ({
  position: 'absolute',
  inset: 0,
  borderWidth: 10,
  borderStyle: 'solid',
  borderRadius: '50%',
  borderColor: theme.palette.layout.darkBorder,
  '> div': {
    position: 'absolute',
    inset: 0,
  },
  [theme.breakpoints.up('sm')]: {
    borderWidth: 15,
  },
}))
```
3) Don't apply '> div'` styles at all, but no idea if that's possible

lubej: Right, moved into bellow useEffect, which sets zoomedIn the parent component.

buberdds (approved):

src/app/pages/HomePage/Graph/ParaTimeSelector/index.tsx
lubej: Right, this is required to the border, in order case it would beed to be included into svg. Which makes it more difficult to display the border on top of the graph.

PR #761, opened 2023-07-20
Improve performance when displaying event lists, using newly available data

csillag:
~This depends on https://github.com/oasisprotocol/nexus/pull/491.~

Please note that while this version is backward compatible with the currently deployed staging or production versions of nexus, the performance improvements will only be realized if the new version of nexus is deployed.

### DIFF
--- a/src/app/components/Tokens/TokenTransfers.tsx
+++ b/src/app/components/Tokens/TokenTransfers.tsx
@@ -15,7 +15,7 @@ import ArrowForwardIcon from '@mui/icons-material/ArrowForward'
 import { TokenTransferIcon } from './TokenTransferIcon'
 import { RoundedBalance } from '../RoundedBalance'
 import { useScreenSize } from '../../hooks/useScreensize'
-import { fromBaseUnits } from '../../utils/helpers'
+import { fromBaseUnits, getEthAccountAddressFromBase64 } from '../../utils/helpers'
 import Skeleton from '@mui/material/Skeleton'
 import { TokenLink } from './TokenLink'
 import { PlaceholderLabel } from '../../utils/PlaceholderLabel'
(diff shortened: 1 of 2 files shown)


src/app/components/Tokens/TokenTransfers.tsx
```diff
@@ -234,16 +277,23 @@ export const TokenTransfers: FC<TokenTransfersProps> = ({
...
             ]
           : []),
         {
           key: 'value',
           align: TableCellAlign.Right,
-          content: <DelayedEventBalance event={transfer} tickerAsLink={differentTokens} />,
+          content: transfer.evm_log_details?.token_type ? (
+            <DirectEventBalance event={transfer} tickerAsLink={differentTokens} />
+          ) : (
+            <DelayedEventBalance event={transfer} tickerAsLink={differentTokens} />
```
lukaw3d, line 295: Names of these components are indistinguishable

csillag: OK, so the idea was the the first one (`DirectEventBalance`) uses that that is directly available, while the second one (`DelayedEventBalance`) launches another request, and will display the data when the new server roundtrip has finished.

What names would you suggest?

lukaw3d: I would delete DirectEventBalance, and change EventBalance

csillag: OK done

csillag:
Tested with latest staging Nexus deployment, seems to work fine. (For testing, I had to change the API endpoints to staging, and enable Emerald.)

src/app/components/Tokens/TokenTransfers.tsx
```diff
@@ -234,16 +262,23 @@ export const TokenTransfers: FC<TokenTransfersProps> = ({
...
             ]
           : []),
         {
           key: 'value',
           align: TableCellAlign.Right,
-          content: <DelayedEventBalance event={transfer} tickerAsLink={differentTokens} />,
+          content: transfer.evm_log_details?.token_type ? (
+            <EventBalance event={transfer} tickerAsLink={differentTokens} />
+          ) : (
+            <DelayedEventBalance event={transfer} tickerAsLink={differentTokens} />
```
lukaw3d, line 280: Can delayed versions of these components ever actually obtain more information, if backend JOIN couldn't?
Or is this so that we can deploy before API is updated, and later remove it?

csillag: Yes, the intention behind the current code is that it should work optimally if the newly added data is present, and fall back gracefully to lower performance if it's not.

As far as I can understand, this data will _always_ be there _for transfers_, so _maybe_ we can remove the delayed codepath later, but then again, it's a but vague what do we consider to be a transfer, so I am not enterely certain about that. (But I'll try to check with the Nexus devs.)

csillag:
Now that the relevant Nexus changes have been merged, this PR has also been updated accordingly.

csillag:
Please note that while this version still works with the currently deployed staging or production versions of nexus, the performance improvements will only be realized if the new version of nexus is deployed.

PR #762, opened 2023-07-20
Improve performance when displaying token balance list

csillag:
~This depends on https://github.com/oasisprotocol/nexus/pull/492.~
~This also depends on #761. (Since the Nexus  PRs also depend on each other.)~

Please note that while this version is backward compatible with the currently deployed staging or production versions of nexus, the performance improvements will only be realized if the new version of nexus is deployed.

### DIFF
--- a/.changelog/762.internal.md
+++ b/.changelog/762.internal.md
@@ -1,0 +1,1 @@
+When listing token balances, use locally available data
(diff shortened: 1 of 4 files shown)


csillag:
Tested with latest staging Nexus deployment, seems to work fine. (For testing, I had to change the API endpoints to staging, and enable Emerald.)

csillag:
Now that the relevant Nexus changes have been merged, this PR has also been updated accordingly.

Issue #501, opened 2023-06-12
[Bug]: PageLayout stories are broken


buberdds:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### STEPS TO REPRODUCE

Open PageLayout stories

### EXPECTED BEHAVIOR

The First issue is 
`You must either specify both network and layer or none of them. You can not have one but not the other.` which is related to `useScopeParam` 

that can be fix by updating emeraldRoute to 
```
const emeraldRoute = {
  routePath: '/:network/:layer',
  routeParams: { network: 'mainnet', layer: 'emerald' },
}
```

The second issue
`No QueryClient set, use QueryClientProvider to set one`
Did not look at this yet, but PageLayout is not longer presentation/dumb component and prob we need to provide QueryClientProvider to `.storybook/preview.js` and mock responses. Maybe there's some lib/addon to handle that out of a box.

PR #787, opened 2023-08-03
Fix Storybook issues

buberdds:
Closes https://github.com/oasisprotocol/explorer/issues/501

### DIFF
--- a/src/stories/PageLayout.stories.tsx
+++ b/src/stories/PageLayout.stories.tsx
@@ -2,6 +2,8 @@ import { Meta, StoryFn, StoryObj } from '@storybook/react'
 import { withRouter } from 'storybook-addon-react-router-v6'
 import { EmptyState } from '../app/components/EmptyState'
 import { PageLayout } from '../app/components/PageLayout'
+import { Network } from '../types/network'
+import { Layer } from '../oasis-nexus/api'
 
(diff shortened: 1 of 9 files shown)


src/stories/PageLayout.stories.tsx
```diff
@@ -15,36 +17,36 @@ const Template: StoryFn<typeof PageLayout> = args => (
   </PageLayout>
 )
 
-const emeraldRoute = {
-  routePath: '/:layer',
-  routeParams: { layer: 'emerald' },
+const sapphireRoute = {
+  routePath: '/:network/:layer',
+  routeParams: { network: Network.mainnet, layer: Layer.sapphire },
 }
```
csillag, line 24: Funny thing, I think this is char by char the same as my (unfinished) https://github.com/oasisprotocol/explorer/pull/502/files.

csillag (approved):
OK, this obviously works, but I have very little idea about what the `mockServiceWorker` does, or why does it do it. I think a little bit of in-line documentation (for the whole file, not the the individual functions) would go a long way.

buberdds:
If we prefer not to store auto gen service worker file we can run `npx msw init .storybook/public` in yarn `start-storybook` and add a file to gitignore

PR #760, opened 2023-07-20
Improve performance when displaying token lists, using newly available data

csillag:
This depends on https://github.com/oasisprotocol/nexus/pull/493.

PR #796, opened 2023-08-09
Remove fallbacks for old Nexus versions

csillag:
When implementing #760, #761 and #762, we made it so that Explorer works with both the old and the new versions of Nexus. (But of course we have better performance with new Nexus.)

Now that the new version of Nexus is running everywhere, we can remove the code paths only used with old versions.


lukaw3d (commented):
I would rename to "Remove fallbacks for old Nexus API"

PR #797, opened 2023-08-09
Follow api stats changes

csillag:
See the API changes described here: https://github.com/oasisprotocol/nexus/pull/477

PR #812, opened 2023-08-22
Fix charts by syncing with the latest tx_volume stats API changes

buberdds:
- sync with new props
- follow new BE naming convention

Relates to https://github.com/oasisprotocol/nexus/pull/477

Charts on prod are broken due to API change. There's draft PR, but Kristof is out this week. I don't think we need to switch date-fns functions from [start* to end*](https://github.com/oasisprotocol/explorer/pull/797). Output of FE grouping will be the same. We need to sync prop names and use both query params (window_size_seconds, window_step_seconds) for tx_volume.

### DIFF
--- a/src/app/utils/chart-utils.ts
+++ b/src/app/utils/chart-utils.ts
@@ -14,7 +14,7 @@ export enum ChartDuration {
   ALL_TIME = 'ALL_TIME',
 }
 
-export const dailyLimitWithoutBuffer = (60 / 5) * 24 // full day for 5 minutes buckets,
+export const dailyLimitWithoutBuffer = (60 / 5) * 24 // full day for 5 minutes windows
 
 /*
   window_size_seconds refers to the number of data points in one chunk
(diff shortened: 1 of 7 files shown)


src/app/utils/chart-utils.ts
```diff
@@ -4,35 +4,42 @@ import isSameMonth from 'date-fns/isSameMonth'
...
 }
 
-export const dailyLimitWithoutBuffer = (60 / 5) * 24 // full day for 5 minutes buckets,
+export const dailyLimitWithoutBuffer = (60 / 5) * 24 // full day for 5 minutes windows
 
 export const durationToQueryParams = {
   [ChartDuration.TODAY]: {
-    bucket_size_seconds: 60 * 5,
+    window_size_seconds: fiveMinutesWindowSize,
+    window_step_seconds: fiveMinutesWindowSize,
```
lukaw3d, line 22: Can you add comment about the difference

buberdds: done

lukaw3d (approved):

src/app/utils/chart-utils.ts
lukaw3d: Oh oops, hovering on them already explains the difference. It shows field docs because of `satisfies { [duration in ChartDuration]: GetLayerStatsTxVolumeParams }`

PR #822, opened 2023-08-24
Show number of token transfers in Token page details

buberdds:
I am not entirely sure, but during one of a meeting we wanted to show total number of token transfers in tokens details page. It looks like a prop was added lately (not deployed yet).

### DIFF
--- a/.changelog/822.trivial.md
+++ b/.changelog/822.trivial.md
@@ -1,0 +1,1 @@
+Show number of token transfers in Token page details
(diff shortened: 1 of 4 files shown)


lukaw3d (approved):
I think we also want it in `Token Snapshot` when it gets deployed

buberdds:
> I think we also want it in `Token Snapshot` when it gets deployed

Ok I will double check that.

PR #892, opened 2023-09-26
Use new runtime freshness info

### LINKED ISSUE #882: Use new /status fields
See new fields https://github.com/oasisprotocol/explorer/pull/799
Usable in e.g. useRuntimeFreshness

csillag:
Closes #882.

### DIFF
--- a/.changelog/892.internal.md
+++ b/.changelog/892.internal.md
@@ -1,0 +1,1 @@
+Use new runtime freshness info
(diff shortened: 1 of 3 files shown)


csillag:
Turns out that the new field doesn't really mean what I thought it meant. This needs to be clarified.

csillag:
According to today's meeting, we want to refine this some more, so I'm putting this back to draft state.

PR #893, opened 2023-09-26
Handle new event types (placeholders)

csillag:
The goal of this PR is to remove syntax errors introduced by the latest auto-generated API wrapper code in #888.

So this incorporated the new API code from that PR, and also the required fixes in other related code.

Please note that [Handle new event types “consensus.Delegate”,  “consensus.Undelegate”](https://app.clickup.com/t/8692qqm49) is still required; this code only adds placeholders, marked with TODO comments, where the actual new stuff will have to go. (Icons, data fields, etc.)

### DIFF
--- a/.changelog/893.internal.md
+++ b/.changelog/893.internal.md
@@ -1,0 +1,1 @@
+Add basic handling of new event types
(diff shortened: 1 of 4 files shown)


csillag:
I have updated this PR to contain only that which has been left out of #906.

csillag:
This has been refreshed again.

csillag:
Please note that now there is data on the nexus DB which use these new types, and trying to explore that data leads to a crash in the current explorer. This PR fixes that. (But still, this is not ready for review, since it's intentionally a placeholder only, to avoid syntax errors and crashes)

PR #904, opened 2023-09-28
TX details: support switching between hex and base64 data display

csillag:


matevz:
I propose we use the existing Oasis<->ETH toggle for switching between Oasis and ETH format. The intuition is the following:
- Ethereum tooling use hex, so you won't come far with Base64 without a 3rd party decoder
- Oasis tooling (e.g. Oasis CLI, oasis-node logs) use Base64 format everywhere

csillag:
A friendly reminder: this is still waiting for feedback from @donouwens

csillag:
Could you please check the current implementation?

csillag:
@donouwens I think we should discuss this a bit more, since after I implemented it, @matevz has [explained](https://github.com/oasisprotocol/explorer/pull/904#issuecomment-1739077138) that he has actually meant something different...

PR #905, opened 2023-09-28
TX details: always show raw data

csillag:
Even for encrypted transactions, now we can show the raw data, like this:

image


csillag:
@kostko says:

> Yeah that looks right

matevz (approved):

PR #909, opened 2023-09-29
Add NFT instance details page

csillag:
See the design [here](https://www.figma.com/file/8dwSyYOOm3vlgbvxPGT29p/Block-Explorer-(Post-Launch)?type=design&node-id=6092-212988&mode=design&t=IcBb5lsaTPunvu04-0) (Also move to the right)

image

Currently powered by mock data. Obviously, the metadata and history tabs are not yet there, waiting for the backend on those. (Even the format is not yet known)

W.I.P.

Built on top of #893, in order to avoid syntax errors caused by importing the latest API definition.

### DIFF
--- a/.changelog/909.feature.md
+++ b/.changelog/909.feature.md
@@ -1,0 +1,1 @@
+Add NFT feature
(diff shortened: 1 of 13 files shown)


buberdds:
I will re-write commit history after CR.

Testing URL
https://2c4654de.oasis-explorer.pages.dev//mainnet/emerald/token/oasis1qqj54vxucwtn74pkhh8xp29aatkplzh0wuhmzwa5/instance/38

For local testing URL with image - stage API only (REACT_APP_API)
http://localhost:1234/mainnet/emerald/token/oasis1qzwx9zys4g89a8nfc9qg3e8y6cpwfyjatvznjknd/instance/242

API can throw errors like `{"msg":"storage error: can't scan into dest[14]: cannot scan NULL into *string"}`. BE is aware of that.

PR #910, opened 2023-09-29
RouterTabs: add support for hierarchical routes

csillag:
Added new, opt-in behavior:

If we have tabs like `/a`, `/b`, `/c`,
and the current path is `/a/2`,
then we should be able to mark tab `a` as active.

The router already supports this, but until now,
the tab menu component itself did not.

This PR fixes that.

### DIFF
--- a/src/app/components/RouterTabs/index.tsx
+++ b/src/app/components/RouterTabs/index.tsx
@@ -3,22 +3,102 @@ import { NonScrollingRouterLink } from '../NonScrollingRouterLink'
 import Tabs from '@mui/material/Tabs'
 import Tab from '@mui/material/Tab'
 
+type TabConfig = {
+  label: string
+  to: string
+  visible?: boolean
+}
+
 type RouterTabsProps<Context> = {
-  tabs: {
-    label: string
-    to: string
-    visible?: boolean
-  }[]
+  tabs: TabConfig[]
   context?: Context
+
+  /**
+   * Do we want to highlight tabs as selected when using hierarchical routes?
+   *
+   * For example, we have tabs with these paths
+   *
+   * /item/a
+   * /item/b
+   * /item/c
+   *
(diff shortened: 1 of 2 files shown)


src/app/components/RouterTabs/index.tsx
```diff
@@ -3,22 +3,37 @@ import { NonScrollingRouterLink } from '../NonScrollingRouterLink'
...
+
+  if (matchPartialPath) {
+    const tabPaths = tabs.map(getPathname)
+    const shortestPath = tabPaths.sort((p1, p2) => p1.length - p2.length)[0]
+    const shortestPathIsRoot = tabPaths.every(path => path.startsWith(shortestPath))
+    const rootPath = shortestPathIsRoot ? shortestPath : undefined
+    currentTab = tabs.find(tab => {
+      const tabPath = getPathname(tab)
+      return tabPath === pathname || (tabPath !== rootPath && pathname.startsWith(tabPath))
+    })
```
lukaw3d, line 35: I honestly can't parse this code

csillag: Added more documentation in code comments

lukaw3d:
From sync meeting:

Happy path:
- `/token-transfers/subpath` would be a partial match for `/token-transfers` tab

Tougher cases:
- https://explorer.dev.oasis.io/mainnet/sapphire/address/0xCCD706B1c30d1c6F46F74885C3B13e3Ea28ce8AF
default tab at `/` (rootPath) is Transactions
- `/token-transfers` is not a partial match of `/` tab
- `/subpath` is not a partial match of `/` tab. If we wanted to support subpaths in default tab (rootPath) would we be able to differentiate subpaths that should apply outside the tabs?

buberdds:
For now we will go with simplified direct child https://github.com/oasisprotocol/explorer/pull/1049

PR #916, opened 2023-10-02
Remove unused getNameForTicker

lukaw3d:


lukaw3d:
I guess it's used in many places

PR #978, opened 2023-10-27
Update setup-python action for CI lint

csillag:


csillag:
No, this is not so simple...

PR #989, opened 2023-10-31
Display TX format for unencrypted transactions

csillag:
We have already been displaying format for encrypted transactions, but not for plain ones.
I moved the field (for encrypted TXs) higher, right next to type, and added the same field for plain transactions, too.

Encrypted TX:

image

Plain TX:

image


csillag:
@donouwens approves.

lukaw3d (approved):

PR #990, opened 2023-10-31
Show block-level events in block details 

csillag:
~Built on top of #991 for some refactoring.~

~Requires (and includes) #893 in order to avoid crashing on new event types.~

Design is [here](https://www.figma.com/file/8dwSyYOOm3vlgbvxPGT29p/Block-Explorer-(Post-Launch)?type=design&node-id=7528-52787&mode=design&t=b5Zb7ulkt0gHJMOl-0).

Can we tested with this path: `/testnet/sapphire/block/3038913`

### DIFF
--- a/src/app/components/RuntimeEvents/EventListFilterSwitch.tsx
+++ b/src/app/components/RuntimeEvents/EventListFilterSwitch.tsx
@@ -1,0 +1,71 @@
+import { FC } from 'react'
+import Chip from '@mui/material/Chip'
+import CheckIcon from '@mui/icons-material/Check'
+import { useTranslation } from 'react-i18next'
+import { COLORS } from '../../../styles/theme/colors'
+import { TFunction } from 'i18next'
+import Box from '@mui/material/Box'
+
+export const EventFilterMode = {
+  All: 'all',
+  NonTX: 'nonTX',
+} as const
+
+const eventFilterModes: EventFilterMode[] = [EventFilterMode.All, EventFilterMode.NonTX]
+
+const getEventFilterModeName = (t: TFunction, mode: EventFilterMode): string => {
+  switch (mode) {
+    case 'all':
+      return t('runtimeEvent.filter.all')
+    case 'nonTX':
+      return t('runtimeEvent.filter.nonTx')
+  }
+}
+
+const Pill: FC<{ label: string; selected: boolean; onSelect: () => void }> = ({
+  label,
+  selected,
+  onSelect,
+}) => {
+  return (
+    <Chip
+      icon={selected ? <CheckIcon htmlColor={'#FFFFFF'} /> : undefined}
+      label={label}
+      onClick={onSelect}
+      sx={{
+        display: 'flex',
+        height: 30,
+        padding: '3px 10px',
+        alignItems: 'center',
+        borderRadius: 9,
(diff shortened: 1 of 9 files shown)


csillag:
We decided to do the tweaking of the event filter in a separate PR.

csillag:
@lukaw3d this is pretty similar to #992 (already approved), but it's for blocks (not accounts), has the extra filtering. Can you please review this one, too? Then I can merge both at the same time. (And use the same filtering.)

src/app/components/RuntimeEvents/EventListFilterSwitch.tsx
```diff
@@ -0,0 +1,71 @@
...
+        border: `1px solid ${COLORS.brandMedium}`,
+        background: selected ? COLORS.brandMedium : COLORS.brandLightBlue,
+        color: selected ? 'white' : 'black',
+      }}
+    />
+  )
+}
+
+// eslint-disable-next-line @typescript-eslint/no-redeclare
+export type EventFilterMode = (typeof EventFilterMode)[keyof typeof EventFilterMode]
```
lukaw3d, line 50: How does this + `export const EventFilterMode = {` work

src/app/pages/BlockDetailPage/EventsCard.tsx
```diff
@@ -0,0 +1,81 @@
...
+}) => {
+  const { t } = useTranslation()
+  if (scope.layer === Layer.consensus) {
+    // Loading events for consensus blocks is not yet supported.
+    // Should use useGetConsensusEvents()
+    throw AppErrors.UnsupportedLayer
+  }
+  const eventsQuery = useGetRuntimeEvents(scope.network, scope.layer, {
+    block: blockHeight,
+    // TODO: search for tx_hase = null
```
lukaw3d, line 31:
```suggestion
    // TODO: search for tx_hash = null
```

src/app/pages/BlockDetailPage/EventsCard.tsx
```diff
@@ -0,0 +1,81 @@
...
+  filterMode,
+}) => {
+  const { t } = useTranslation()
+  if (scope.layer === Layer.consensus) {
+    // Loading events for consensus blocks is not yet supported.
+    // Should use useGetConsensusEvents()
+    throw AppErrors.UnsupportedLayer
+  }
+  const eventsQuery = useGetRuntimeEvents(scope.network, scope.layer, {
+    block: blockHeight,
```
lukaw3d, line 30: I would add `limit: 100, // We want to avoid pagination here, if possible`

src/app/components/RuntimeEvents/EventListFilterSwitch.tsx
csillag: On is a type, the other is a const. TS compiler is smart enough to figure out which one we want in each situation. (The type is actually eliminated during the transpilation phase, so in the JS we only have the const.)

In any case, this is not something I came up with here, we already have a similar solution for `Layer` and other things, too.

lukaw3d (approved):

src/app/pages/BlockDetailPage/EventsCard.tsx
csillag: Done

PR #1012, opened 2023-11-10
Prevent app crash when rendering new event types

buberdds:
Prevents regression in the future, follows https://github.com/oasisprotocol/explorer/pull/893, reported in https://app.clickup.com/t/8692qqm49

### DIFF
--- a/playwright/tests/accounts.spec.ts
+++ b/playwright/tests/accounts.spec.ts
@@ -1,0 +1,123 @@
+import { Page, expect, test } from '@playwright/test'
+import { RuntimeAccount, RuntimeEventList } from '../../src/oasis-nexus/api'
+
+async function setup(page: Page) {
+  await page.route(
+    'https://api.coingecko.com/api/v3/simple/price?ids=oasis-network&vs_currencies=usd',
+    route => {
+      // Don't respond
+    },
+  )
+  await page.route('**/v1/', route => {
+    // Don't respond
+  })
+  await page.route('**/v1/sapphire/status', route => {
+    // Don't respond
+  })
+  await page.route(
+    '**/v1/sapphire/transactions?limit=10&offset=0&rel=oasis1qq2v39p9fqk997vk6742axrzqyu9v2ncyuqt8uek',
+    route => {
+      route.fulfill({
+        body: JSON.stringify({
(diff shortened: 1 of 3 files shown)


playwright/tests/accounts.spec.ts
```diff
@@ -0,0 +1,123 @@
...
+        },
+      } satisfies Partial<RuntimeAccount>),
+    })
+  })
+  await page.route('**/v1/sapphire/events?rel=oasis1qq2v39p9fqk997vk6742axrzqyu9v2ncyuqt8uek', route => {
+    route.fulfill({
+      body: JSON.stringify({
+        is_total_count_clipped: false,
+        total_count: 3,
+        events: [
```
lukaw3d, line 54: I would add a type like `} satisfies Partial<RuntimeAccount>` above, in case any fields get renamed in the future

src/app/components/RuntimeEvents/RuntimeEventDetails.tsx
```diff
@@ -264,13 +264,17 @@ export const RuntimeEventDetails: FC<{
...
               {addressSwitchOption === AddressSwitchOption.Oasis && <CopyToClipboard value={event.body.to} />}
             </dd>
-            <dt>{t('runtimeEvent.fields.amount')}</dt>
-            <dd>
-              {t('common.valueInToken', {
-                ...getPreciseNumberFormat(event.body.amount.Amount),
-                ticker: event.body.amount.Denomination,
-              })}
-            </dd>
+            {event.body.amount?.Amount && event.body.amount?.Denomination && (
```
lukaw3d, line 267: I'd add TODO. This is only needed because some TODO events temporarily abuse this for rendering.

Or, just:
```tsx
    case RuntimeEventType.consensus_accountsdelegate: // TODO show this properly
    case RuntimeEventType.consensus_accountsundelegate_start: // TODO show this properly
    case RuntimeEventType.consensus_accountsundelegate_done: // TODO show this properly
      return (
        <div>
          <b>{event.type}</b>
          <br />
          <pre>{JSON.stringify(event, null, ' ')}</pre>
        </div>
      )
    default:
```

lukaw3d (approved):

Issue #1006, opened 2023-11-07
[UX]: Mobile layout doesn't show all tabs


lukaw3d:
### IMPACT

- [ ] High
- [X] Medium
- [ ] Low

### URGENCY

- [ ] Critical
- [ ] Major
- [X] Medium
- [ ] Minor

### DESCRIPTION OF ISSUE

Contract detail view has a lot of tabs: Transactions, Token Transfers, Tokens ERC-20, NFTs ERC-721, Code

When tabs don't fit on screen width they just disappear.

### DEVICE / VIEWPORT

- [ ] Desktop
- [X] Mobile

### STEPS TO REPRODUCE

1. open https://explorer.oasis.io/mainnet/sapphire/address/0x8Bc2B030b299964eEfb5e1e0b36991352E56D2D3/tokens/erc-721#tokens on small screen or mobile
2. try to see the bytecode tab

![small screen](https://github.com/oasisprotocol/explorer/assets/3758846/d3e05559-ce6d-4923-80c1-7c05dcf55bf1)

### ALTERNATIVES AND SUGGESTIONS

Use https://mui.com/material-ui/react-tabs/#scrollable-tabs

buberdds:
@donouwens here's MUI option enabled for tab scrolling

https://7c8907c1.oasis-explorer.pages.dev/mainnet/sapphire/address/0x8Bc2B030b299964eEfb5e1e0b36991352E56D2D3/tokens/erc-721#tokens

PR #1018, opened 2023-11-14
Enable scrollable tabs

buberdds:
Discussed options on slack with @donouwens. Closes https://github.com/oasisprotocol/explorer/issues/1006


src/app/components/RouterTabs/index.tsx
```diff
@@ -22,7 +22,7 @@ export function RouterTabs<Context>({ tabs, context }: RouterTabsProps<Context>)
 
   return (
     <>
-      <Tabs value={currentTab?.to}>
+      <Tabs value={currentTab?.to} variant="scrollable" scrollButtons={false}>
```
lukaw3d, line 25: Note: works on touch screens only

Issue #1037, opened 2023-11-23
[Bug]: Address formatting unexpectedly changing to Oasis


lubej:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### SUMMARY

Formatting not correct on account with 0 activity

### STEPS TO REPRODUCE

1. Open [https://explorer.oasis.io/testnet/sapphire/address/0x551b64b1747F6349259258FBFd1C03F4F782eC78](https://explorer.oasis.io/testnet/sapphire/address/0x551b64b1747F6349259258FBFd1C03F4F782eC78) in web browser
2. Even though address was provided in ETH format, it gets transformed to Oasis format in this case oasis1qz64azk6qr5nzq537l5cgpd4uxltrhpkhqdx5tqy

### EXPECTED BEHAVIOR

1. Explorer should respect the format given through param, and not change the formatting.

lubej:
In case of a address with activity it works as expected. 

Example: [https://explorer.oasis.io/testnet/sapphire/address/0x4768EF6e10cB229Cf7a3A99973Bac9020f0850b5](https://explorer.oasis.io/testnet/sapphire/address/0x4768EF6e10cB229Cf7a3A99973Bac9020f0850b5)

PR #1039, opened 2023-11-27
Fix address formatting on empty account

lubej:
Resolves https://github.com/oasisprotocol/explorer/issues/1037.

### DIFF
--- a/src/app/pages/AccountDetailsPage/AccountNFTCollectionCard.tsx
+++ b/src/app/pages/AccountDetailsPage/AccountNFTCollectionCard.tsx
@@ -31,9 +31,9 @@ export const accountNFTCollectionContainerId = 'nftCollection'
 
 export const AccountNFTCollectionCard: FC<AccountDetailsContext> = ({ scope, address }) => {
   const { t } = useTranslation()
-  const oasisContractAddress = useLoaderData() as string
+  const contractAddress = useLoaderData() as string
   const { inventory, isFetched, isLoading, isTotalCountClipped, pagination, totalCount } =
-    useAccountTokenInventory(scope, address, oasisContractAddress)
+    useAccountTokenInventory(scope, address, contractAddress)
   const firstToken = inventory?.length ? inventory?.[0].token : undefined
 
   return (
--- a/src/app/pages/TokenDashboardPage/hook.ts
+++ b/src/app/pages/TokenDashboardPage/hook.ts
@@ -182,11 +182,23 @@ export const useAccountTokenInventory = (scope: SearchScope, address: string, to
     throw AppErrors.UnsupportedLayer
     // There are no tokens on the consensus layer.
   }
-  const query = useGetRuntimeAccountsAddressNfts(network, layer, address, {
-    limit: NUMBER_OF_INVENTORY_ITEMS,
-    offset: offset,
-    token_address: tokenAddress,
-  })
+
+  const oasisAddress = useTransformToOasisAddress(address)
+  const query = useGetRuntimeAccountsAddressNfts(
+    network,
+    layer,
+    oasisAddress!,
+    {
+      limit: NUMBER_OF_INVENTORY_ITEMS,
+      offset: offset,
+      token_address: tokenAddress,
+    },
+    {
+      query: {
+        enabled: !!oasisAddress,
+      },
+    },
+  )
   const { isFetched, isLoading, data } = query
   const inventory = data?.data.evm_nfts
 
--- a/src/app/utils/route-utils.ts
+++ b/src/app/utils/route-utils.ts
@@ -147,13 +147,6 @@ export const addressParamLoader = ({ params }: LoaderFunctionArgs): string => {
   return params.address!
 }
 
-export const contractAddressParamLoader = async ({ params }: LoaderFunctionArgs) => {
-  validateAddressParam(params.contractAddress!)
-  // TODO: remove conversion when API supports querying by EVM address
-  const address = await getOasisAddress(params.contractAddress!)
-  return address
-}
-
 export const blockHeightParamLoader = async ({ params }: LoaderFunctionArgs) => {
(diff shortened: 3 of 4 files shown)


src/app/pages/AccountDetailsPage/AccountTokenTransfersCard.tsx
```diff
@@ -23,19 +23,17 @@ export const AccountTokenTransfersCard: FC<AccountDetailsContext> = ({ scope, ad
...
   )
 }
 
-const AccountTokenTransfers: FC<AccountDetailsContext> = ({ scope, address }) => {
+const AccountTokenTransfers: FC<AccountDetailsContext> = ({ scope, address, account }) => {
   const { t } = useTranslation()
-
   const { isLoading, isFetched, results } = useTokenTransfers(scope, address)
 
-  const { account } = useAccount(scope, address)
```
lubej, line 38: Not sure if this was by design, but these useAccount calls are duplicated, so I just moved the account to the OutletContext

src/oasis-nexus/api.ts
```diff
@@ -284,7 +304,7 @@ export const useGetRuntimeAccountsAddress: typeof generated.useGetRuntimeAccount
           if (status !== 200) return data
           return groupAccountTokenBalances({
             ...data,
-            address_eth: getEthAccountAddressFromPreimage(data.address_preimage),
+            address_eth: getEthAddressForAccount(data, searchParam),
```
lukaw3d, line 307: Do we expect future API to behave like this?

src/oasis-nexus/api.ts
```diff
@@ -267,12 +281,18 @@ export const useGetConsensusAccountsAddress: typeof generated.useGetConsensusAcc
...
+// TODO: remove conversion when API supports querying by EVM address
+type useGetRuntimeAccountsAddressExtended = (
+  searchParam: string | undefined,
+  ...rest: Parameters<typeof generated.useGetRuntimeAccountsAddress>
+) => UseQueryResult<
+  Awaited<ReturnType<typeof GetRuntimeAccountsAddress>>,
+  HumanReadableErrorResponse | NotFoundErrorResponse
+> & { queryKey: QueryKey }
+
+export const useGetRuntimeAccountsAddress: useGetRuntimeAccountsAddressExtended = (searchParam, ...rest) => {
```
lukaw3d, line 293: I think it would be much prettier if we can call useGetRuntimeAccountsAddress with address=oasis1|0x, and move address loader logic into this wrapper. Especially if API will behave like that in the future

lubej: I am not sure, based on this [TODO](https://github.com/oasisprotocol/explorer/blob/40bf1517148c0a60c82dd92207a66c6fa3d32f0b/src/app/utils/route-utils.ts#L139)(Written by you) I would expect that the API will support 0x addresses in the future.

But I agree, it would be much cleaner if I move this to the API wrapper call. It does affect the token address call as well. Will look into that.

src/app/pages/AccountDetailsPage/AccountTokenTransfersCard.tsx
lukaw3d: oh no some of these changes are worse. https://94637592.oasis-explorer.pages.dev/mainnet/sapphire/address/0x9BC6E51BCc4A23dfF5ede63ed35BB0cFFCB3468c/tokens/erc-20#tokens shows `This account holds no ERC-20 tokens` instead of a loading state

src/app/hooks/useTransformToOasisAddress.ts
```diff
@@ -0,0 +1,25 @@
...
+import { getOasisAddress } from '../utils/helpers'
+
+export const useTransformToOasisAddress = (ethOrOasisAddress: string): string | null => {
+  const [oasisAddress, setOasisAddress] = useState<string | null>(null)
+  const shouldUpdate = useRef(false)
+
+  useEffect(() => {
+    const transformToOasisAddress = async (addr: string) => {
+      const oasisAddr = await getOasisAddress(addr)
+      if (shouldUpdate.current) {
```
lukaw3d, line 11: Is there an edge case this is needed for?

lukaw3d (approved):

buberdds:
and it breaks nfts where fetch looks like 
`https://nexus.stg.oasis.io/v1/emerald/accounts/oasis1qz9vum3457ngpqe2utn22cun5p3qe3hpvvskc9rh/nfts?token_address=oasis1qz9vum3457ngpqe2utn22cun5p3qe3hpvvskc9rh`

instead of 
`https://nexus.stg.oasis.io/v1/emerald/accounts/oasis1qz9vum3457ngpqe2utn22cun5p3qe3hpvvskc9rh/nfts?token_address=oasis1qzk7lqppqv0nyvy22e2zdy3tqqpk2e3w2yjua6lq`

so instead of contract address account address is returned.

lubej:
> and it breaks nfts where fetch looks like `https://nexus.stg.oasis.io/v1/emerald/accounts/oasis1qz9vum3457ngpqe2utn22cun5p3qe3hpvvskc9rh/nfts?token_address=oasis1qz9vum3457ngpqe2utn22cun5p3qe3hpvvskc9rh`
> 
> instead of `https://nexus.stg.oasis.io/v1/emerald/accounts/oasis1qz9vum3457ngpqe2utn22cun5p3qe3hpvvskc9rh/nfts?token_address=oasis1qzk7lqppqv0nyvy22e2zdy3tqqpk2e3w2yjua6lq`
> 
> so instead of contract address account address is returned.



> This has a bug. Search on master is broken.
> 
> ```
> helpers.ts:43 Uncaught (in promise) Error: Invalid address
>     at b (helpers.ts:43:15)
>     at a (useTransformToOasisAddress.ts:11:31)
> ```

Will make a follow up PR. Thanks for noticing.

PR #1049, opened 2023-11-29
Add support for handling child tab routes

buberdds:
Alternate solution to https://github.com/oasisprotocol/explorer/pull/910

Keep router tab selected for a child route. Needed for NFT project.

### DIFF
--- a/src/app/components/RouterTabs/index.tsx
+++ b/src/app/components/RouterTabs/index.tsx
@@ -1,4 +1,4 @@
-import { useLocation, Outlet } from 'react-router-dom'
+import { useLocation, Outlet, useMatches } from 'react-router-dom'
 import { NonScrollingRouterLink } from '../NonScrollingRouterLink'
 import Tabs from '@mui/material/Tabs'
 import Tab from '@mui/material/Tab'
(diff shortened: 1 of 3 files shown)


src/app/components/RouterTabs/index.tsx
```diff
@@ -18,13 +18,21 @@ function getPathname(tab: { to: string }) {
 
 export function RouterTabs<Context>({ tabs, context }: RouterTabsProps<Context>) {
   const { pathname } = useLocation()
-  const currentTab = tabs.find(tab => getPathname(tab) === pathname)
+  let targetTab = tabs.find(tab => getPathname(tab) === pathname)
+  const matches = useMatches()
+
+  if (!targetTab) {
+    /// the last index is the current route, -2 is the first parent in route hierarchy
+    const parentPathname = matches?.at(-2)?.pathname
```
lubej, line 26: Haven't run this, but hardcoding this to 2 level deep seems like a wrong approach. Can we take the first child of `matches` as the top parent? To support "true" hierarchy?

src/app/components/RouterTabs/index.tsx
```diff
@@ -18,13 +18,21 @@ function getPathname(tab: { to: string }) {
 
 export function RouterTabs<Context>({ tabs, context }: RouterTabsProps<Context>) {
```
lukaw3d, line 19: Ideally add some tests, e.g. what tab is selected if tabs are
```tsx
tabs = [
  "/mainnet/sapphire/address/oasis1qry5d5zxs58w9deu5ara5zaamhkzqtg99s00ls90",
  "/mainnet/sapphire/address/oasis1qry5d5zxs58w9deu5ara5zaamhkzqtg99s00ls90/token-transfers#transfers",
  "/mainnet/sapphire/address/oasis1qry5d5zxs58w9deu5ara5zaamhkzqtg99s00ls90/tokens/erc-20#tokens",
  "/mainnet/sapphire/address/oasis1qry5d5zxs58w9deu5ara5zaamhkzqtg99s00ls90/tokens/erc-721#tokens",
  "/mainnet/sapphire/address/oasis1qry5d5zxs58w9deu5ara5zaamhkzqtg99s00ls90/code#code",
]
```

if location is:
/mainnet/sapphire/address/oasis1qry5d5zxs58w9deu5ara5zaamhkzqtg99s00ls90/tokens/erc-721
/mainnet/sapphire/address/oasis1qry5d5zxs58w9deu5ara5zaamhkzqtg99s00ls90/tokens/erc-721/token1#token1
/mainnet/sapphire/address/oasis1qry5d5zxs58w9deu5ara5zaamhkzqtg99s00ls90/other
/mainnet/sapphire/address/oasis1qry5d5zxs58w9deu5ara5zaamhkzqtg99s00ls90#transfers

lukaw3d (approved):

src/app/components/RouterTabs/index.tsx
buberdds: oh yeah I just need to handle a direct child routes - I've updated commit messages to be more specific. "The first child of `matches`" is root path '/' I am not sure how it can help. To handle any number of nested child routes we would need to do sth like: based on a current route, find an index in `matches` based on the closest `to` value in `tabs` array.

src/app/components/RouterTabs/index.tsx
buberdds: I will add them once I will wrap up other NFT UI work.

PR #1051, opened 2023-11-30
Mz/collection card clean up

buberdds:
- fix issue with showing pagination
- extract image gallery so it can be used in ERC-721 gallery


lubej (approved):
LGTM

PR #1061, opened 2023-12-04
Fix address formatting on empty account follow up

lubej:
- search broken due to useTransformToOasisAddress throwing on empty account
- NFTs query broken due to the wrong contract address supplied


src/app/hooks/useTransformToOasisAddress.ts
```diff
@@ -8,6 +8,10 @@ export const useTransformToOasisAddress = (ethOrOasisAddress: string): string |
     let shouldUpdate = true
 
     const transformToOasisAddress = async (addr: string) => {
+      if (!ethOrOasisAddress) {
```
lukaw3d, line 11: where is empty string coming from?

lubej: Search, this hook I believe `useRuntimeAccountConditionally`

lukaw3d: Ah, otherwise we would need to disable transformation until `options.query.enabled`

src/app/hooks/useTransformToOasisAddress.ts
```diff
@@ -8,6 +8,10 @@ export const useTransformToOasisAddress = (ethOrOasisAddress: string): string |
...
+        return
+      }
+
       const oasisAddr = await getOasisAddress(addr)
       if (shouldUpdate) {
         setOasisAddress(oasisAddr)
```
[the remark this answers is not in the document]
lubej: Yeah, good point.

lukaw3d (approved):

PR #1081, opened 2023-12-08
Implement searching for blocks by hash

csillag:


src/app/pages/SearchResultsPage/hooks.ts
```diff
@@ -134,16 +163,16 @@ export function useRuntimeTokenConditionally(
 
 export const useSearch = (q: SearchParams) => {
   const queries = {
-    blockHeight: useBlocksConditionally(q.blockHeight),
-    // TODO: searchQuery.blockHash when API is ready
+    blockHeight: useBlocksByHeightConditionally(q.blockHeight),
+    blockHash: useBlocksByHashConditionally(q.blockHash),
```
lukaw3d, line 167: self note: validation is not missing; block hash validation code is already there https://github.com/oasisprotocol/explorer/blob/9156cc22413840bd385601c75c78499c2ed70c47/src/app/components/Search/search-utils.ts#L114 https://github.com/oasisprotocol/explorer/blob/9156cc22413840bd385601c75c78499c2ed70c47/src/app/components/Search/search-utils.ts#L62

lukaw3d (approved):

PR #1098, opened 2023-12-14
Assemble changes for 1.4.0 release

buberdds:
To fully support NFTs we need to wait for reindexing, but based on Product meeting we want to deploy the latest code this sprint.

### DIFF
--- a/.changelog/1000.internal.md
+++ b/.changelog/1000.internal.md
@@ -1,1 +1,0 @@
-Update dependency @types/react to v18.2.35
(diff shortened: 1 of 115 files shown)


csillag (commented):
What happened to PR #1098?
(The one with the wrong number in the changelog fragment filename)

I dont see it here, with neither the correct nor the wrong number...

lubej (approved):

buberdds:
@csillag we don't care about it at the end, because you marked it as `trivial`. Trivial changes are ignored by towncrier when building changeset. If you want I can add it to internal changes if it is important.

PR #1129, opened 2024-01-04
Enable Consensus routing

buberdds:
Consensus mockups don't match most of our ParaTimes views so we prob need to setup separated views. To discuss how we want to handle routes

here we can keep dynamic segments for ParaTimes and we will be able to add specific Consensus (static segment) routes

```
          {
            path: 'consensus',
            errorElement: withDefaultTheme(<RoutingErrorPage />),
            loader: scopeLoader,
            children: [
              {
                path: '',
                element: <ConsensusDashboardPage />,
              },
            ],
          },
          // our current runtimes paths
```
Will match backend design (consensus - runtimes separation) more. We keep current Paratimes as is. ~W can rename some stuff like :layer to :runtime if we go with splats.~

possible routing options:
- dynamic segments for ParaTimes static for Consensus
- dynamic segments only 
- static segments
- (breaking change) re-define routes hierarchy (mainnet/runtime/:layer, mainnet/consensus)

### DIFF
--- a/src/app/hooks/useScopeParam.ts
+++ b/src/app/hooks/useScopeParam.ts
@@ -1,4 +1,4 @@
-import { useParams, useRouteError } from 'react-router-dom'
+import { useRouteLoaderData, useParams, useRouteError } from 'react-router-dom'
 import { Network } from '../../types/network'
 import { RouteUtils } from '../utils/route-utils'
 import { AppError, AppErrors } from '../../types/errors'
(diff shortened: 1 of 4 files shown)


src/routes.tsx
```diff
@@ -56,111 +56,116 @@ export const routes: RouteObject[] = [
...
-          {
-            path: `block/:blockHeight`,
-            element: <BlockDetailPage />,
-            loader: blockHeightParamLoader,
-          },
-          {
-            path: `address/:address`,
-            element: <AccountDetailsPage />,
-            loader: addressParamLoader(),
+            path: ':layer',
```
lukaw3d, line 63: I'm not sure we need splat to add `path: 'consensus'` later
https://lukaw3d.github.io/react-router-v6-searchable-docs/start/overview/#ranked-route-matching 

Another issue will be `useParams`. It won't return layer=consensus in that case.
One idea:
```diff
  path: ':layer',
+ loader: async ({ params }) => params,
+ id: 'scope1',

  path: 'consensus',
  loader: async ({ params }) => ({...params, layer: consensus}),
  id: 'scope2',

- useParams
+ useLoaderData('scope1') ?? useLoaderData('scope2')
```

lukaw3d (approved):

src/app/hooks/useScopeParam.ts
```diff
@@ -14,14 +14,23 @@ type ScopeInfo = SearchScope & {
   valid: boolean
 }
 
+type Scope = {
+  layer: Layer | undefined
+  network: Network | undefined
```
lukaw3d, line 19: I don't think these fields can be undefined. It would be caught by loaders. I think useScopeParam only needs to check that one of the useRouteLoaderData is found

lukaw3d: But that was also true previously, so was `.valid` not needed, and always true?

lukaw3d: or do we never use the errors thrown in loaders

lukaw3d: large refactor in https://github.com/oasisprotocol/explorer/pull/1159

buberdds: ah I can see the plan is to merge this one first.

PR #1134, opened 2024-01-05
Improve error messages

csillag:
Always try to properly differentiate between:
 - Illegal request (wrong address etc)
 - Failure to load data
 - Data not found

| Situation | Before | After |
| -------- | --- | --- |
| Can't load block details |  block_before_cant_load | block_after_cant_load 
| Can't load TX details | tx_before_cant_load  | tx_after_cant_load  |
| Can't load Token data | token_before_cant_load | token_after_cant_load |
| Token does not exist |  token_before_not_found | token_after_not_found |

### DIFF
--- a/src/app/pages/BlockDetailPage/index.tsx
+++ b/src/app/pages/BlockDetailPage/index.tsx
@@ -28,12 +28,21 @@ export const BlockDetailPage: FC = () => {
     // We should use useGetConsensusBlocksHeight()
   }
   const blockHeight = parseInt(useParams().blockHeight!, 10)
-  const { isLoading, data } = useGetRuntimeBlockByHeight(
+  const { isLoading, data, error } = useGetRuntimeBlockByHeight(
     scope.network,
     scope.layer, // This is OK, since consensus is already handled separately
     blockHeight,
   )
-  if (!data && !isLoading) {
+
+  if (!data?.data && !isLoading) {
+    const errorCode = (error as any)?.code
+    switch (errorCode) {
+      case 'ERR_NETWORK':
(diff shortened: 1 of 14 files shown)


src/locales/en/translation.json
```diff
@@ -152,6 +152,8 @@
     "value": "{{value, number}}"
   },
   "errors": {
+    "canNotLoadData": "Can not load data",
+    "canNotReachDataSource": "We were unable to reach the data source to request this data. This is probably a temporary (network?) problem. Please try again later.",
```
lukaw3d, line 156: > "(network?)"

should we avoid the word "network" and use connection/connectivity?

cc @donouwens

src/app/pages/BlockDetailPage/index.tsx
```diff
@@ -28,12 +28,21 @@ export const BlockDetailPage: FC = () => {
...
-  if (!data && !isLoading) {
+
+  if (!data?.data && !isLoading) {
+    const errorCode = (error as any)?.code
+    switch (errorCode) {
+      case 'ERR_NETWORK':
+        throw AppErrors.CannotLoadData
+      default:
+      // TODO: look for other error codes, too.
+      // (Currently we are not aware of anything else)
```
lukaw3d, line 44: Others: https://github.com/axios/axios/blob/6d4c421/README.md#error-types

ETIMEDOUT, ECONNABORTED, ERR_BAD_RESPONSE would indicate backend issue

src/locales/en/translation.json
csillag: Friendly reminder: this is waiting for input from @donouwens

PR #1138, opened 2024-01-09
Enable Consensus layer conditionally

buberdds:
- make Consensus layer enabled for non prod envs to unblock dev and access stage previews 
- (not sure about this) split between ENABLED_RUNTIMES and _CONSENSUS for now as we don't want to provide search suggestions and we will need Consensus specific components like ConsensusLearningMaterial

to make Consensus rendering happy we will need (https://github.com/oasisprotocol/explorer/pull/1133, and  routing changes for example https://github.com/oasisprotocol/explorer/pull/1129)

### DIFF
--- a/src/app/utils/route-utils.ts
+++ b/src/app/utils/route-utils.ts
@@ -5,15 +5,18 @@ import { AppError, AppErrors } from '../../types/errors'
 import { EvmTokenType, Layer } from '../../oasis-nexus/api'
 import { Network } from '../../types/network'
 import { SearchScope } from '../../types/searchScope'
+import { stableDeploys } from '../../config'
 
-export type SpecifiedPerEnabledLayer<T = any> = {
-  [N in keyof (typeof RouteUtils)['ENABLED_LAYERS_FOR_NETWORK']]: {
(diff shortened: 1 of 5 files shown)


lukaw3d (approved):

src/app/utils/route-utils.ts
```diff
@@ -5,15 +5,18 @@ import { AppError, AppErrors } from '../../types/errors'
 import { EvmTokenType, Layer } from '../../oasis-nexus/api'
 import { Network } from '../../types/network'
 import { SearchScope } from '../../types/searchScope'
+import { deploys } from '../../config'
 
-export type SpecifiedPerEnabledLayer<T = any> = {
-  [N in keyof (typeof RouteUtils)['ENABLED_LAYERS_FOR_NETWORK']]: {
-    [L in keyof (typeof RouteUtils)['ENABLED_LAYERS_FOR_NETWORK'][N]]: T
+export const isProduction = deploys.production.some(url => window.location.origin === url)
```
buberdds, line 10: exclude for staging too

PR #1145, opened 2024-01-11
Handle URLs with invalid networks

csillag:
We don't want to crash on URLs like https://explorer.oasis.io/abc/xyz.

This turns out to be surprisingly difficult to achieve, but here we go.

### DIFF
--- a/.changelog/1145.bugfix.md
+++ b/.changelog/1145.bugfix.md
@@ -1,0 +1,1 @@
+Handle URLs with invalid networks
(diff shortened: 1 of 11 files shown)


csillag:
I'm not quite satisfied with the semantics of `SearchScope` and `SearchScopeCandidate` (and the ambiguity of the missing `valid` flag), I'll try to improve this a little more..

csillag:
This has mostly been obsoleted by #1159, except the design-related parts. I'm closing this, and will open a new one with just those parts.

PR #1146, opened 2024-01-11
Support recovering from invalid URLs using search

csillag:
When launching a search from an invalid URL (like, invalid network or invalid layer), we should just jump to the global search page.

Implemented on top of #1145.


csillag:
This is no longer necessary, so closing.

PR #1147, opened 2024-01-11
Handle NFT image loading errors

buberdds:
- show no preview state based on image load error (we cannot rely on `image` prop send via API for now)
- will work with base64-encoded Data URLs 

updated:
Screenshot from 2024-01-11 13-25-58
Screenshot from 2024-01-11 13-26-09

master:
Screenshot from 2024-01-11 13-28-54

Screenshot from 2024-01-11 13-29-04

### DIFF
--- a/src/app/pages/NFTInstanceDashboardPage/InstanceImageCard.tsx
+++ b/src/app/pages/NFTInstanceDashboardPage/InstanceImageCard.tsx
@@ -5,6 +5,7 @@ import { Button } from '@mui/base/Button'
 import CardContent from '@mui/material/CardContent'
 import Card from '@mui/material/Card'
 import ContrastIcon from '@mui/icons-material/Contrast'
+import Link from '@mui/material/Link'
 import Skeleton from '@mui/material/Skeleton'
 import Tooltip from '@mui/material/Tooltip'
 import OpenInFullIcon from '@mui/icons-material/OpenInFull'
@@ -73,10 +74,12 @@ type InstanceImageCardProps = {
 }
 
 export const InstanceImageCard: FC<InstanceImageCardProps> = ({ isFetched, isLoading, nft }) => {
+  const { t } = useTranslation()
   const [darkMode, setDarkMode] = useState(false)
   const [previewOpen, setPreviewOpen] = useState(false)
   const handlePreviewOpen = () => setPreviewOpen(true)
   const handlePreviewClose = () => setPreviewOpen(false)
+  const [imageLoadError, setImageLoadError] = useState(false)
 
   return (
     <Card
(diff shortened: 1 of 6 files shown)


src/app/pages/NFTInstanceDashboardPage/InstanceImageCard.tsx
```diff
@@ -91,11 +94,28 @@ export const InstanceImageCard: FC<InstanceImageCardProps> = ({ isFetched, isLoa
...
+            <Box sx={{ flex: 1, display: 'flex', alignItems: 'center' }}>
+              <Box
+                sx={{
+                  display: 'flex',
+                  flexDirection: 'column',
+                  alignItems: 'center',
+                }}
+              >
+                <NoPreview placeholderSize={imageSize} />
+                <Link href={nft.image} rel="noopener noreferrer" target="_blank">
```
lukaw3d, line 112: This looks vulnerable to `image="javascript:alert(1)"`

buberdds: Can our backend can send this instead of ipfs url? Anyway, you mean to show Link only when isNftImageUrlValid=true, right?

lukaw3d: I'm not sure if backend allows it, but I would prefer to keep validProtocols on frontend too

lukaw3d: Luckily react says `Warning: A future version of React will block javascript: URLs as a security precaution.` otherwise I might add
```tsx
const originalSetAttribute = Element.prototype.setAttribute
Element.prototype.setAttribute = function (...args) {
  const [name, value] = args
  if ((name === 'href' || name === 'src') && !isNftImageUrlValid(value)) {
    console.error('blocked suspicious setAttribute', src, value)
    return
  }
  return originalSetAttribute.apply(this, args)
}
```

lukaw3d: related issues in wallet from validator metadata:
https://github.com/oasisprotocol/oasis-wallet-web/pull/199
https://github.com/oasisprotocol/oasis-wallet-web/issues/185

lukaw3d (approved):

src/app/pages/NFTInstanceDashboardPage/InstanceImageCard.tsx
buberdds: Ok for now I will merge as is, because there is `isNftImageUrlValid` condition that prevents rendering a link.

PR #1165, opened 2024-01-20
If user clicks Sourcify show a note there's a delay after verifying

lukaw3d:
Thoughts?

note-sourcify

### DIFF
--- a/src/app/components/ContractVerificationIcon/index.tsx
+++ b/src/app/components/ContractVerificationIcon/index.tsx
@@ -1,4 +1,4 @@
-import { FC, ReactNode } from 'react'
+import { FC, ReactNode, useState } from 'react'
 import { Trans, useTranslation } from 'react-i18next'
 import Box from '@mui/material/Box'
 import CheckCircleIcon from '@mui/icons-material/CheckCircle'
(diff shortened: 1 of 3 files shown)


src/app/components/ContractVerificationIcon/index.tsx
```diff
@@ -105,10 +107,12 @@ export const VerificationIcon: FC<{ address_eth: string; verified: boolean; noLi
                   rel="noopener noreferrer"
                   target="_blank"
                   sx={{ fontWeight: 400, color: 'inherit', textDecoration: 'underline' }}
+                  onClick={verified ? undefined : () => setExplainDelay(true)}
                 />
               ),
             }}
-          />
+          />{' '}
+          {explainDelay && t('contract.verification.explainVerificationDelay')}
```
buberdds, line 115: I would just show additional description always for not verified contract in details page. Not sure sure users will catch different copy when they switch context to another tab.

buberdds (approved):

src/app/components/ContractVerificationIcon/index.tsx
lukaw3d: yeah :/ but it doesn't apply 99.99% of the time

lukaw3d: Maybe if worded like: "Verify through Sourcify and give it a minute."

PR #1176, opened 2024-01-24
Use proper checksum ETH address format in Runtime events

### LINKED ISSUE #1141: [Bug]: Addresses in EVM events are not displayed with capitalization checksum

- [X] I have searched the existing issues

### SUMMARY

_No response_

### STEPS TO REPRODUCE

- https://explorer.dev.oasis.io/testnet/sapphire/address/0x000000000000000000000000000000000000dEaD
- image

### EXPECTED BEHAVIOR

0x000000000000000000000000000000000000dEaD

csillag:
Closes #1141.

### DIFF
--- a/src/oasis-nexus/api.ts
+++ b/src/oasis-nexus/api.ts
@@ -21,6 +21,7 @@ import { getTickerForNetwork, NativeTicker } from '../types/ticker'
 import { useTransformToOasisAddress } from '../app/hooks/useTransformToOasisAddress'
 import { useEffect, useState } from 'react'
 import { RpcUtils } from '../app/utils/rpc-utils'
+import { toChecksumAddress } from '@ethereumjs/util'
 
 export * from './generated/api'
 export type { RuntimeEvmBalance as Token } from './generated/api'
@@ -685,6 +686,14 @@ export const useGetRuntimeEvmTokensAddress: typeof generated.useGetRuntimeEvmTok
   })
 }
 
+const fixChecksumAddressInEvmEventParam = (param: generated.EvmEventParam): generated.EvmEventParam =>
+  param.evm_type === 'address'
+    ? {
+        ...param,
+        value: toChecksumAddress(param.value as string),
+      }
+    : param
+
 export const useGetRuntimeEvents: typeof generated.useGetRuntimeEvents = (
   network,
   runtime,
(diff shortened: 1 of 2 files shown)


src/app/components/Account/AccountLink.tsx
```diff
@@ -7,6 +7,8 @@ import { RouteUtils } from '../../utils/route-utils'
 import Typography from '@mui/material/Typography'
 import { COLORS } from '../../../styles/theme/colors'
 import { SearchScope } from '../../../types/searchScope'
+import { isValidEthAddress } from '../../utils/helpers'
+import { toChecksumAddress } from '@ethereumjs/util'
 
 export const AccountLink: FC<{
```
lukaw3d, line 13: Copy To Clipboard still copies non-checksum address.
And I don't think CopyToClipboard component should be doing transformations, so the best place to convert is in useGetRuntimeEvents, but it's also the hardest; event body is `any` and different event types have different fields.

csillag: > Copy To Clipboard still copies non-checksum address.

Indeed. Do we want it something else? I assumed we want the checksum format for displaying only.

> I don't think CopyToClipboard component should be doing transformations,

Nor do I, that's why I added it to the `AccountLink` component.

>  so the best place to convert is in useGetRuntimeEvents,

I'm not sure about that. We are using `AccountLink` in many other places, too, where it's not related to runtime event. To me, `AccountLink` seems to be the natural choice.

> but it's also the hardest; event body is any and different event types have different fields.

When formatting runtime events, when we know that the type is address, we always use `AccountLink` for displaying, so adding it here seems to solve the `any` problem as well as we currently can.

lukaw3d: I think it should copy checksum format too. API returns checksum format everywhere else, so display and copy of e.g. addresses in Transaction Details are both in checksum format. Same in list of Tokens.

csillag: >  the best place to convert is in useGetRuntimeEvents

OK, moved tra transformation into `useGetRuntimeEvents`.

src/oasis-nexus/api.ts
```diff
@@ -714,6 +723,7 @@ export const useGetRuntimeEvents: typeof generated.useGetRuntimeEvents = (
               ) {
                 return {
                   ...event,
+                  evm_log_params: event.evm_log_params?.map(fixChecksumAddressInEvmEventParam),
```
lukaw3d, line 726: oh, it's this easy. No need to check event types because 0x addresses only appear in evm.log params. Not in e.g. `body` of consensus_accounts.deposit

lukaw3d (approved):

src/oasis-nexus/api.ts
csillag: Indeed.

PR #1177, opened 2024-01-24
Update active nodes number

buberdds:
Update fallback values as staging API returns 0 (issues with Consensus data).


csillag (approved):
Reviewed, LGTM.

(Maybe it would make sense to describe the source if this data in the commit message?)

PR #1178, opened 2024-01-24
In block details, add prev and next buttons 

csillag:
As requested on Slack by @kostko 
Design is [here]

The implementation looks like this:
image

### DIFF
--- a/src/app/components/BlockNavigationButtons/index.tsx
+++ b/src/app/components/BlockNavigationButtons/index.tsx
@@ -1,0 +1,57 @@
+import { FC } from 'react'
+import { Link as RouterLink } from 'react-router-dom'
+import { useTranslation } from 'react-i18next'
+import PaginationItem from '@mui/material/PaginationItem'
+import Box from '@mui/material/Box'
+import Tooltip from '@mui/material/Tooltip'
+import { SearchScope } from '../../../types/searchScope'
+import { COLORS } from '../../../styles/theme/colors'
+import { useRuntimeFreshness } from '../OfflineBanner/hook'
+import { RouteUtils } from '../../utils/route-utils'
+
+export const PrevBlockButton: FC<{ scope: SearchScope; currentRound: number }> = ({
+  scope,
+  currentRound,
+}) => {
+  const { t } = useTranslation()
+  const disabled = currentRound === 0
+  return (
+    <Tooltip title={disabled ? t('blocks.viewingFirst') : t('blocks.viewPrevious')} placement="top">
+      <Box>
+        <PaginationItem
+          component={RouterLink}
+          to={RouteUtils.getBlockRoute(scope, currentRound - 1)}
+          type="previous"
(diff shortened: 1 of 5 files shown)


csillag:
Friendly reminder: this is waiting for input from @donouwens

csillag:
Now this has the blessing of @donouwens , so ready for review!

.env
```diff
@@ -2,7 +2,7 @@ REACT_APP_BUILD_DATETIME=0
 REACT_APP_BUILD_SHA=sha0000000000000000000000000000000000000
 REACT_APP_BUILD_VERSION=
 # REACT_APP_API=http://localhost:8008/v1/
-REACT_APP_API=https://nexus.stg.oasis.io/v1/
-REACT_APP_TESTNET_API=https://testnet.nexus.stg.oasis.io/v1/
-# REACT_APP_API=https://nexus.oasis.io/v1/
-# REACT_APP_TESTNET_API=https://testnet.nexus.oasis.io/v1/
+# REACT_APP_API=https://nexus.stg.oasis.io/v1/
```
buberdds, line 5: revert ?

src/app/pages/BlockDetailPage/index.tsx
```diff
@@ -53,6 +60,47 @@ export type BlockDetailRuntimeBlock = RuntimeBlock & {
   markAsNew?: boolean
 }
 
+const PrevBlockButton: FC<{ scope: SearchScope; currentRound: number }> = ({ scope, currentRound }) => {
```
buberdds, line 63: optional: If we plan to use it in Consensus it's worth to put this into some common components folder.

.env
csillag: It's in a separate commit, which I intend to drop before merging.

buberdds: ahh Ok

buberdds (approved):

src/app/components/BlockNavigationButtons/index.tsx
```diff
@@ -0,0 +1,57 @@
...
+  )
+}
+
+export const NextBlockButton: FC<{ scope: SearchScope; currentRound: number }> = ({
+  scope,
+  currentRound,
+}) => {
+  const { latestBlock } = useRuntimeFreshness(scope)
+  const { t } = useTranslation()
+  const disabled = latestBlock === currentRound
```
lukaw3d, line 43: I would have allowed users to go into 404 rather than have an extra request :D

csillag: My understanding is that we are checking the status for every page render anyway, to see if we need to display a banner or something. Results are cached, so checking it again (during the same page render) doesn't actually result in an extra request.

csillag: See https://github.com/oasisprotocol/explorer/pull/1201 for this improvement (the "live" version of the next button)

PR #1189, opened 2024-01-27
Handle new transfer event types

csillag:
We see them in the data now, for example [here](https://explorer.dev.oasis.io/mainnet/sapphire/tx/0x6990c0f6d4178090e07c9d48fc38fcb987e5daed54a940fa17a057922f17c344):

image

After the fix:

image

### DIFF
--- a/.changelog/1189.trivial.md
+++ b/.changelog/1189.trivial.md
@@ -1,0 +1,1 @@
+Handle new transfer event types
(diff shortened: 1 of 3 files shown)


csillag:
Waiting for a full list of newly inroduced values from @Andrew7234.

lukaw3d:
these don't look like transfer events. looks like user-defined evm events

Andrew7234:
> these don't look like transfer events. looks like user-defined evm events

Yeah they are user-defined.

Issue #226, opened 2023-03-17
[Feature]: Show a message or error where we don't support pagination, but number of items reaches the limit


lukaw3d:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### DESCRIPTION

examples:
- https://github.com/oasisprotocol/explorer/blob/1c1563e5669e539b8bee85d900e25e91a4d1cd00/src/app/components/Transactions/Logs.tsx#L17
- https://github.com/oasisprotocol/explorer/blob/577dd973efc7a30e02ab445aa8c612f2b90f82f4/src/oasis-indexer/generated/api.ts#L590-L593

lukaw3d:
PontusX might hit these limits

lukaw3d:
https://explorer.dev.oasis.io/mainnet/emerald/address/0x6Cb9750a92643382e020eA9a170AbB83Df05F30B
events

lukaw3d:
https://github.com/oasisprotocol/explorer/pull/1398 isPotentiallyIncomplete / MAX_LOADED_ADDRESSES

lukaw3d:
https://github.com/oasisprotocol/explorer/pull/1673

PR #1192, opened 2024-01-27
Add support for searching for network proposal by name

csillag:
The searching happens on the client side, since Nexus has no support for it.

### DIFF
--- a/src/app/components/Search/search-utils.ts
+++ b/src/app/components/Search/search-utils.ts
@@ -100,6 +100,11 @@ export const validateAndNormalize = {
       return searchTerm.toLowerCase()
     }
   },
+  networkProposalNameFragment: (searchTerm: string) => {
+    if (searchTerm?.length >= textSearchMininumLength) {
+      return searchTerm.toLowerCase()
+    }
+  },
 } satisfies { [name: string]: (searchTerm: string) => string | undefined }
 
 export function isSearchValid(searchTerm: string) {
--- a/src/app/pages/SearchResultsPage/hooks.ts
+++ b/src/app/pages/SearchResultsPage/hooks.ts
@@ -13,6 +13,8 @@ import {
   EvmTokenList,
   EvmToken,
   useGetRuntimeBlockByHash,
+  Proposal,
+  useGetConsensusProposalsByName,
 } from '../../../oasis-nexus/api'
 import { RouteUtils } from '../../utils/route-utils'
 import { SearchParams } from '../../components/Search/search-utils'
@@ -24,7 +26,7 @@ function isDefined<T>(item: T): item is NonNullable<T> {
 export type ConditionalResults<T> = { isLoading: boolean; results: T[] }
 
 type SearchResultItemCore = HasScope & {
-  resultType: 'block' | 'transaction' | 'account' | 'contract' | 'token'
+  resultType: 'block' | 'transaction' | 'account' | 'contract' | 'token' | 'proposal'
 }
 
 export type BlockResult = SearchResultItemCore & RuntimeBlock & { resultType: 'block' }
@@ -37,7 +39,15 @@ export type ContractResult = SearchResultItemCore & RuntimeAccount & { resultTyp
 
 export type TokenResult = SearchResultItemCore & EvmToken & { resultType: 'token' }
 
-export type SearchResultItem = BlockResult | TransactionResult | AccountResult | ContractResult | TokenResult
+export type ProposalResult = SearchResultItemCore & Proposal & { resultType: 'proposal' }
+
+export type SearchResultItem =
+  | BlockResult
+  | TransactionResult
+  | AccountResult
+  | ContractResult
+  | TokenResult
+  | ProposalResult
 
 export type SearchResults = SearchResultItem[]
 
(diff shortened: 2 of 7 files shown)


csillag:
We decided to create the proposal details page first, so that we can link to them. When it's ready, I will revisit this one.

csillag:
> We decided to create the proposal details page first, so that we can link to them. When it's ready, I will revisit this one.

Proposal details page coming at #1202. So This PR is now waiting for that one.

src/app/components/Search/search-utils.ts
```diff
@@ -129,3 +135,8 @@ export const useParamSearch = () => {
 }
 
 export type SearchParams = ReturnType<typeof useParamSearch>
+
+export const getProposalTextMatcherFor = (textFragment: string | undefined) =>
+  textFragment
+    ? (proposal: Proposal): boolean => !!proposal.handler && proposal.handler?.indexOf(textFragment) !== -1
```
lukaw3d, line 141: `.includes(` is more readable

src/app/pages/SearchResultsPage/ResultsGroupByType.tsx
```diff
@@ -47,17 +47,22 @@ export function ResultsGroupByType<T>({ title, results, resultComponent, link, l
...
+        return (
+          <div key={i}>
+            {resultComponent(item)}
+            {!!itemLink && (
+              <Box sx={{ display: 'flex', justifyContent: 'center', mt: 5 }}>
+                <ViewResultButton component={RouterLink} to={itemLink}>
+                  {linkLabel}
+                </ViewResultButton>
+              </Box>
+            )}
```
lukaw3d, line 61: I would revert this, link to nothing, and merge

src/oasis-nexus/api.ts
```diff
@@ -811,6 +815,7 @@ export const useGetConsensusProposals: typeof generated.useGetConsensusProposals
             proposals: data.proposals.map(proposal => {
               return {
                 ...proposal,
+                network,
```
lukaw3d, line 818: I would expect `layer: 'consensus'`. How does `isItemInScope` work otherwise?
and how do SearchResultItemCore types allow this :scream:

src/app/pages/SearchResultsPage/hooks.ts
```diff
@@ -182,6 +215,10 @@ export const useSearch = (q: SearchParams) => {
     .map(l => l.evm_tokens)
     .flat()
     .sort((t1, t2) => t2.num_holders - t1.num_holders)
+  const proposals = queries.proposals.results
+    .map(l => l.proposals)
+    .flat()
+    .filter(getProposalTextMatcherFor(q.networkProposalNameFragment))
```
lukaw3d, line 221: Filtering should happen inside useNetworkProposalsConditionally, or even create useGetConsensusProposalsByName

src/app/components/Search/search-utils.ts
csillag: OK done

src/app/pages/SearchResultsPage/ResultsGroupByType.tsx
csillag: Reverted. (But there won't be a need to "link to nothing" for long, since the proposal details page is ready for review.)

src/oasis-nexus/api.ts
csillag: I am adding `layer` in the search hooks, here: https://github.com/oasisprotocol/explorer/pull/1192/files#diff-b129d850541c740e998adaf540687f07f1f6b8012922be706f48efea08bc36a5R236

src/app/pages/SearchResultsPage/hooks.ts
csillag: > Filtering should happen inside useNetworkProposalsConditionally, or even create useGetConsensusProposalsByName

OK, created `useGetConsensusProposalsByName()`.

src/oasis-nexus/api.ts
csillag: Never mind, moved it here, for easier reading.

csillag:
Rebased on top of the merged proposal details page; so we can not link and redirect to proposals from search results.

lukaw3d (approved):

src/oasis-nexus/api.ts
```diff
@@ -854,6 +855,23 @@ export const useGetConsensusProposalsProposalId: typeof generated.useGetConsensu
   })
 }
 
+export const useGetConsensusProposalsByName = (network: Network, nameFragment: string | undefined) => {
+  const query = useGetConsensusProposals(network, {}, { query: { enabled: !!nameFragment } })
```
lukaw3d, line 859: (related: https://github.com/oasisprotocol/explorer/issues/226)

buberdds:
Are we Ok with releasing this on prod when Consensus layer is not enabled?

csillag:
> Are we Ok with releasing this on prod when Consensus layer is not enabled? Should we disabled this type of search results?

That shouldn't be an issue, since we are launching the search for proposals like this: 
https://github.com/oasisprotocol/explorer/blob/dfbdc42b21418a2d5ddd359a5e33a2bd63ca9761/src/app/pages/SearchResultsPage/hooks.ts#L177-L182

In other words, we will only search networks where the consensus layer is enabled.

PR #1193, opened 2024-01-27
Proposal name search with highlight demo

csillag:
This is a demo branch for #1192 and #646 together.


csillag:
We have discussed everything we needed to discuss, so we no longer need this demo. Closing. (Will continue to work on the two related PRs separately.)

PR #1197, opened 2024-01-29
Consensus validators list

buberdds:
Does not include stats section. I am not 100% sure it will be part of MVP.

### DIFF
--- a/src/app/components/VerticalList/index.tsx
+++ b/src/app/components/VerticalList/index.tsx
@@ -1,0 +1,10 @@
+import { COLORS } from 'styles/theme/colors'
+import { styled } from '@mui/material/styles'
+import Box from '@mui/material/Box'
+
+export const VerticalList = styled(Box)(({ theme }) => ({
+  display: 'flex',
+  flexDirection: 'column',
+  gap: `0 ${theme.spacing(2)}`,
+  backgroundColor: COLORS.brandDark,
+}))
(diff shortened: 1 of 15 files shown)


src/app/components/VerticalList/index.tsx
```diff
@@ -0,0 +1,10 @@
+import { COLORS } from 'styles/theme/colors'
+import { styled } from '@mui/material/styles'
+import Box from '@mui/material/Box'
+
+export const VerticalList = styled(Box)(({ theme }) => ({
```
csillag, line 5: Great, this could also be used for `ProposalDetailView`. (It's already on master branch, you'll see it if you rebase this.)

csillag (approved):
Reviewed. LGTM.

Needs a rebase.
Too bad there are no actual validator images to test.

src/app/components/VerticalList/index.tsx
buberdds: I will enable proposal list in a next PR. For now we show that list in Dashboard card only

src/routes.tsx
```diff
@@ -79,6 +81,15 @@ export const routes: RouteObject[] = [
...
             loader: proposalIdParamLoader,
           },
+          {
+            path: `validators`,
+            element: <ValidatorsPage />,
+          },
+          {
+            path: `validators/:address`,
+            element: <ValidatorDetailsPage />,
+            loader: addressParamLoader(),
```
lukaw3d, line 91: this allows hex addresses

buberdds: will remove this line and add TODO to add it later when API will use address as param. Turns out it is using entity_id which is not what we want in URL.

PR #1201, opened 2024-01-29
Improve block details next button

csillag:
When displaying the latest block details, keep polling.
The idea is that the "View next block" button should be enabled as soon as a new block is available.

Please note that
 - No polling happens when we are displaying any other block besides the latest
 - As soon as we detect that the currently displayed block is no longer the latest block, and so the "view next block" button can be enabled, the polling stops.

### DIFF
--- a/src/app/components/BlockNavigationButtons/index.tsx
+++ b/src/app/components/BlockNavigationButtons/index.tsx
@@ -40,7 +40,9 @@ export const NextBlockButton: FC<{ scope: SearchScope; currentRound: number }> =
 }) => {
   const { latestBlock } = useRuntimeFreshness(scope)
   const { t } = useTranslation()
-  const disabled = latestBlock === currentRound
+  const disabled = !!latestBlock && currentRound >= latestBlock
+  // If the next button is disabled, we want to poll the freshness info, because it will probably be enabled in a few secs
+  useRuntimeFreshness(scope, { polling: disabled }) // This will trigger refreshing to see if there isa new block
   return (
     <Tooltip title={disabled ? t('blocks.viewingLatest') : t('blocks.viewNext')} placement="top">
       <Box>
--- a/src/app/components/BlockNavigationButtons/index.tsx
+++ b/src/app/components/BlockNavigationButtons/index.tsx
@@ -48,7 +48,7 @@ export const NextBlockButton: FC<{ scope: SearchScope; currentRound: number }> =
           component={RouterLink}
           to={RouteUtils.getBlockRoute(scope, currentRound + 1)}
           type="next"
-          disabled={latestBlock === currentRound}
+          disabled={disabled}
           sx={{ background: COLORS.grayMediumLight }}
         />
       </Box>
--- a/src/app/components/OfflineBanner/hook.ts
+++ b/src/app/components/OfflineBanner/hook.ts
@@ -28,12 +28,17 @@ export const useConsensusFreshness = (network: Network) => {
   }
 }
 
-export const useRuntimeFreshness = (scope: SearchScope): FreshnessInfo => {
+export const useRuntimeFreshness = (
+  scope: SearchScope,
+  queryParams: { polling?: boolean } = {},
+): FreshnessInfo => {
   const isApiReachable = useIsApiReachable(scope.network).reachable
   if (scope.layer === Layer.consensus) {
     throw new AppError(AppErrors.UnsupportedLayer)
   }
-  const query = useGetRuntimeStatus(scope.network, scope.layer)
+  const query = useGetRuntimeStatus(scope.network, scope.layer, {
+    query: { refetchInterval: queryParams.polling ? 8000 : undefined },
+  })
   const data = query.data?.data
   const lastUpdate = useFormattedTimestampStringWithDistance(data?.latest_block_time)
   const latestBlock = data?.latest_block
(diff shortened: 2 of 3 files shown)


src/app/components/BlockNavigationButtons/index.tsx
```diff
@@ -44,11 +45,13 @@ export const NextBlockButton: FC<{ scope: SearchScope; currentRound: number }> =
   return (
     <Tooltip title={disabled ? t('blocks.viewingLatest') : t('blocks.viewNext')} placement="top">
       <Box>
+        {/*If the next button is disabled, we want to poll the freshness info, because it will probably be enabled in a few secs*/}
+        {disabled && <RuntimeFreshnessTracker scope={scope} />}
```
lukaw3d, line 49: Home come this workaround is needed, instead of a few lines above:
```diff
- const { latestBlock } = useRuntimeFreshness(scope)
+ const { latestBlock } = useRuntimeFreshness(scope, true)
```

lukaw3d: i mean `const { latestBlock } = useRuntimeFreshness(scope, disabled)`

csillag: Circular definition. Disabled is determined based on runtime freshness. So we can't use disabled as a parameter of determining runtime freshness...

lukaw3d: Then maybe
```ts
const staleStatus = useRuntimeFreshness(scope)
const { latestBlock } = useRuntimeFreshness(scope, currentRound >= staleStatus.latestBlock)
```

src/app/components/OfflineBanner/hook.ts
```diff
@@ -28,12 +28,14 @@ export const useConsensusFreshness = (network: Network) => {
   }
 }
 
-export const useRuntimeFreshness = (scope: SearchScope): FreshnessInfo => {
+export const useRuntimeFreshness = (scope: SearchScope, polling = false): FreshnessInfo => {
```
lukaw3d, line 31: I think `polling` should be a named parameter

src/app/components/BlockNavigationButtons/index.tsx
```diff
@@ -44,11 +45,13 @@ export const NextBlockButton: FC<{ scope: SearchScope; currentRound: number }> =
   return (
```
lukaw3d, line 45: `const disabled = currentRound >= latestBlock`
depending on react-query config, it could return slightly stale runtime status result

src/app/components/BlockNavigationButtons/index.tsx
lukaw3d: hmmm, but staleStatus would not even be stale, if your workaround works

csillag: > hmmm, but staleStatus would not even be stale, if your workaround works

Indeed, the data returned by the first call should be sufficient; at the second time, we just need to start the polling. So in the end I went with this:

```
  const disabled = latestBlock === currentRound
  useRuntimeFreshness(scope, disabled)
```

I propped the component which used to do the same.

src/app/components/OfflineBanner/hook.ts
csillag: Done

src/app/components/BlockNavigationButtons/index.tsx
csillag: OK, done

src/app/components/OfflineBanner/hook.ts
lukaw3d: I was thinking `(scope: SearchScope, queryParams: { polling?: boolean } = {}): FreshnessInfo`

src/app/components/OfflineBanner/hook.ts
```diff
@@ -28,12 +28,14 @@ export const useConsensusFreshness = (network: Network) => {
...
 
-export const useRuntimeFreshness = (scope: SearchScope): FreshnessInfo => {
+export const useRuntimeFreshness = (scope: SearchScope & { polling?: boolean }): FreshnessInfo => {
   const isApiReachable = useIsApiReachable(scope.network).reachable
   if (scope.layer === Layer.consensus) {
     throw new AppError(AppErrors.UnsupportedLayer)
   }
-  const query = useGetRuntimeStatus(scope.network, scope.layer)
+  const query = useGetRuntimeStatus(scope.network, scope.layer, {
+    query: { refetchInterval: scope.polling ? 5000 : undefined },
```
lukaw3d, line 37: note, average block time is 6.2 seconds

lukaw3d (approved):

src/app/components/BlockNavigationButtons/index.tsx
```diff
@@ -40,15 +40,17 @@ export const NextBlockButton: FC<{ scope: SearchScope; currentRound: number }> =
 }) => {
   const { latestBlock } = useRuntimeFreshness(scope)
   const { t } = useTranslation()
-  const disabled = latestBlock === currentRound
+  const disabled = !!latestBlock && currentRound >= latestBlock
+  // If the next button is disabled, we want to poll the freshness info, because it will probably be enabled in a few secs
+  useRuntimeFreshness({ ...scope, polling: disabled })
```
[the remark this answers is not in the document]
csillag: OK

src/app/components/OfflineBanner/hook.ts
csillag: Yes, I changed it so before merging.

PR #1202, opened 2024-01-30
Add basic proposal details page

csillag:
Design is [here](https://www.figma.com/file/8dwSyYOOm3vlgbvxPGT29p/Block-Explorer-(Post-Launch)?type=design&node-id=8960-67644&mode=design&t=ryRSEAVnWV2RwqbZ-0).

This PR only handles the routing, the various error corner cases, and the main card.

Also, there is some data that is part of the design, but not yet available for Nexus. We will add that later.

Also, link to the proposal detail pages from the proposal list.


src/app/pages/ProposalDetailsPage/index.tsx
```diff
@@ -0,0 +1,99 @@
...
+export const ProposalDetailView: FC<{
+  proposal: Proposal
+  isLoading?: boolean
+  showLayer?: boolean
+  standalone?: boolean
+}> = ({ proposal, isLoading, showLayer = false, standalone = false }) => {
+  const { t } = useTranslation()
+  const { isMobile } = useScreenSize()
+  if (isLoading) return <TextSkeleton numberOfRows={7} />
+  console.log('loaded proposal', proposal)
```
lukaw3d, line 48: TODO?

src/oasis-nexus/api.ts
```diff
@@ -822,6 +827,31 @@ export const useGetConsensusProposals: typeof generated.useGetConsensusProposals
...
+    ...options,
+    request: {
+      ...options?.request,
+      transformResponse: [
+        ...arrayify(axios.defaults.transformResponse),
+        (data: generated.Proposal, headers, status) => {
+          if (status !== 200) return data
+          return {
+            ...data,
+            network,
```
lukaw3d, line 845: https://github.com/oasisprotocol/explorer/pull/1192#discussion_r1470516266

lukaw3d (approved):

src/app/pages/ProposalDetailsPage/index.tsx
csillag: Removed

src/oasis-nexus/api.ts
csillag: Yes, adding the consensus layer here now.

PR #1205, opened 2024-01-30
Update a comment in NextBlockButton

lukaw3d:
Fixes https://github.com/oasisprotocol/explorer/pull/1201#discussion_r1471330411


csillag (approved):
It might be a little more clear, yes.

PR #1214, opened 2024-01-31
Display the votes on the network proposal details page

csillag:
Display votes on the proposal details page.

Design is [here](https://www.figma.com/file/8dwSyYOOm3vlgbvxPGT29p/Block-Explorer-(Post-Launch)?node-id=6092%3A212988).

Some of the required data is not yet provided by nexus, including:
 - Timestamps for votes
 - Voting power accompanying the votes

Therefore, some parts are not implemented yet.

PR #1218, opened 2024-02-01
Proposal votes filter demo

csillag:
This branch is a combination of #646 and #1214, just for demo. No need to review.


csillag:
Since the highlighting code has been now merged, there is no need for a separate branch to combine the features for a demo, hence, closing.

PR #1219, opened 2024-02-01
Hide next/prev block buttons on mobile lists and search results

buberdds:
Closes https://github.com/oasisprotocol/explorer/pull/1221

### DIFF
--- a/src/app/pages/BlockDetailPage/index.tsx
+++ b/src/app/pages/BlockDetailPage/index.tsx
@@ -42,7 +42,7 @@ export const BlockDetailPage: FC = () => {
   return (
     <PageLayout>
       <SubPageCard featured title={t('common.block')}>
-        <BlockDetailView isLoading={isLoading} block={block} />
+        <BlockDetailView enableBlockNavigation isLoading={isLoading} block={block} />
       </SubPageCard>
       {!!block?.num_transactions && <TransactionsCard scope={scope} blockHeight={blockHeight} />}
       <EventsCard scope={scope} blockHeight={blockHeight} />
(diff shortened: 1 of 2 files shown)


src/app/pages/BlockDetailPage/index.tsx
```diff
@@ -89,8 +89,12 @@ export const BlockDetailView: FC<{
       <dd>
         <BlockLink scope={block} height={block.round} />
         <CopyToClipboard value={block.round.toString()} />
-        <PrevBlockButton scope={block} currentRound={block.round} />
-        <NextBlockButton scope={block} currentRound={block.round} />
+        {!standalone && (
```
lukaw3d, line 92: Issue: https://5603e983.oasis-explorer.pages.dev/search?q=2%2C265%2C700
still appears on search results too

lukaw3d: (sidenote, i'm confused what standalone means.)

lukaw3d: I would expect standalone to mean single block view, not the opposite https://5603e983.oasis-explorer.pages.dev/mainnet/sapphire/block/2265700

buberdds: hard to say what it means, but I think naming does not matter now. Let's fix buttons and imo let's refactor vertical tables to rely only on css. This way we will get rid of that prop (unless it is used in more contexts that I am  not aware of)

buberdds: added `enableBlockNavigation` prop

lukaw3d (approved):

PR #1230, opened 2024-02-05
Use correct, custom formatter in Age component

buberdds:
fixes https://github.com/oasisprotocol/explorer/pull/1194

### DIFF
--- a/src/app/components/Age/__tests__/index.test.tsx
+++ b/src/app/components/Age/__tests__/index.test.tsx
@@ -1,0 +1,20 @@
+import { screen } from '@testing-library/react'
+import { renderWithProviders } from '../../../utils/renderWithProviders'
+import { Age } from '../'
+
+describe('Age', () => {
+  beforeEach(() => {
+    jest.useFakeTimers()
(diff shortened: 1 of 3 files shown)


kostko (approved):

src/app/components/Age/__tests__/index.test.tsx
```diff
@@ -0,0 +1,20 @@
...
+import { Age } from '../'
+
+describe('Age', () => {
+  beforeEach(() => {
+    jest.useFakeTimers()
+    jest.setSystemTime(new Date('2024-02-05T10:14:35.000Z'))
+  })
+
+  afterEach(() => {
+    jest.useRealTimers()
```
lubej, line 12: I think `afterEach` is missing `runOnlyPendingTimers` call. According to docs here: https://testing-library.com/docs/using-fake-timers/

buberdds: not sure we need it here
https://jestjs.io/docs/timer-mocks#run-pending-timers

PR #1232, opened 2024-02-06
Consensus validators details layout 

buberdds:
Initial work around details page. Adding screenshots because API expects entity_id as param which is not the best idea.

Screenshot from 2024-02-06 17-58-19
Screenshot from 2024-02-06 17-58-37


src/app/pages/ValidatorDetailsPage/index.tsx
```diff
@@ -1,24 +1,63 @@
...
+import { useRequiredScopeParam } from 'app/hooks/useScopeParam'
+import { useLoaderData } from 'react-router-dom'
+import { AddressLoaderData } from 'app/utils/route-utils'
+import { ValidatorSnapshot } from './ValidatorSnapshot'
 
 export const ValidatorDetailsPage: FC = () => {
-  return <PageLayout>{/* TODO: Implement validator details cards */}</PageLayout>
+  const { isMobile } = useScreenSize()
+  const scope = useRequiredScopeParam()
+  // TODO: currently API does not work with address query param. Wait for API update or switch to entity_id
```
csillag, line 24: Yeah, that sucks. Also when looking at network change proposal votes, we get the address of the validators, and not the entity ID.

src/app/components/Validators/ValidatorLink.tsx
```diff
@@ -0,0 +1,31 @@
...
+  network: Network
+}
+
+export const ValidatorLink: FC<ValidatorLinkProps> = ({ address, name, network }) => {
+  const { isTablet } = useScreenSize()
+  const to = RouteUtils.getValidatorRoute(network, address)
+  return (
+    <Typography variant="mono" component="span" sx={{ color: COLORS.brandDark, fontWeight: 700 }}>
+      {isTablet ? (
+        <TrimLinkLabel label={address} to={to} />
```
csillag, line 23: Don't we also want `name || address` here, instead of just `address` ?
(We do that for the non-tablet version a few lines below)

src/app/pages/ValidatorDetailsPage/index.tsx
buberdds: in addition entity_id is using valid URL chars (like  / + ) that won't be escaped by encodeURIComponent (we would need custom function). Anyway, added a note for backend to change current query param.

lukaw3d (approved):

src/app/components/TrimLinkLabel/index.tsx
```diff
@@ -13,7 +13,25 @@ type TrimLinkLabelProps = {
 
 export const TrimLinkLabel: FC<TrimLinkLabelProps> = ({ label, to, plain }) => {
   const trimmedLabel = trimLongString(label)
+  if (!trimmedLabel) {
+    return null
+  }
+  return <TrimLink label={label} to={to} trimmedLabel={trimmedLabel} plain={plain} />
+}
+
+export const TrimLinkNameLabel: FC<TrimLinkLabelProps> = ({ label, to, plain }) => {
```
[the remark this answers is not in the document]
csillag: I have another implementation for this on another feature branch.

https://github.com/oasisprotocol/explorer/pull/1246/files#diff-8572d402dfed00938becca747e726fe56d9b54f3dbbd75dee395c02ef85cb957

I didn't duplicate the component, just added a prop for strategy (ie. where to truncate).

I haven't decided yet which one I like better, just pointing out the parallel.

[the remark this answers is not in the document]
buberdds: Ok, I will move it away from common components too as it looks like every place will need different trimming params.

PR #1237, opened 2024-02-08
Consensus ParaTimes dashboard section

buberdds:


src/app/pages/ConsensusDashboardPage/ParaTimesCard.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+  return (
+    <Card>
+      <CardHeader
+        disableTypography
+        component="h3"
+        title={
+          <CardHeaderWithCounter
+            label={t('paratimes.listTitle')}
+            totalCount={3}
```
lukaw3d, line 52: Should count enabled+disabled runtimes until we have API

src/app/pages/ConsensusDashboardPage/RuntimePreview.tsx
```diff
@@ -0,0 +1,211 @@
...
+      <Box sx={{ display: 'flex', justifyContent: 'center', paddingTop: 3 }}>
+        {status && (
+          <>
+            <PanelButton
+              activePanel={panel}
+              ariaLabel="Transactions"
+              panel="transactions"
+              setPanel={setPanel}
+            />
+            <PanelButton activePanel={panel} ariaLabel="Accounts" panel="accounts" setPanel={setPanel} />
```
lukaw3d, line 184: Translations

src/app/utils/route-utils.ts
```diff
@@ -124,6 +124,20 @@ export abstract class RouteUtils {
     return Object.values(Layer).filter(layer => RouteUtils.ENABLED_LAYERS_FOR_NETWORK[network][layer])
   }
 
+  static getRuntimesForNetwork(network: Network, isEnabled: boolean): Runtime[] {
+    return Object.values(Runtime).filter(
+      runtime => RouteUtils.ENABLED_LAYERS_FOR_NETWORK[network][runtime] === isEnabled,
+    )
+  }
+
+  static getEnabledRuntimesForNetwork(network: Network): Runtime[] {
```
lukaw3d, line 133: This seems too easily confused with getEnabledLayersForNetwork.

I would refactor all these into: `getAllLayersForNetwork(network): {enabled: [], disabled: []}` + ignore consensus where it is consumed.

src/locales/en/translation.json
```diff
@@ -479,6 +479,15 @@
     "coming": "Coming",
     "soon": "soon"
   },
+  "paratimes": {
+    "active": "Active",
+    "activeNodes": "{{nodes}} active",
+    "blockNumber": "Block number:",
+    "inactive": "Suspended",
```
lukaw3d, line 486: @donouwens this implies that Cipher runtime is suspended, not just that indexer doesn't support it.
image

do we need more status options, or is this just to demo how suspended ones will look like, and we should change Cipher to "Active" but with no data?

buberdds: @donouwens, we don't want to show new ParaTime Pontus-X here, right?

src/app/utils/route-utils.ts
buberdds: done

src/locales/en/translation.json
buberdds: I have changed "Suspended" to "Coming soon" for now. My preference would be to remove status row entirely.

lukaw3d: I expect in the future we are supposed to get data from API about the *actual* status of runtimes

src/app/pages/ConsensusDashboardPage/RuntimePreview.tsx
```diff
@@ -0,0 +1,211 @@
...
+      </StyledList>
+      <StyledBox>
+        <Box gap={3} sx={{ flex: 1, display: 'flex', flexDirection: 'column', alignItems: 'center' }}>
+          {status && (
+            <Box sx={{ width: '100%' }}>
+              {panel === 'transactions' && (
+                <TransactionsChartCard
+                  scope={{
+                    layer: runtime,
+                    network: 'mainnet',
```
lukaw3d, line 146: TODO: un-hardcode mainnet

lukaw3d (approved):

src/locales/en/translation.json
buberdds: Oh I thought it's rather avail/or not avail in Explorer, but yeah useRuntimeFreshness makes sense too. Will add a ticket to address this in the future.

PR #1242, opened 2024-02-09
Support limiting scope to one network or layer - full version

csillag:
It's now possible to specify a given network or layer in the .env file, and when given, the scope of the whole
explorer will be limited to that specific network, or even a specific layer within that network.
No data external to the required scope will be shown. The network (and the layer) will not be part of the URLs.

You can specify just the network, or just the layer, or both. This all results in different behavior.

   * * *

This is part of a larger effort to white-label the Explorer, so that it can be deployed for specific applications.

### DIFF
--- a/.changelog/1242.feature.md
+++ b/.changelog/1242.feature.md
@@ -1,0 +1,1 @@
+Support limiting scope to one network or layer
--- a/src/app/components/LayerPicker/index.tsx
+++ b/src/app/components/LayerPicker/index.tsx
@@ -77,8 +77,8 @@ type LayerPickerContentProps = Omit<LayerPickerProps, 'open'>
 
 enum LayerPickerTabletStep {
   Network,
-  ParaTime,
-  ParaTimeDetails,
+  Layer,
+  LayerDetails,
 }
 
 const LayerPickerContent: FC<LayerPickerContentProps> = ({ isOutOfDate, onClose, onConfirm }) => {
@@ -87,7 +87,7 @@ const LayerPickerContent: FC<LayerPickerContentProps> = ({ isOutOfDate, onClose,
   const { network, layer } = useRequiredScopeParam()
   const [selectedLayer, setSelectedLayer] = useState<Layer>(layer)
   const [selectedNetwork, setSelectedNetwork] = useState<Network>(network)
-  const [tabletStep, setTabletStep] = useState<LayerPickerTabletStep>(LayerPickerTabletStep.ParaTimeDetails)
+  const [tabletStep, setTabletStep] = useState<LayerPickerTabletStep>(LayerPickerTabletStep.LayerDetails)
   const selectNetwork = (newNetwork: Network) => {
     const enabledLayers = RouteUtils.getEnabledLayersForNetwork(newNetwork)
     const targetLayer = enabledLayers.includes(selectedLayer) ? selectedLayer : enabledLayers[0]
@@ -104,7 +104,7 @@ const LayerPickerContent: FC<LayerPickerContentProps> = ({ isOutOfDate, onClose,
       {isTablet && (
         <TabletActionBar>
           <div>
-            {tabletStep === LayerPickerTabletStep.ParaTime && (
+            {tabletStep === LayerPickerTabletStep.Layer && (
               <TabletBackButton
                 variant="text"
                 startIcon={<KeyboardArrowLeft />}
@@ -115,12 +115,12 @@ const LayerPickerContent: FC<LayerPickerContentProps> = ({ isOutOfDate, onClose,
                 {t('layerPicker.viewNetworks')}
               </TabletBackButton>
             )}
-            {tabletStep === LayerPickerTabletStep.ParaTimeDetails && (
+            {tabletStep === LayerPickerTabletStep.LayerDetails && (
               <TabletBackButton
                 variant="text"
                 startIcon={<KeyboardArrowLeft />}
                 onClick={() => {
-                  setTabletStep(LayerPickerTabletStep.ParaTime)
+                  setTabletStep(LayerPickerTabletStep.Layer)
                 }}
               >
                 {t('layerPicker.viewLayers')}
@@ -140,12 +140,12 @@ const LayerPickerContent: FC<LayerPickerContentProps> = ({ isOutOfDate, onClose,
                 selectedNetwork={selectedNetwork}
                 setSelectedNetwork={network => {
                   selectNetwork(network)
-                  setTabletStep(LayerPickerTabletStep.ParaTime)
+                  setTabletStep(LayerPickerTabletStep.Layer)
                 }}
               />
             </Grid>
           )}
-          {(!isTablet || (isTablet && tabletStep === LayerPickerTabletStep.ParaTime)) && (
+          {(!isTablet || (isTablet && tabletStep === LayerPickerTabletStep.Layer)) && (
             <Grid xs={12} md={3}>
               <LayerMenu
                 activeLayer={layer}
@@ -154,12 +154,12 @@ const LayerPickerContent: FC<LayerPickerContentProps> = ({ isOutOfDate, onClose,
                 selectedNetwork={selectedNetwork}
                 setSelectedLayer={layer => {
                   setSelectedLayer(layer)
-                  setTabletStep(LayerPickerTabletStep.ParaTimeDetails)
+                  setTabletStep(LayerPickerTabletStep.LayerDetails)
                 }}
               />
             </Grid>
           )}
-          {(!isTablet || (isTablet && tabletStep === LayerPickerTabletStep.ParaTimeDetails)) && (
+          {(!isTablet || (isTablet && tabletStep === LayerPickerTabletStep.LayerDetails)) && (
             <Grid xs={12} md={6}>
               <LayerDetails
                 handleConfirm={handleConfirm}
--- a/src/app/components/LayerPicker/index.tsx
+++ b/src/app/components/LayerPicker/index.tsx
@@ -13,7 +13,7 @@ import { useRequiredScopeParam } from '../../hooks/useScopeParam'
 import { NetworkMenu } from './NetworkMenu'
 import { LayerMenu } from './LayerMenu'
 import { LayerDetails } from './LayerDetails'
-import { RouteUtils } from '../../utils/route-utils'
+import { scopeFreedom, RouteUtils } from '../../utils/route-utils'
 import { styled } from '@mui/material/styles'
 import KeyboardArrowLeft from '@mui/icons-material/KeyboardArrowLeft'
 import { useScreenSize } from '../../hooks/useScreensize'
@@ -104,28 +104,33 @@ const LayerPickerContent: FC<LayerPickerContentProps> = ({ isOutOfDate, onClose,
       {isTablet && (
         <TabletActionBar>
           <div>
-            {tabletStep === LayerPickerTabletStep.Layer && (
-              <TabletBackButton
-                variant="text"
-                startIcon={<KeyboardArrowLeft />}
-                onClick={() => {
-                  setTabletStep(LayerPickerTabletStep.Network)
-                }}
-              >
-                {t('layerPicker.viewNetworks')}
-              </TabletBackButton>
-            )}
-            {tabletStep === LayerPickerTabletStep.LayerDetails && (
-              <TabletBackButton
-                variant="text"
-                startIcon={<KeyboardArrowLeft />}
-                onClick={() => {
-                  setTabletStep(LayerPickerTabletStep.Layer)
-                }}
-              >
-                {t('layerPicker.viewLayers')}
-              </TabletBackButton>
-            )}
+            {
+              // Do we need a "back to networks" button ?
+              ((scopeFreedom === 'network-layer' && tabletStep === LayerPickerTabletStep.Layer) || // Stepping back from layers
+                (scopeFreedom === 'network' && tabletStep === LayerPickerTabletStep.LayerDetails)) && ( // Stepping back from details, skipping layers
+                <TabletBackButton
+                  variant="text"
+                  startIcon={<KeyboardArrowLeft />}
+                  onClick={() => {
+                    setTabletStep(LayerPickerTabletStep.Network)
+                  }}
+                >
+                  {t('layerPicker.viewNetworks')}
+                </TabletBackButton>
+              )
+            }
+            {scopeFreedom !== 'network' &&
+              tabletStep === LayerPickerTabletStep.LayerDetails && ( // Stepping back from details, going to layers
+                <TabletBackButton
+                  variant="text"
+                  startIcon={<KeyboardArrowLeft />}
+                  onClick={() => {
+                    setTabletStep(LayerPickerTabletStep.Layer)
+                  }}
+                >
+                  {t('layerPicker.viewLayers')}
+                </TabletBackButton>
+              )}
           </div>
           <MobileNetworkButton isOutOfDate={isOutOfDate} network={network} layer={layer} onClick={onClose} />
         </TabletActionBar>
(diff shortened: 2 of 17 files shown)


csillag:
@donouwens  says: 

> All the limiting cases make 100% sense to me on desktop and mobile.
> [...]
> I’m happy to give a go 🟢  and 👍🏻  what you have right now @csillag

src/app/components/LayerPicker/index.tsx
```diff
@@ -77,8 +77,8 @@ type LayerPickerContentProps = Omit<LayerPickerProps, 'open'>
 
 enum LayerPickerTabletStep {
   Network,
-  ParaTime,
```
lubej, line 80: Doesn't ParaTime make more sense than Layer, and the LayerPicker should be renamed to ParaTime

src/app/components/PageLayout/Header.tsx
```diff
@@ -49,7 +66,7 @@ export const Header: FC = () => {
               showText={!scrollTrigger && !isMobile}
             />
           </Grid>
-          {scope && (
+          {scope && needsScopeSelector() && (
```
lubej, line 69: I would avoid doing function calls, isn't this static? In a sense that it wouldn't change on each re-render?

src/app/components/PageLayout/NetworkButton.tsx
```diff
@@ -115,11 +119,14 @@ export const StyledMobileNetworkButton = styled(Button)(({ theme }) => ({
 
 export const MobileNetworkButton: FC<NetworkButtonProps> = ({ isOutOfDate, layer, network, onClick }) => {
   const { t } = useTranslation()
-  const labels = getLayerLabels(t)
+
+  const label = hasFixedLayer() // If we are fixed to a layer,
+    ? getNetworkNames(t)[network] // let's show the name of the network,
+    : getLayerLabels(t)[layer] // otherwise, the name of the layer.
```
lubej, line 125: This could be some utility, it is duplicted twice in this file.

src/app/components/PageLayout/NetworkSelector.tsx
```diff
@@ -84,7 +84,7 @@ const NetworkSelectorView: FC<NetworkSelectorViewProps> = ({ isOutOfDate, layer,
       {!isMobile && (
         <NetworkButton isOutOfDate={isOutOfDate} layer={layer} network={network} onClick={handleDrawerOpen} />
       )}
-      {!isTablet && network !== Network.mainnet && (
+      {!hasFixedLayer() && !isTablet && network !== Network.mainnet && (
```
lubej, line 87: Avoid function calls, this probably stays the same for the whole session. So the returned value, can be static.

src/app/components/ThemeByNetwork/index.tsx
```diff
@@ -14,6 +15,10 @@ export const ThemeByNetwork: FC<{ network: Network; children: React.ReactNode }>
...
 )
 
-export const withDefaultTheme = (node: ReactNode) => (
+export const withMainnetTheme = (node: ReactNode) => (
   <ThemeByNetwork network={Network.mainnet}>{node}</ThemeByNetwork>
 )
+
+export const withDefaultTheme = (node: ReactNode) => (
+  <ThemeByNetwork network={getFixedNetwork() ?? Network.mainnet}>{node}</ThemeByNetwork>
+)
```
lubej, line 24: Should we change theme? Unless with fixed ParaTime only, you can have both mainnet and tesnet?

src/app/hooks/useSearchQueryNetworkParam.ts
```diff
@@ -12,6 +12,12 @@ export const useSearchQueryNetworkParam = (): {
   setNetwork: (network: Network) => void
 } => {
   const [searchParams, setSearchParams] = useSearchParams()
+  if (hasFixedNetwork()) {
```
lubej, line 15: Think that this could be static.

src/app/pages/HomePage/Graph/ParaTimeSelector/index.tsx
```diff
@@ -284,7 +285,7 @@ const ParaTimeSelectorCmp: FC<ParaTimeSelectorProps> = ({
             <HelpScreen setParaTimeStep={setStep} />
           )}
         </ParaTimeSelectorGlobe>
-        {step === ParaTimeSelectorStep.Explore && (
+        {!hasFixedNetwork() && step === ParaTimeSelectorStep.Explore && (
```
lubej, line 288: Showing graph with fixed network/paratime, makes no sense. Homepage should be different in that case, or directly navigate to dashbaord.

src/app/pages/SearchResultsPage/GlobalSearchResultsView.tsx
```diff
@@ -30,6 +30,25 @@ export const GlobalSearchResultsView: FC<{
 
   const themes = getThemesForNetworks()
   const networkNames = getNetworkNames(t)
+
+  if (hasFixedNetwork()) {
```
lubej, line 34: Isn't there a better way to filter search results? Does it makes queries to all the networks/layers still, in case of fixed network/paratime?

src/app/utils/route-utils.ts
```diff
@@ -7,6 +7,32 @@ import { Network } from '../../types/network'
 import { SearchScope } from '../../types/searchScope'
 import { isStableDeploy } from '../../config'
 import { getSearchTermFromRequest } from '../components/Search/search-utils'
+import { isLayerNotHidden } from '../../types/layers'
+
+export const isTesting = () => process.env.NODE_ENV === 'test'
```
lubej, line 12: This should be in env utils. Not here

lubej (commented):
Didn't check all the code yet, partial review.

src/app/components/LayerPicker/index.tsx
csillag: Why? We have consensus, and then we have the paratimes. Here, we can select both, not only the paratimes. The  term that covers both is layer, right?

src/app/hooks/useSearchQueryNetworkParam.ts
csillag: Fixed

src/app/components/PageLayout/NetworkSelector.tsx
csillag: Fixed

src/app/pages/HomePage/Graph/ParaTimeSelector/index.tsx
csillag: Yes, homepage is different.

We _don't_ show the graph when we have a fixed layer. We still show it when we have a fixed network, just without the network switch.

You can check out the demos here: 
- [limited to a network](https://e97738f8.oasis-explorer.pages.dev/)
- [limited to a layer](https://84b385ed.oasis-explorer.pages.dev/)
-  [limited to both network and layer](https://4700fb51.oasis-explorer.pages.dev/)

src/app/utils/route-utils.ts
csillag: OK, moved to `utils/env.ts`

src/app/components/PageLayout/NetworkButton.tsx
csillag: OK, created the `getNetworkButtonLabel()` function.

src/app/pages/HomePage/Graph/ParaTimeSelector/index.tsx
csillag: So the only (limited) case when we still want to show the graph is when we are limited to a network, but layer selection is free. In that case, we can still show the selector graph, just hide the mainnet/testnet switch. And this hiding is the code change here. So I think we are good here.

src/app/pages/SearchResultsPage/GlobalSearchResultsView.tsx
csillag: Yes, good point. We shouldn't even make the queries. Will fix.

csillag: So, looking at the code, we were already limiting the queries to the right network, but not the right layer. Now I have fixed that, too.

So we don't need to filter when displaying the results, so removed that. (But we still need custom code for displaying, because we don't even want to offer other networks, so that part remains.)

csillag:
> Didn't check all the code yet, partial review.

Yeah, thanks, responded to / solved all comments, except one that needs more thinking.

src/app/components/LayerPicker/index.tsx
lubej: Ah you are right, I always forget Consensus :) Layer makes perfect sense.

src/app/utils/route-utils.ts
```diff
@@ -7,6 +7,33 @@ import { Network } from '../../types/network'
...
+
+export const fixedNetwork = isTesting
+  ? undefined
+  : (process.env.REACT_APP_FIXED_NETWORK as Network | undefined)
+export const fixedLayer = isTesting ? undefined : (process.env.REACT_APP_FIXED_LAYER as Layer | undefined)
+
+export type ScopeFreedom =
+  | 'network' // We can select only the network
+  | 'layer' // We can select only the layer
+  | 'both' // We can select both network and layer
```
lubej, line 21: Just a nit, but I would name it network-layer or something similar.

src/types/layers.ts
```diff
@@ -33,5 +33,6 @@ export const doesAnyOfTheseLayersSupportEncryptedTransactions = (layers: Layer[]
   uniq(layers).some(doesLayerSupportEncryptedTransactions)
 
 export const isLayerHidden = (layer: Layer): boolean => hiddenLayers.includes(layer)
+export const isLayerNotHidden = (layer: Layer): boolean => !hiddenLayers.includes(layer)
```
lubej, line 36: Not sure we need this? `!isLayerHidden(layer)`?

src/routes.tsx
```diff
@@ -41,13 +50,79 @@ import { ProposalDetailsPage } from './app/pages/ProposalDetailsPage'
...
+ */
+const unlessFixedOnConsensus = (route: RouteObject): RouteObject[] => (isFixedOnConsensus ? [] : [route])
+
+/**
+ * Use the passed RouteObject if we have a fixed layer but no network
+ *
+ * Normally, this will just wrap the RouteObject into an array.
+ * Unless the condition is true, because in that case, and empty array will be returned.
+ */
+const ifFixedLayerButNoNetwork = (route: RouteObject): RouteObject[] =>
```
lubej, line 100: I would skip `if` naming in a function.

src/routes.tsx
```diff
@@ -41,13 +50,79 @@ import { ProposalDetailsPage } from './app/pages/ProposalDetailsPage'
...
+  if (layer) {
+    const network = RouteUtils.getEnabledNetworksForLayer(layer)[0]
+    if (network) {
+      to = RouteUtils.getDashboardRoute({ network, layer })
+    }
+  }
+  useEffect(() => {
+    if (to) navigate(to)
+  })
+  return <span>Wait...</span>
```
lubej, line 123: Better loader here?

src/routes.tsx
```diff
@@ -41,13 +50,79 @@ import { ProposalDetailsPage } from './app/pages/ProposalDetailsPage'
...
+/**
+ * In case of being restricted to a specific layer, jump to a network
+ *
+ * This should be rendered on the landing page,
+ * in order to redirect to a network, since we don't want the
+ * opening graph.
+ */
+const RedirectToNetwork: FC = () => {
+  const navigate = useNavigate()
+  const layer = fixedLayer
```
lubej, line 112: This declaration is kind of useless.

src/app/components/ThemeByNetwork/index.tsx
```diff
@@ -14,6 +15,10 @@ export const ThemeByNetwork: FC<{ network: Network; children: React.ReactNode }>
...
 )
 
-export const withDefaultTheme = (node: ReactNode) => (
+export const withMainnetTheme = (node: ReactNode) => (
   <ThemeByNetwork network={Network.mainnet}>{node}</ThemeByNetwork>
 )
+
+export const withDefaultTheme = (node: ReactNode) => (
+  <ThemeByNetwork network={fixedNetwork ?? Network.mainnet}>{node}</ThemeByNetwork>
+)
```
lubej, line 24: Still not sure why we need 2 Themes? And are we making sure, we are not accidentally including both?

src/app/components/PageLayout/Header.tsx
```diff
@@ -8,6 +8,23 @@ import { NetworkSelector } from './NetworkSelector'
...
+    case 'network':
+      return RouteUtils.getEnabledNetworksForLayer(fixedLayer!).length > 1
+    case 'layer':
+      return RouteUtils.getVisibleLayersForNetwork(fixedNetwork!).length > 1
+    case 'both':
+      return RouteUtils.getEnabledScopes().length > 1
+    case 'none':
+      return false
+  }
+})() // We will define and call this function only once, and then store the result.
```
lubej, line 27: Does this needs to be an IIFE? I would move this part to utils, and just call the function here, if it is needed only in Header. Or just get the static result from utils, since it probably won't change.

src/app/components/LayerPicker/index.tsx
```diff
@@ -96,6 +96,9 @@ const LayerPickerContent: FC<LayerPickerContentProps> = ({ isOutOfDate, onClose,
   }
   const handleConfirm = () => onConfirm(selectedNetwork, selectedLayer)
 
+  // What kind of selection do we need?
+  const selectionMode = scopeFreedom
```
lubej, line 100: Kinda useless declaration.

src/app/pages/HomePage/Graph/NetworkSelector/index.tsx
```diff
@@ -119,12 +120,19 @@ const NetworkSelectorButton = forwardRef(
...
 }
 
-export const NetworkSelector: FC<NetworkSelectProps> = ({ network, setNetwork }) => {
+export const NetworkSelector: FC<NetworkSelectProps> = ({ network, setNetwork, forLayer }) => {
   const { t } = useTranslation()
 
-  const options = RouteUtils.getEnabledNetworks().map(network => ({
+  const networks = forLayer
+    ? RouteUtils.getEnabledNetworksForLayer(forLayer)
+    : RouteUtils.getEnabledNetworks()
```
lubej, line 131: I have seen this piece of code multiple times now, should be moved to utils I guess.

src/app/utils/route-utils.ts
csillag: OK. Done.

src/routes.tsx
csillag: Yeah it comes from the age when `fixedLayer()` was still a function. Now removed.

src/app/components/LayerPicker/index.tsx
csillag: Removed

src/types/layers.ts
csillag: The difference between `!isLayerHidden(layer)` and `isLayerNotHidden()` is that we can use the latter as an arg for `.filter()`.

I find `layers.filter(isLayerNotHidden)` more readable than `layers.filter(layer => !isLayerHidden(layer))`

I think the extra declaration here is worth so that we can get an easier to read code elsewhere.

What do you think?

src/routes.tsx
csillag: Actually, we are going to be redirected right away, so we don't even need to render a loader. Returning null now, which seems to make it faster somehow.

src/routes.tsx
csillag: The semantics here is that we want to include this option only _if_ we have a fixed layer, but no (fixed) network.
So the function is not testing this condition, but doing something conditionally _if_ the condition is fulfilled.

What naming would you suggest?

Maybe `whenFixedLayerButNoNetwork()` ?

src/app/components/PageLayout/Header.tsx
csillag: Moved function to route-utils.

src/app/components/ThemeByNetwork/index.tsx
csillag: This is also being discussed in another thread, so closing this thread.

src/app/components/ThemeByNetwork/index.tsx
csillag: - `withMainnetTheme()` is only ever used for the external part of the `HomePage` (ie. opening graph). We must use it there, because the `HomePage` is not equipped to be displayed with the `TestTheme`.  (We would get black text on dark blue bg, etc.)
- However, the internal part of the `HomePage` (which shows the actual graph) will still get a theme corresponding to the network currently selected.
- The other wrapper, `withDefaultTheme()` is being used in other locations, when we can use any theme corresponding to the (default) network.

The reason we need 2 wrappers now, when we could do with 1 earlier, is that now  the default network might be testnet, while earlier it was always mainnet, and so when using `withDefaultTheme`, we might get the testnet theme as the default theme, which is not suitable for everywhere.

Please see the demos about how this works in practice.

src/app/pages/HomePage/Graph/NetworkSelector/index.tsx
csillag: Done

src/types/layers.ts
csillag: Removed it anyway.

src/routes.tsx
lubej: Yeah, maybe the one you suggested, or `getFixedLayerButNoNetwork`. I don't know, but `if` feels to me counterintuitive.

src/app/components/ThemeByNetwork/index.tsx
lubej: Ok, still think that a simple if condition could fix this. And we could just use one theme. Or the HOC could accept additional params, to make the theme fixed, in the `HomePage` case. Either way, I guess it's fine.

src/app/components/PageLayout/Header.tsx
```diff
@@ -49,7 +50,7 @@ export const Header: FC = () => {
               showText={!scrollTrigger && !isMobile}
             />
           </Grid>
-          {scope && (
+          {scope && isScopeSelectorNeeded(scope) && (
```
lubej, line 53: Not a fan of function calls inside JSX.

src/app/utils/route-utils.ts
```diff
@@ -7,6 +7,33 @@ import { Network } from '../../types/network'
 import { SearchScope } from '../../types/searchScope'
 import { isStableDeploy } from '../../config'
 import { getSearchTermFromRequest } from '../components/Search/search-utils'
+import { isLayerHidden } from '../../types/layers'
+import { isTesting } from './env'
+
+export const fixedNetwork = isTesting
+  ? undefined
+  : (process.env.REACT_APP_FIXED_NETWORK as Network | undefined)
+export const fixedLayer = isTesting ? undefined : (process.env.REACT_APP_FIXED_LAYER as Layer | undefined)
```
lubej, line 16: Why is this needed? Why would we want to unset fixed network/layer for testing only?

lubej (approved):
LGTM overall

src/app/components/PageLayout/Header.tsx
csillag: Would you prefer having it declared as a const, before the return command?

src/app/utils/route-utils.ts
csillag: Because we want to use simple routes and paths for the tests, and don't want to rewrite them for complicated, conditional logic. 

It's easiest to always run the tests against the basic configuration, independently of how we plan to deploy the app.

In fact, _not_ unsetting for playwright also breaks _those_ tests, which is still an unsolved issue. (How do we easily detect that we are running inside playwrighet?)

src/routes.tsx
csillag: > > Maybe `whenFixedLayerButNoNetwork()` ?

> Yeah, maybe the one you suggested, 

OK, renamed.

src/app/components/ThemeByNetwork/index.tsx
csillag: >  Or the HOC could accept additional params, to make the theme fixed, in the HomePage case.

Did that now. How do you like it now?

src/app/components/PageLayout/Header.tsx
csillag: never mind, moved to a const before the JSX part.

.changelog/1242.feature.md
```diff
@@ -0,0 +1 @@
+Support limiting scope to one network or layer
```
lukaw3d, line 1: This seems hard to maintain. Did any other approaches have issues?

csillag: > This seems hard to maintain.

Well, I guess we can simplify it a great deal if we give up having the shortest possible URLs.... lemme prepare and alternate PR with that version.

csillag: How about #1269?

lukaw3d: It's not a different approach tho

Some other approaches:
- Do we only need to support one of `pinning network | pinning layer | pinning network+layer` for PontusX? Does it matter? Are there issues supporting just one?
- Issues if we tried to unify "fixed" vs unfixed by moving enabled layers to .env.production?
- Can react router intercept and rewrite paths
- ..

PR #1244, opened 2024-02-10
Support for white-labeling Explorer

csillag:
This PR adds several optional configuration options to `.env`, which can be used when re-branding Explorer.

- Use custom App title
- Use custom App description
- Hide build banners (which would otherwise reveal that this is an unofficial / development build)
- Override social texts and links

Also, we can now support displaying 3 new links of social links (see at the end):

image

(Icons suggested by @donouwens )


src/index.tsx
```diff
@@ -26,9 +28,27 @@ const root = ReactDOM.createRoot(document.getElementById('root') as HTMLElement)
...
 root.render(
   <React.StrictMode>
-    <QueryClientProvider client={queryClient}>
-      <RouterProvider router={router} />
-      <ReactQueryDevtools initialIsOpen={false} />
-    </QueryClientProvider>
+    <HelmetProvider>
+      <QueryClientProvider client={queryClient}>
+        {(!!customAppTitle || !!customAppDescription) && (
+          <Helmet>
```
buberdds, line 34: The idea of handling all these things conditionally looks error prone to me. Add HelmetProvider, but don't use it always, add Helmet conditionally, conditional meta tags, values from env, i18n or HTML.

We don't need to support dynamic meta tags. Can't we set meta tags during build time in index.html?

.env
```diff
@@ -6,3 +6,13 @@ REACT_APP_BUILD_VERSION=
 # REACT_APP_TESTNET_API=https://testnet.nexus.stg.oasis.io/v1/
 REACT_APP_API=https://nexus.oasis.io/v1/
 REACT_APP_TESTNET_API=https://testnet.nexus.oasis.io/v1/
+# REACT_APP_NO_BUILD_BANNERS=true
```
buberdds, line 9: With all these entries, can we define env.d.ts?

src/app/utils/externalLinks.ts
```diff
@@ -1,14 +1,32 @@
...
+
+const hasCustomSocialLinks =
+  !!customTelegram || !!customTwitter || !!customDiscord || !!customYouTube || !!customReddit
+
+export const socialMedia = hasCustomSocialLinks
+  ? {
+      telegram: customTelegram,
+      twitter: customTwitter,
+      discord: customDiscord,
+      isDiscordStillValid: true,
```
buberdds, line 17: Let's remove this one.

src/config.ts
```diff
@@ -127,3 +127,6 @@ export const deploys = {
 
 const stableDeploys = [...deploys.production, deploys.staging]
 export const isStableDeploy = stableDeploys.some(url => window.location.origin === url)
+
+export const customAppTitle = process.env.REACT_APP_TITLE
```
buberdds, line 131: Functions like getAppTitle would be useful so components don't reimplement checks

csillag: Would you prefer having a function instead of a const? (I don't think this can change without reloading the app.)

src/app/utils/externalLinks.ts
csillag: Done

src/index.tsx
csillag: >  Can't we set meta tags during build time in index.html?

Would be nice, and less error prone, that's for sure.

.env
```diff
@@ -6,3 +6,13 @@ REACT_APP_BUILD_VERSION=
...
+# REACT_APP_TITLE=Magic Explorer
+# REACT_APP_DESC=The official explorer for my cool project
+# REACT_APP_SOCIAL_HEADER=Check us out
```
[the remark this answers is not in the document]
csillag: Indeed. I'll get rid of that part. They can do that in a fork. Removed.

src/index.tsx
csillag: OK, I got rid of Helmet, and using the `posthtml-expression` parcel plugin to write those meta tags at compile time.

src/config.ts
csillag: OK, introduced `getAppTitle()`.

.env
csillag: Declared the potential entries `global.d.ts`.

.env
```diff
@@ -6,3 +6,14 @@ REACT_APP_BUILD_VERSION=
...
+# REACT_APP_TITLE=Magic Explorer
+# REACT_APP_DESC=The official explorer for my cool project
+# REACT_APP_SOCIAL_TELEGRAM=https://t.me/oasisprotocolcommunity
+# REACT_APP_SOCIAL_TWITTER=https://twitter.com/oasisprotocol
+# REACT_APP_SOCIAL_DISCORD=https://oasis.io/discord
+# REACT_APP_SOCIAL_YOUTUBE=https://www.youtube.com/channel/UC35UFPcZ2F1wjPxhPrSsESQ
+# REACT_APP_SOCIAL_REDDIT=https://www.reddit.com/r/oasisnetwork/
+# REACT_APP_SOCIAL_LINKEDIN=https://www.linkedin.com/company/oasisprotocol
+# REACT_APP_SOCIAL_DOCS=https://oasisprotocol.org/developers#overview
+# REACT_APP_SOCIAL_HOME=https://oasisprotocol.org/
```
lubej, line 19: At least `.env.production` should be set to this variables, otherwise it will break current functionality. Even `.env` should default to Oasis. Unless we use `.env` as template, and start using `.env.development` for development builds.

posthtml.config.js
```diff
@@ -0,0 +1,14 @@
...
+module.exports = {
+  plugins: {
+    'posthtml-expressions': {
+      locals: {
+        APP_TITLE: process.env.REACT_APP_TITLE ?? 'Oasis Explorer',
+        APP_DESCRIPTION: process.env.REACT_APP_DESC ?? 'Official explorer for the Oasis Network.',
+      },
+    },
+  },
+}
```
lubej, line 14: Don't see a reason why we need a default here. And I would remove `posthtml-expressions` dependency.

src/app/components/BuildBanner/index.tsx
```diff
@@ -3,10 +3,12 @@ import { useTranslation } from 'react-i18next'
 import { deploys } from '../../../config'
 import { StickyAlert } from '../StickyAlert'
 
+const useBuildBanners = process.env.REACT_APP_NO_BUILD_BANNERS !== 'true'
```
lubej, line 6: I prefer, `ENABLE` statement flags, so I would do it other way around. But not a big deal. It can stay this way.

src/app/utils/externalLinks.ts
```diff
@@ -1,14 +1,37 @@
...
-  reddit: 'https://www.reddit.com/r/oasisnetwork/',
-}
+const customTelegram = process.env.REACT_APP_SOCIAL_TELEGRAM
+const customTwitter = process.env.REACT_APP_SOCIAL_TWITTER
+const customDiscord = process.env.REACT_APP_SOCIAL_DISCORD
+const customYouTube = process.env.REACT_APP_SOCIAL_YOUTUBE
+const customReddit = process.env.REACT_APP_SOCIAL_REDDIT
+const customLinkedin = process.env.REACT_APP_SOCIAL_LINKEDIN
+const customDocs = process.env.REACT_APP_SOCIAL_DOCS
+const customHome = process.env.REACT_APP_SOCIAL_HOME
```
lubej, line 10: Remove `custom` wording, as it should default to Oasis anyway.

src/app/utils/externalLinks.ts
```diff
@@ -1,14 +1,37 @@
...
+    }
+  : {
+      telegram: 'https://t.me/oasisprotocolcommunity',
+      twitter: 'https://twitter.com/oasisprotocol',
+      discord: 'https://oasis.io/discord',
+      // This API link is for testing if invite is still valid.
+      isDiscordStillValid: 'https://oasis.io/discord/invite-api-check',
+      youtube: 'https://www.youtube.com/channel/UC35UFPcZ2F1wjPxhPrSsESQ',
+      reddit: 'https://www.reddit.com/r/oasisnetwork/',
+    }
```
lubej, line 34: This should default to Oasis in `.env` file.

src/config.ts
```diff
@@ -154,3 +155,5 @@ export const deploys = {
 
 const stableDeploys = [...deploys.production, deploys.staging]
 export const isStableDeploy = stableDeploys.some(url => window.location.origin === url)
+
+export const getAppTitle = (t: TFunction) => process.env.REACT_APP_TITLE ?? t('appName')
```
lubej, line 159: This should be only `REACT_APP_TITLE`, no default.

src/locales/en/translation.json
```diff
@@ -286,7 +286,10 @@
     "reddit": "Reddit",
     "telegram": "Telegram",
     "twitter": "X",
-    "youtube": "Youtube"
+    "youtube": "Youtube",
+    "linkedin": "LinkedIn",
+    "docs": "Documentation",
+    "home": "Home"
```
lubej, line 292: Isn't it there already definition for `Home` somewhere?

package.json
```diff
@@ -119,6 +119,7 @@
     "msw": "^1.3.2",
     "msw-storybook-addon": "^1.10.0",
     "parcel": "2.11.0",
+    "posthtml-expressions": "^1.11.3",
```
buberdds, line 122: throws warning 
https://github.com/oasisprotocol/explorer/actions/runs/7886168455/job/21518858301?pr=1244#step:6:11

src/app/components/Social/index.tsx
```diff
@@ -11,17 +11,23 @@ import twitter from './images/twitter.svg'
...
+import { OverridableComponent } from '@mui/material/OverridableComponent'
+import { SvgIconTypeMap } from '@mui/material/SvgIcon'
 
 type SocialLinkProps = {
   label: string
   href: string
   isMobile: boolean
-  img: string
+  imgSrc?: string
+  img?: OverridableComponent<SvgIconTypeMap> & { muiName: string }
```
buberdds, line 27: Can we just pass an icon component as ReactNode and also remove React.createElement ?

src/config.ts
lubej: And probably appName can be removed, if it is not being used anywhere else

.env
csillag: >  Even `.env` should default to Oasis.

Done that now.

posthtml.config.js
csillag: > Don't see a reason why we need a default here. 

The original idea was that all of these values in `.env` are optional, only required for re-branding, but everything works properly even without them.

We have changed that thinking now, so these values are required, so the defaults are no longer needed, so I have removed them.

csillag: > I would remove `posthtml-expressions` dependency.

What method would you suggest for writing the required meta tags in index.html?

The original implementation used `React-Helmet`, but @buberdds suggested doing it statically at compile time.

As far as I know, with parcel, the way to use env variables in html at compile times is the `posthtml-expressions` plugin. Do you have another suggestion?

src/app/utils/externalLinks.ts
csillag: Yeah, removed those consts.

src/app/utils/externalLinks.ts
csillag: Yes, now we have the oasis links in the `.env` file.

src/config.ts
csillag: Yes, I have done both now.

src/locales/en/translation.json
csillag: No, apparently there isn't.

src/app/components/BuildBanner/index.tsx
csillag: Yeah I agree. Inverted the naming and semantics.

package.json
csillag: Yes, it's a known issue that using JS code for posthtml kills caching of html builds.
However, for deployment, we are hosting the statically build files anyway, and for devs, the loss of speed is insignificant. (The code still builds very fast.)

So I don't think this is an issue. (Just ugly.)

src/app/components/Social/index.tsx
csillag: Done that now.

.env
csillag: > At least `.env.production` should be set to this variables, otherwise it will break current functionality. 

Why is that? As far as I understand, both `.env` and `.env.production` will be read and superimposed. Given that, why should we repeat values?

lubej: https://www.npmjs.com/package/dotenv#should-i-have-multiple-env-files

> Avoid custom set ups that work in inheritance somehow (.env.production inherits values form .env for example). It is better to duplicate values if necessary across each .env.environment file.

posthtml.config.js
csillag: > Could you try `@plasmohq/parcel-transformer-inject-env`

Yes, that works, too. Replaced `posthtml-expression` with that one.

.env
csillag: OK, duplicated values as suggested.

lubej (approved):
LGTM

.env
lukaw3d: ew

src/config.ts
```diff
@@ -146,11 +146,21 @@ export const paraTimesConfig = {
...
+        .split(',')
+        .filter(s => !!s)
+        .map(s => s.trim())
+    : []
+
 export const deploys = {
-  production: ['https://explorer.oasis.io', 'https://explorer.prd.oasis.io'],
-  staging: 'https://explorer.stg.oasis.io',
+  production: parseUrl(process.env.REACT_APP_PROD_URL),
+  staging: parseUrl(process.env.REACT_APP_STAGING_URL),
```
lukaw3d, line 159: A better name would be "splitUrls", and process.env.REACT_APP_PRODUCTION_URLS

src/global.d.ts
```diff
@@ -0,0 +1,25 @@
...
+      REACT_APP_TITLE: string
+      REACT_APP_DESC: string
+      REACT_APP_SOCIAL_TELEGRAM?: string
+      REACT_APP_SOCIAL_TWITTER?: string
+      REACT_APP_SOCIAL_DISCORD?: string
+      REACT_APP_SOCIAL_YOUTUBE?: string
+      REACT_APP_SOCIAL_REDDIT?: string
+      READ_APP_SOCIAL_LINKEDIN?: string
+      READ_APP_SOCIAL_DOCS?: string
+      READ_APP_SOCIAL_HOME?: string
```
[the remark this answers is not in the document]
csillag: Fixed in #1250.

src/config.ts
csillag: Fixed in #1250.

.env
csillag: I'm also not happy about the duplication, but apparently, this is indeed what the developers of the `.env` parsing system have envisioned.

I would much prefer a single source of truth, as opposed to synchronously updating duplicates....
... but I guess the tradeoff is that we don't have to remember who is inheriting what value from where, everything is always flat...

lubej: We could do as I suggested in the first comment. Use `.env` as a template(default) values. But we should not use `.env` for building. And add `.env.development` for development builds.

PR #1245, opened 2024-02-11
Add Pontus-X as a hidden layer

csillag:
- It's not shown on the landing page graph
- It's not listed in the layer selector
- It can be accessed using the correct URL
- Currently all the data is mock data, coming for Sapphire


src/app/components/LayerPicker/LayerDetails.tsx
```diff
@@ -15,7 +15,8 @@ import { Layer } from '../../../oasis-nexus/api'
 import { Link as RouterLink } from 'react-router-dom'
 import { docs } from '../../utils/externalLinks'
 import { TextList, TextListItem } from '../TextList'
-import { getLayerLabels, getNetworkIcons } from '../../utils/content'
+import { getNetworkIcons } from '../../utils/content'
+import { getLayerNames } from '../../../types/layers'
```
buberdds, line 19: Why labels are defined in types folder?

src/app/components/Search/search-utils.ts
```diff
@@ -49,6 +50,13 @@ export const searchSuggestionTerms = {
       suggestedTokenFragment: 'USD',
     },
     cipher: undefined,
+    pontusx: {
```
buberdds, line 53: We don't wan't to show this runtime in search results for now.

src/app/pages/HomePage/Graph/Graph/graph-utils.ts
```diff
@@ -28,6 +28,7 @@ export abstract class GraphUtils {
           y: 0.7 * height,
         }
       case Layer.sapphire:
+      case Layer.pontusx:
```
buberdds, line 31: Please add TODO: note

src/app/pages/ParatimeDashboardPage/LearningMaterials.tsx
```diff
@@ -99,6 +100,24 @@ const getContent = (t: TFunction) => {
         },
       },
       [Layer.cipher]: undefined,
+      [Layer.pontusx]: {
+        // TODO: add proper data
```
buberdds, line 104: Why in one place runtime is hidden and here we wan't to show info ?

src/app/utils/externalLinks.ts
```diff
@@ -48,6 +48,7 @@ export const faucet = {
   [Layer.consensus]: faucetUrl,
   [Layer.emerald]: `${faucetParaTimeBaseUrl}emerald`,
   [Layer.sapphire]: `${faucetParaTimeBaseUrl}sapphire`,
+  [Layer.pontusx]: `${faucetParaTimeBaseUrl}pontusx`,
```
buberdds, line 51: Do we support faucet for 3rd parties?

src/oasis-nexus/v1.yaml
```diff
@@ -0,0 +1,2893 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/OAI/OpenAPI-Specification/main/schemas/v3.1/schema.json
```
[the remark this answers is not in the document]
csillag: By mistake. Removing.

src/types/layers.ts
```diff
@@ -24,9 +25,10 @@ const layerOrder: Record<Layer, number> = {
   [Layer.sapphire]: 2,
   [Layer.emerald]: 3,
   [Layer.cipher]: 4,
+  [Layer.pontusx]: 5,
 }
 
-const hiddenLayers: Layer[] = []
+const hiddenLayers: Layer[] = [Layer.pontusx]
```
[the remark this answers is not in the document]
csillag: What we want here is that the layer _is_ enabled in the sense that exploring it fully works, but we don't advertise it in the opening graph or the layer selector menu.

This is a different use-case compared to only enabling the layer in dev builds, but not in production.

src/app/pages/HomePage/Graph/Graph/graph-utils.ts
csillag: Done

src/app/components/LayerPicker/LayerDetails.tsx
csillag: For historical reasons, I guess? They were always there; all that happened here is that I have discovered that we had two copies of the exact same function (with the same goal and implementation), so I removed one of them.

Would you prefer keeping the other one (which used to live in /utils/content), and removing the one from the types folder?

src/app/components/Search/search-utils.ts
csillag: Good point, will fix that.

src/app/components/LayerPicker/LayerDetails.tsx
buberdds: ah ok we can move it in the future.

src/app/pages/ParatimeDashboardPage/LearningMaterials.tsx
buberdds: So it looks like we need to adjust a type or create a new util type

src/app/utils/externalLinks.ts
buberdds: if we don't know that yet I am not expecting we will get answers soon. Let's revisit in the future.

src/types/layers.ts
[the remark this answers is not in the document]
buberdds: Are you sure about that? The initial requirement was "we'd support it on the landing page of our explorer".

src/app/pages/HomePage/Graph/GraphTooltipMobile/index.tsx
```diff
@@ -163,6 +165,20 @@ const useLayerTooltipMap = (network: Network): Record<Layer, TooltipInfo> => {
           }
         : {}),
     },
+    [Layer.pontusx]: {
```
lubej, line 168: So wait, is this supposed to show in Graph/Homepage or no? Because if no, I wouldn't change any code of Graph, but use some partial(can be new) type of ParaTimes, and use only the ones that Graph actually support. Because for every new ParaTime we need to modify the Graph for no reason.

src/app/pages/ParatimeDashboardPage/LearningMaterials.tsx
```diff
@@ -99,6 +100,24 @@ const getContent = (t: TFunction) => {
         },
       },
       [Layer.cipher]: undefined,
+      [Layer.pontusx]: {
```
lubej, line 103:
```suggestion
      [Layer.pontusx]: undefined,
```

csillag: Again, it's not that simple, because our current type definitions _demand_ content here.

We have a complicated auto-generated utility type here, which will need to be updated..

lubej: Right, we don't even show Cipher, I was just looking at the code. Still... It would make sense to hide the whole learning section, if undefined, not to specify properties.

src/types/layers.ts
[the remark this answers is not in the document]
csillag: I have opened a discussion about this in the relevant chat.

src/app/components/Search/search-utils.ts
csillag: > We don't wan't to show this runtime in search results for now.

Fixed.

src/app/pages/HomePage/Graph/GraphTooltipMobile/index.tsx
csillag: Updated type definition, we no longer have to add unused stuff to the graph code.

src/app/pages/ParatimeDashboardPage/LearningMaterials.tsx
csillag: > So it looks like we need to adjust a type or create a new util type

Adjusted the type for now, so we are using `undefined` now.

We can always add some learning materials if/when we will have proper documentation.

(I will open a thread on the relevant channel about getting the docs.)

src/app/pages/ParatimeDashboardPage/LearningMaterials.tsx
csillag: I fixed the type definitions, so we can use undefined now.

src/app/utils/externalLinks.ts
csillag: > Do we support faucet for 3rd parties?

Now we know that the right way to get test tokens is to send an email. Updated the URL to a `mailto:` link, so the button now opens the email client. This is the best we can do for now.

src/types/layers.ts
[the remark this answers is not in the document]
csillag: The concept of hidden (but accessible) layers has been confirmed by @kostko on chat, so I think we can close this thread.

src/app/components/Search/search-utils.ts
lubej: I would make this `undefined` if possible, the suggested block/tx will show up as not found anyway.

lubej (approved):
LGTM otherwise

src/types/layers.ts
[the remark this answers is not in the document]
buberdds: Just to confirm:
- we don't need hidden layers in Graph because everything is hardcoded there and long term we will show this ParaTime in graph.
- the only place where you want to hide it is Network picker (temporary until Pontus is released). Why we can't just add a condition in layer picker?

src/app/components/Search/search-utils.ts
csillag: We could make this `undefind` (after extending the type a little bit), but the problem is that if don't have search suggestions for a layer, we will just use mainnet sapphire data. So it doesn't help. I think we can leave it as it is now, and update with meaningful suggestions when we have actually started indexing the data.

src/types/layers.ts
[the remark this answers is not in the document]
csillag: Besides the layer picker, we also want to hide this layer in the search results.
That is the two locations where this flag is tested.

Instead of adding two independent guards (one for the search, one for the layer selector), I have defined in in `layers.ts`. How would you do it differently?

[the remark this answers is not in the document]
lubej: Also, search should be disabled for it. So I guess common configuration for hidden layer, makes sense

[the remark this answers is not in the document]
lubej: jinx, you owe me a soda 😂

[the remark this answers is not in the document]
csillag: > jinx, you owe me a soda 😂

👍🏻  Please remind me next time we meet.

Meanwhile, let's merge this and continue with other stuff.

src/app/utils/externalLinks.ts
```diff
@@ -48,6 +48,7 @@ export const faucet = {
   [Layer.consensus]: faucetUrl,
   [Layer.emerald]: `${faucetParaTimeBaseUrl}emerald`,
   [Layer.sapphire]: `${faucetParaTimeBaseUrl}sapphire`,
+  [Layer.pontusx]: 'mailto:contact@delta-dao.com?subject=tokens',
```
lubej, line 51: Well... I missed this change 😓

csillag: And do you have some observation? According to my test, it works great! (I click on the button, MUA opens with email properly prepared.)

lubej: No, just thought that this would be discussed somewhere, prior to merging(heard it mentioned yesterday at a meeting). Probably it would be simple enough to add PontusX to our faucet, with relatively small changes.

Potentially PontusX could track emails. That was/is my primary concern.

csillag: So, as far as I can understand, they have already been distributing test tokens this way, requesting them in email, etc. So there is nothing new here, they just continue the same practice on our network.

However, if we can automate it (and make it more private/secure), I'm all for it.

src/app/pages/SearchResultsPage/GlobalSearchResultsView.tsx
```diff
@@ -33,7 +33,7 @@ export const GlobalSearchResultsView: FC<{
   const otherNetworks = RouteUtils.getEnabledNetworks().filter(isNotMainnet)
   const notificationTheme = themes[Network.testnet]
   const mainnetResults = searchResults.filter(isOnMainnet).sort(orderByLayer)
-  const otherResults = searchResults.filter(isNotOnMainnet).sort(orderByLayer)
+  const otherResults = searchResults.filter(isNotOnMainnet).filter(isNotOnHiddenLayer).sort(orderByLayer)
```
lukaw3d, line 36: Search still always makes requests for Ponuts-X results
(remove `.replace('/v1/pontusx', '/v1/sapphire')` to see them)

src/config.ts
```diff
@@ -108,6 +108,32 @@ const sapphireConfig: LayerConfig = {
...
+    address: undefined,
+    blockGasLimit: undefined,
+    runtimeId: undefined,
+  },
+  testnet: {
+    activeNodes: 8, // TODO use correct number
+    address: 'oasis1qqczuf3x6glkgjuf0xgtcpjjw95r3crf7y2323xd', // TODO use correct address
+    // See max_batch_gas https://github.com/oasisprotocol/sapphire-paratime/blob/main/runtime/src/lib.rs#L166
+    blockGasLimit: 15_000_000,
+    runtimeId: '000000000000000000000000000000000000000000000000a6d1e3ebf60dff6c',
```
lukaw3d, line 123: Use the correct runtimeId and address

csillag: Fixed in #1259.

src/app/pages/SearchResultsPage/GlobalSearchResultsView.tsx
csillag: This is being solved in #1280.

PR #1255, opened 2024-02-14
More white-labeling: support for replacing icons and thumbnail

csillag:
Now web manifest, `og:image` and favicon can also be configured from `.env`

Obviously, when one wants to change the branding, the actual files still need to be added separately, but still, at least the "switching" between different brandings can be done in deployment config, making it easier to maintain multiple versions within the same repo.

(Which is exactly what we intend to do for Pontus-X, at least for the time being.)

Also, added `_META_` to the name of some `.env` entries.


.env
```diff
@@ -7,6 +7,9 @@ REACT_APP_BUILD_VERSION=
 REACT_APP_API=https://nexus.oasis.io/v1/
 REACT_APP_TESTNET_API=https://testnet.nexus.oasis.io/v1/
 REACT_APP_TITLE=Oasis Explorer
+REACT_APP_IMAGE=oasis-og-image.jpg
```
lubej, line 10: Can we add `_META_` into env name, so that it is more clear what is it doing.

lubej (approved):

.env
csillag: > Can we add `_META_` into env name, so that it is more clear what is it doing. And for similar env variables as well.

Done.

PR #1256, opened 2024-02-14
Create Consensus transactions list components

buberdds:


src/app/components/Transactions/ConsensusTransactions.tsx
```diff
@@ -0,0 +1,148 @@
...
+              key: 'to',
+            },
+            {
+              align: TableCellAlign.Right,
+              content: <RoundedBalance value={transaction.fee} ticker={transaction.ticker} />,
+              key: 'fee_amount',
+            },
+            {
+              align: TableCellAlign.Right,
+              // TODO: show RoundedBalance when API returns amount prop
```
csillag, line 128: So, both the amount and the receiver of the TX is missing from the API? 😞

csillag (approved):
Reviewed, LGTM.

csillag:
I would consider using "??" instead of "-" where we have missing data (ie. data that we know exists but is not yet exposed by the API), since "-" could be misinterpreted as meaning "none". What do you think, @donouwens ?

src/app/components/Transactions/ConsensusTransactions.tsx
buberdds: There is backend ticket to fix that https://app.clickup.com/t/869353k3r

buberdds: note for myself: update API layer when props are added (fromBaseUnits) etc

lukaw3d:
I like '-'

but here's how some other options would look like:
image

csillag:
My order of preference:

`-`  < `n/a` < image < `...` < `?` = `??`

PR #1259, opened 2024-02-15
Add more data about the Pontus-X ParaTime

csillag:
~This depends on #1265~.

### DIFF
--- a/src/app/components/LayerPicker/LayerDetails.tsx
+++ b/src/app/components/LayerPicker/LayerDetails.tsx
@@ -24,7 +24,7 @@ import { useScreenSize } from '../../hooks/useScreensize'
 type LayerDetailsContent = {
   description: string
   rpcHttp: string
-  rpcWebSockets: string
+  rpcWebSockets?: string
   chainHexId: string
   chainDecimalId: string
   docs: string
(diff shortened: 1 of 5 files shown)


src/app/components/LayerPicker/LayerDetails.tsx
```diff
@@ -68,6 +68,14 @@ const getDetails = (t: TFunction): Details => ({
       chainDecimalId: '23295',
       docs: docs.sapphire,
     },
+    [Layer.pontusx]: {
+      description: t('layerPicker.testnet.pontusx'),
+      rpcHttp: 'https://pontusx.not.si:443',
+      // rpcWebSockets: 'wss://testnet.sapphire.oasis.dev/ws', // TODO: fix this
```
lubej, line 74: WS support is not even that important. Not sure if Sapphire wrapper even supports it(I could be wrong). The important one is RPC one.

csillag: Yes. It's not that important, that's why I made that an optional field in this description type, and updated the UI code to handle cases when it's missing. I think we can safely merge with this one field missing.

src/config.ts
```diff
@@ -164,3 +173,12 @@ const stableDeploys = [...deploys.production, deploys.staging]
 export const isStableDeploy = stableDeploys.some(url => window.location.origin === url)
 
 export const getAppTitle = () => process.env.REACT_APP_META_TITLE
+
+export const getTickerForScope = ({ network, layer }: SearchScope): NativeTicker => {
+  const networkDefault = getTickerForNetwork(network)
```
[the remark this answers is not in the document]
csillag: It has already been removed as part of the underlying #1265 .

src/app/components/LayerPicker/LayerDetails.tsx
lubej: I would simply delete this line.

lubej (approved):
LGTM

src/app/components/LayerPicker/LayerDetails.tsx
csillag: Deleted.

PR #1260, opened 2024-02-15
Fix a bug in mobile layer selector

csillag:
When opening the layer selector on mobile,
there is button in the top right, showing the
originally selected layer.

Clicking on this button should bring the user back to the originally selected layer, but instead, currently it confirms the current (new) selection.

This change fixes that.

### DIFF
--- a/.changelog/1260.trivial.md
+++ b/.changelog/1260.trivial.md
@@ -1,0 +1,1 @@
+Fix a bug in mobile layer selector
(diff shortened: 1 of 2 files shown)


csillag:
Current behavior: if I'm on sapphire, open the layer selector, click emerald, then this button on the top will say "sapphire" , but if I click on it, it will close the dialog and take me to emerald.

@donouwens says:

>Having it act like a “confirm” button does not make sense to me tbh.
>
> UX principles, and accessibility standards follow the rule that a button should tell the user what it does or what it leads them to.
> Having that button read Sapphire, but take the user to Emerald is not in line with this, and in all honesty does not feel intuitive to me at all.
> 
> Long story short - yes  @csillag  I agree with you that this button should essentially just act as a close button on mobile.

lukaw3d (approved):

PR #1265, opened 2024-02-16
Add support for using different tokens on ParaTimes

csillag:
Up to now, we assumed that the native token on each ParaTime is always the same as the native token on the network.

Unfortunately this is not always true, so now we have to make it possible to use different tokens on paratimes.


src/app/pages/RuntimeTransactionDetailPage/index.tsx
```diff
@@ -102,7 +103,7 @@ export const RuntimeTransactionDetailPage: FC = () => {
     data?.data,
   )
 
-  const tokenPriceInfo = useTokenPrice(getTickerForNetwork(scope.network))
+  const tokenPriceInfo = useTokenPrice(getTickerForScope(scope))
```
lubej, line 106: I would rethink this. Do we want to fetch token price for some other coin? Not sure if we are paying for CoinGecko API, or is it rate limited? Maybe @buberdds knows more

And there is not even a guarantee, that the coin is listed. It will just error out. So there needs to be a consideration of removing the price widget, in case of non ROSE token.

lubej: Also removing every token price information through the application.

buberdds: Your assumptions are correct Matej. We do not pay for API and on a free plan number of requests is limited. I would not worry about limit though, but chances are very high that a token won't be avail in API.

csillag: > Do we want to fetch token price for some other coin? 
> And there is not even a guarantee, that the coin is listed. It will just error out. 

Yes, and that is why we _don't_ actually attempt to fetch them. Here is the code doing the actual fetching:

https://github.com/oasisprotocol/explorer/blob/dc16abce4f809b9b1c8fe910088543ca0704a081/src/coin-gecko/api.ts#L55-L79

We are only ever fetching ROSE, we are faking the data for all others.

So there is no problem with the data not being available.

lubej: There is a problem with that. In case there is no token price data, we should hide the following price widget:

image

Also it makes no sense to show this fields:

image

csillag: >  In case there is no token price data, we should hide the following price widget:

We _are_ hiding the price widget for testnet tokens. Instead, we are advertising the faucet. See [here](https://github.com/oasisprotocol/explorer/blob/csillag/special-ticker-for-paratime/src/app/pages/ParatimeDashboardPage/ParaTimeSnapshot.tsx#L56-L59).

csillag: > Also it makes no sense to show this fields:
> 
> image

Right, and that's why we are also hiding that one, see [here](https://github.com/oasisprotocol/explorer/blob/csillag/special-ticker-for-paratime/src/app/pages/RuntimeTransactionDetailPage/index.tsx#L329-L340).

As far as I can tell, this works as intended.

csillag: (I have now updated the faucet code to better handle the new, par-ParaTime token configuration, but even the original code was smart enough not to show rose price for the testnet when we only have test tokens)

csillag:
Wait, let me adjust this so that we can support multiple tokens for paratimes...

src/app/pages/RuntimeTransactionDetailPage/index.tsx
lubej: Ok, got it. I see it now 👍 Thanks for detailed explanation.

src/app/pages/ParatimeDashboardPage/TestnetFaucet.tsx
```diff
@@ -8,6 +8,8 @@ import { SnapshotCard } from '../../components/Snapshots/SnapshotCard'
 import { COLORS } from '../../../styles/theme/colors'
 import { faucet } from '../../utils/externalLinks'
 import { Layer } from '../../../oasis-nexus/api'
+import { NativeTicker, Ticker } from '../../../types/ticker'
```
lubej, line 11: Just remove this link warning. "'Ticker' is defined but never used"

lubej (approved):

src/app/utils/__tests__/externalLinks.test.ts
```diff
@@ -17,6 +17,7 @@ onlyRunOnCI('externalLinks', () => {
       expect(Object.entries(linksGroup).length).toBeGreaterThan(0)
       // eslint-disable-next-line @typescript-eslint/no-unused-vars
       for (const [_linkName, url] of Object.entries(linksGroup)) {
+        if (Array.isArray(url)) continue // We don't need to go into faucets
```
lubej, line 20: Isn't this a bit of over generalization?

src/app/pages/ParatimeDashboardPage/TestnetFaucet.tsx
csillag: Fixed

src/config.ts
```diff
@@ -1,5 +1,7 @@
 // eslint-disable-next-line no-restricted-imports
-import { Layer } from './oasis-nexus/generated/api' // We get this from the generated code to avoid circular imports
+import { Layer } from './oasis-nexus/generated/api'
+import { getTickerForNetwork, NativeTicker } from './types/ticker'
+import { SearchScope } from './types/searchScope' // We get this from the generated code to avoid circular imports
```
lukaw3d, line 4: Comment is wrong

src/app/utils/__tests__/externalLinks.test.ts
lukaw3d: Please keep externalLinks in a testable format. That is the only purpose of the file.

Move the complex structure somewhere else and import the links into it for mapping

src/app/pages/ParatimeDashboardPage/ParaTimeSnapshot.tsx
```diff
@@ -54,8 +56,8 @@ export const ParaTimeSnapshot: FC<{ scope: SearchScope }> = ({ scope }) => {
           <Nodes scope={scope} />
         </StyledGrid>
         <StyledGrid item xs={22} md={5}>
-          {scope.network === Network.mainnet && <RosePriceCard />}
-          {scope.network === Network.testnet && <TestnetFaucet layer={scope.layer} />}
+          {ticker === Ticker.ROSE && <RosePriceCard />}
+          <TestnetFaucet network={scope.network} layer={scope.layer} ticker={ticker} />
```
lukaw3d, line 60: looks a bit odd without the condition; maybe `{link && <TestnetFaucet link={link} />}`

src/config.ts
csillag: Moved to the right location.

src/app/pages/ParatimeDashboardPage/ParaTimeSnapshot.tsx
csillag: done

src/app/utils/__tests__/externalLinks.test.ts
csillag: Moved the faucet links to a separate file, and removed testing exceptions.

lukaw3d: But faucet link should be tested

src/coin-gecko/api.ts
```diff
@@ -16,74 +17,73 @@ type GetRosePriceParams = {
...
+  const geckoIds = tokens.map(token => token.geckoId).filter((id): id is string => !!id)
+  const { isLoading: geckoIsLoading, data: geckoPrices } = useGetTokenPricesFromGecko(geckoIds, fiatCurrency)
+  const results: AllTokenPrices = {}
+  tokens.forEach(token => {
+    results[token.ticker] = {
+      isLoading: geckoIsLoading,
+      isFree: !!token.free,
+      hasUsedCoinGecko: !!token.geckoId,
+      price: token.geckoId && geckoPrices ? (geckoPrices as any)[token.geckoId] : undefined,
+      fiatCurrency: token.geckoId && geckoPrices ? fiatCurrency : 'xx',
```
lukaw3d, line 85: 'xx'?

csillag: On second thought, we could simply use the provided  `fiatCurrency`, too. It doesn't matter, since (missing the price) it won't be displayed.

csillag: Fixed in #1985.

csillag: Fixed in #1285.

src/app/components/Balance/hooks.ts
```diff
@@ -0,0 +1,83 @@
+import { RuntimeSdkBalance } from '../../../oasis-nexus/api'
+import { AllTokenPrices } from '../../../coin-gecko/api'
+import BigNumber from 'bignumber.js'
+import { Ticker } from '../../../types/ticker'
+
+export const hasRuntimeBalance = (balances: RuntimeSdkBalance[] = []) =>
```
lukaw3d, line 6: unused

src/oasis-nexus/api.ts
```diff
@@ -755,7 +752,7 @@ export const useGetRuntimeEvents: typeof generated.useGetRuntimeEvents = (
                               event.body.amount.Amount,
                               paraTimesConfig[runtime].decimals,
                             ),
-                            Denomination: getTickerForNetwork(network),
+                            Denomination: getTokensForScope({ network, layer: runtime })[0].ticker, // TODO find this out from event data
```
lukaw3d, line 755: We need to differentiate decimals based on denomination. If EUROe has the same number of decimals as TEST, that is only coincidentally and unreliable

src/app/components/Balance/hooks.ts
csillag: Removed in #1285

src/oasis-nexus/api.ts
csillag: Yes, decimals are not yet solved. However this code has been updated later in #1279

PR #1268, opened 2024-02-16
Don't die on missing runtime TX to address

csillag:
For some reason we assumed that there is always a `tx.to_eth` or `tx.to` field,
but that assumption is true, so don't make it.

(What target is missing, don't even try to show anything.)


.changelog/1268.bugfix.md
```diff
@@ -0,0 +1 @@
+Don't die on missing runtime TX to address
```
lukaw3d, line 1: this and PR description are confusing

csillag: how about now?

lubej: How about still confusing?

`Fix optional field to rendering in RuntimeTransactions`

> For some reason we assumed that there is always a tx.to_eth or tx.to field, but that assumption is true, so don't make it.

It was previously presumed that a ‘tx.to_eth’ or ‘tx.to’ field is present, However, both fields are optional.

csillag: Updated accordingly.

lubej (approved):

PR #1279, opened 2024-02-21
Try to auto-detect token type used in runtime transactions

csillag:
Use `tx.body` to figure out if it's something special, or the default native token.

### DIFF
--- a/src/oasis-nexus/api.ts
+++ b/src/oasis-nexus/api.ts
@@ -166,7 +166,6 @@ export const useGetRuntimeTransactions: typeof generated.useGetRuntimeTransactio
   params?,
   options?,
 ) => {
-  const ticker = getTokensForScope({ network, layer: runtime })[0].ticker // TODO: find this out from tx data
   return generated.useGetRuntimeTransactions(network, runtime, params, {
     ...options,
     request: {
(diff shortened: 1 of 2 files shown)


lubej (approved):

src/oasis-nexus/api.ts
```diff
@@ -752,7 +752,9 @@ export const useGetRuntimeEvents: typeof generated.useGetRuntimeEvents = (
                               event.body.amount.Amount,
                               paraTimesConfig[runtime].decimals,
                             ),
-                            Denomination: getTokensForScope({ network, layer: runtime })[0].ticker, // TODO find this out from event data
+                            Denomination:
+                              event.body?.Denomination ??
```
lukaw3d, line 756: This doesn't exist. `event.body?.amount.Denomination` exists.
But `event.body.amount.Denomination === ''` above means that it is empty

csillag: I have seen both existing on different data, IIRC.

csillag: Well, maybe not.

csillag: Right, when this code is executed, it's always empty. Removed useless condition in #1285.

Issue #1276, opened 2024-02-20
Can we reduce the number of concepts for hidden scopes


lukaw3d:
- ENABLED_LAYERS_FOR_NETWORK
- wantedScope (not collapsed in search)
- isLayerHidden (not searchable, but is discoverable)
- fixedLayer (pinned, single enabled layer)
- scopeFreedom

(previously also: scope.valid https://github.com/oasisprotocol/explorer/pull/1159 )

lukaw3d:
If we only ENABLE pontus-x on dev/staging, can we remove HIDDEN layers? Differences: 
- direct links to pontus-x on production won't work https://explorer.oasis.io/testnet/pontusx/block/393840
- searching on staging will find pontus-x results

PR #1280, opened 2024-02-21
Don't run searches against hidden layers

csillag:
(Except of course when we are within those layers already.)


lukaw3d:
(related https://github.com/oasisprotocol/explorer/issues/1276#issuecomment-1957816357)

lukaw3d (approved):

PR #1284, opened 2024-02-21
Consensus Snapshot cards tweaks

buberdds:
Add more details to Consensus snapshot cards

with mocked data:
Screenshot from 2024-02-22 13-31-00

### DIFF
--- a/src/locales/en/translation.json
+++ b/src/locales/en/translation.json
@@ -148,6 +148,7 @@
     }
   },
   "consensusSnapshot": {
+    "blockHeight": "Block Height <BlockHeight />",
     "title": "Consensus Snapshot",
     "totalCirculation": "This is <StakedPercentage /> of the total circulation"
   },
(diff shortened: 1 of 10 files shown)


src/locales/en/translation.json
```diff
@@ -147,7 +147,11 @@
       "explainVerificationDelay": "If you have just verified a contract, it should take a few minutes to update here."
     }
   },
-  "consensusSnapshot": "Consensus Snapshot",
+  "consensusSnapshot": {
+    "blockHeight": "Block Height <BlockHeight />",
+    "title": "Consensus Snapshot",
+    "totalCirculation": "This is <StakedPercentage /> of the total circulation"
```
lubej, line 153: Saw this after looking at a screenshot `%` is missing.

buberdds: added formatter and a comment to validate this later when syncing with API

PR #1290, opened 2024-02-23
Tune links and hashes

csillag:
@donouwens wrote:

> **TX details, mobile view:**
> we should use the space in the row more to minimise the chance of needing to truncate.

Up to now, on mobile, we were always truncating hashes to a fixed length (the same as in tables).

With this change, we can now display them adaptively, that is, taking up all the space that is available.

| Component |Before | After| Points of interest |
|  ---- | ----- | ---- | ---- |
| TX details on mobile | image |  image | Fixed lost space and clashing tooltips for "From" and "To" |
| Account details on mobile | image  |  image  | Fixed lost space for address  |

A similar change has also been implemented for block hash links and other similar things.

### DIFF
--- a/src/app/components/AdaptiveTrimmer/AdaptiveTrimmer.tsx
+++ b/src/app/components/AdaptiveTrimmer/AdaptiveTrimmer.tsx
@@ -1,0 +1,140 @@
+import { FC, ReactNode, useCallback, useEffect, useRef, useState } from 'react'
+import Box from '@mui/material/Box'
+import InfoIcon from '@mui/icons-material/Info'
+import { trimLongString } from '../../utils/trimLongString'
+import { MaybeWithTooltip } from './MaybeWithTooltip'
+
+type AdaptiveTrimmerProps = {
+  text: string | undefined
+  strategy: 'middle' | 'end'
+  extraTooltip?: ReactNode
+}
+
+/**
+ * Display content, potentially shortened as needed.
+ *
+ * This component will do automatic detection of available space,
+ * and determine the best way to display content accordingly.
+ */
+export const AdaptiveTrimmer: FC<AdaptiveTrimmerProps> = ({ text = '', strategy = 'end', extraTooltip }) => {
+  // Initial setup
+  const fullLength = text.length
+  const textRef = useRef<HTMLDivElement | null>(null)
+
+  // Data about the currently rendered version
+  const [currentContent, setCurrentContent] = useState('')
+  const [currentLength, setCurrentLength] = useState(0)
+
+  // Known good - this fits
+  const [largestKnownGood, setLargestKnownGood] = useState(0)
+
+  // Known bad - this doesn't fit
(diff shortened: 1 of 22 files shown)


csillag:
Works with Chrome, but not with FF. Lemme fix that...

csillag:
Works with Chrome, but not with FF. Lemme fix that...

csillag:
> 1. When I view the preview on mobile it's still truncated ("0x1234...abcdef") - that' correct behaviour, right?

It was not. Now fixed.

Firefox: 

image

Chrome:

image

> 2. Would we ever go over two lines on any (normal) screen (mobile/tablet/desktop)?

For account address, no, not before we bring in named accounts in another PR.

src/app/components/AdaptiveTrimmer/AdaptiveTrimmer.tsx
```diff
@@ -0,0 +1,140 @@
...
+            // This the best we can do, for now
+            setInDiscovery(false)
+          } else {
+            // So far, so good, but we should try something longer
+            attemptShortenedContent(Math.floor((newLargestKnownGood + smallestKnownBad) / 2))
+          }
+        }
+      }
+    }
+  }, [inDiscovery, currentLength, largestKnownGood, smallestKnownBad, attemptShortenedContent, fullLength])
```
lukaw3d, line 108: Too much complexity for a small improvement

lukaw3d: If we want this, please implement it using CSS. That will also fix <kbd>CTRL+F</kbd> "0x123456789" finding "0x123...789"

e.g. https://github.com/oasisprotocol/oasis-wallet-web/compare/master...lw/abandoned/address-ellipsis

csillag: A pure CSS solution would be tempting, if it can fulfill all the requirements. Things to check:

- Can we instruct CSS to shorten the string by adding eclipses either in the middle or the end, based on our specific preference?
- If CSS is doing the shortening, then how do we know that we need to add a tooltip?

lukaw3d: - yes
- no

csillag: Can the CSS solution do something similar to [this](https://a193f9df.oasis-explorer.pages.dev/testnet/pontusx/search?q=kunst)?

(When opened with narrow horizontal size, the name will be dynamically resized, beginning and end hidden, but it will always be displayed in a way that the highlight is within the visible section, even if it's not at the middle.)

csillag: OK, going with css.

csillag:
This branch has too many merge conflicts now. I have rebased the code (on the Pontus-X branch), and extracted the relevant parts again into #1396. Closing this as obsolete.

PR #1291, opened 2024-02-23
Fix keyboard accessibility of layer picker component

buberdds:
enable tab + arrows nav within network and layer menus

### DIFF
--- a/src/app/components/LayerPicker/LayerMenu.tsx
+++ b/src/app/components/LayerPicker/LayerMenu.tsx
@@ -82,6 +82,7 @@ export const LayerMenuItem: FC<LayerMenuItemProps> = ({
         setSelectedLayer(layer)
       }}
       selected={activeLayerSelection}
+      tabIndex={activeLayerSelection ? 0 : -1}
     >
       <ListItemText>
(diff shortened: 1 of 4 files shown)


src/app/components/LayerPicker/LayerMenu.tsx
```diff
@@ -82,6 +82,7 @@ export const LayerMenuItem: FC<LayerMenuItemProps> = ({
         setSelectedLayer(layer)
       }}
       selected={activeLayerSelection}
+      tabIndex={activeLayerSelection ? 0 : -1}
```
lukaw3d, line 85: I'm surprised this is needed: I can't reproduce needing it in https://mui.com/material-ui/react-menu/#dense-menu demo, despite modifying towards what we have.

:shrug:  somehow this doesn't work https://github.com/mui/material-ui/blob/b16289b2944996bf578300d5f1290ca998c08260/packages/mui-material/src/MenuList/MenuList.js#L264. maybe if `LayerMenuItem` had `tabIndex` prop and forwarded it to MenuItem, it would work

lukaw3d (approved):

PR #947, opened 2023-10-16
Matomo Analytics (using localStorage)

kaja-osojnik:
Adding a Cookie Banner visual component (without the tracking functionality)
Before:
Screen Shot 2023-10-16 at 12 00 02

After: 
Screen Shot 2023-10-16 at 11 57 30

PR #1294, opened 2024-02-23
Matomo Analytics (using matomo cookies for storage)

lukaw3d:
Alternative to https://github.com/oasisprotocol/explorer/pull/947


package.json
```diff
@@ -77,6 +77,7 @@
     "date-fns": "3.3.1",
     "ethers": "^6.10.0",
     "i18next": "23.8.2",
+    "matomo-types": "git+https://github.com/lukaw3d/matomo-types.git#018cf2cc8c62f7e467f88941725c93c369bffe3b",
```
lubej, line 80: This would be nice, if it was a deployed npm package.

I see that this was forked, any reason why we can't directly reference the original package? Since we are only using types, it shouldn't affect the build.

src/app/components/AnalyticsConsent/redact.ts
```diff
@@ -0,0 +1,15 @@
+export function redactURL(u: string) {
```
lubej, line 1: This looks really great, but are we sure we want redacted urls? Don't see a reason, unless the reason is searching tied to IP address?

It does add a benefit of easily aggregating pages that were visited though.

src/app/components/AnalyticsConsent/initializeMatomo.ts
```diff
@@ -0,0 +1,88 @@
...
+        if (!getCookieWorkaround(CONSENT_COOKIE_NAME) && !getCookieWorkaround(CONSENT_REMOVED_COOKIE_NAME)) {
+          response = 'never-responded'
+        }
+      }
+    },
+  ])
+  return response
+}
+
+function getCookieWorkaround(cookieName: string) {
```
lubej, line 84: Does cookie name changes? Do we have any control over that on BE side?

src/app/components/AnalyticsConsent/index.tsx
```diff
@@ -0,0 +1,120 @@
...
+  const { isMobile } = useScreenSize()
+  const [hasAccepted, setHasAccepted] = useState<matomo.HasAccepted>(matomo.hasAccepted())
+
+  useEffect(() => {
+    window._paq.push([
+      function () {
+        // Matomo loaded
+        setHasAccepted(matomo.hasAccepted())
+        // TODO: bug: still sometimes gets stuck with 'error_matomo_not_loaded' ???
+        console.log('loaded', matomo.hasAccepted())
```
lubej, line 39: In order to be sure that matomo has loaded, onload callback should be added to the script. Since we are dynamically loading it.

One thing that could be helpful is to use useEffectLayout instead of useEffect.

src/app/components/PageLayout/Footer.tsx
```diff
@@ -129,7 +131,8 @@ export const Footer: FC<FooterProps> = ({ scope, mobileSearchAction }) => {
               </StyledLinksGroup>
             </StyledBox>
             <Typography variant="footer">
-              {isTablet ? t('footer.mobileTitle') : t('footer.title')} | {currentYear}
+              {isTablet ? t('footer.mobileTitle') : t('footer.title')} |
```
lubej, line 134: The same, this one should always be `!isTablet`. Or am I missing something?

src/app/components/AnalyticsConsent/index.tsx
```diff
@@ -0,0 +1,120 @@
...
+      },
+    ])
+  }, [])
+
+  const location = useLocation()
+  const [previousURL, setPreviousURL] = useState(document.referrer)
+  useEffect(() => {
+    // Warning: When not opted in, window._paq array will grow indefinitely.
```
lubej, line 47: Can we check here if user has opted in? Otherwise just ignore the code bellow. Don't see I reason, why we would track if user is not opted in. Unless the reasoning is that after you opt-in we will gathered all the data we collected, after user opts in. But in that case I don't like the solution - not even sure what the legal position is in that case.

src/app/components/AnalyticsConsent/initializeMatomo.ts
```diff
@@ -0,0 +1,88 @@
...
+      if (v) {
+        params.set(name, redactURL(v))
+      }
+    }
+
+    return params.toString()
+  },
+])
+
+export function loadMatomo() {
```
lubej, line 38: Does Matomo have any "flush" feature. In a sense that, if we listen to `beforeunload` event. We could force the analytics propagation to BE, otherwise the data will be lost upon page reload.

package.json
lukaw3d: I dont think the fork is on npm + I had to fix some types. I'd like to publish this into @types/matomo, but I'm not sure how to correctly attribute+license it (DefinitelyTyped doesn't have license files per project)

src/app/components/AnalyticsConsent/redact.ts
lukaw3d: Yeah, unlinkability of multiple accounts a single user is often visiting (not just ip in this case: unique visitor id)

package.json
lubej: In that case I would rather see if the types are just copied inside the project to type definition file. And replaced with published npm package, once done.

lukaw3d: Yep, can do that

is there a security concern with github dependencies in npm?

src/app/components/AnalyticsConsent/initializeMatomo.ts
lukaw3d: Cookie seems to only be used on the frontend https://matomo.oasis.io/matomo.js

package.json
lubej: No, not really a security concern. But it is a bit inconsistent with other packages. Since it's only types, it shouldn't matter. But if it was code related, Renovate + Dependabot wouldn't pick up the package for example.

src/app/components/AnalyticsConsent/initializeMatomo.ts
lubej: Disregard my original comment, I was confused because I am used to different `getCookie` snippet. Can we just call the function `getCookie`? Not sure why `Workaround` is mentioned?

lukaw3d: I named it "*Workaround" because I'd like to delete this regex implementation if we don't need it for matomo anymore. I don't think this regex implementation is very safe and robust and probably shouldn't be reused by other code.
At the same time, I'm not sure if using a different more safe implementation could introduce discrepancies with matomo internals.

src/app/components/AnalyticsConsent/initializeMatomo.ts
lubej: Doing some research around `beforeunload`, I don't see this implementation as a problem. You can make a non-blocking HTTP request and close the page. In case that HTTP request fails, the code at least tried to submit the analytics collected - and if lost, it is not such a big deal.

src/app/components/AnalyticsConsent/initializeMatomo.ts
lubej: Got it. If you think it is unsafe, I would name the function `getCookieUnsafe`, workaround naming seems weird to me.

Either way, I like this piece of code with Regex. You need to parse cookies somehow. And I prefer this piece to be included in the code instead of using external library - that would add unnecessary footprint. I would just move it to `CookiesUtils` or something similar.

lukaw3d: Made an upstream issue about the regex implementation: https://github.com/matomo-org/matomo/issues/21998

src/app/components/AnalyticsConsent/initializeMatomo.ts
lukaw3d: Yeah, it seems to be sending tracking data after each action or after a short wait. And docs say it uses navigator.sendBeacon

https://developer.matomo.org/api-reference/tracking-javascript#advanced-uses
> disableAlwaysUseSendBeacon: navigator.sendBeacon which is enabled by default
> setRequestQueueInterval: defaults to 2.5 seconds

src/app/components/AnalyticsConsent/index.tsx
lukaw3d: Instead of onload I'm relying on `window._paq.push([function () { ... }])` being processed

src/app/components/AnalyticsConsent/index.tsx
lukaw3d: +1

Looks like the code comment doesn't apply when using matomo's consent storage (comment was copied from localStorage PR). Matomo automatically throws away events that were pushed before consent. But I'll add an explicit `if consented`.

(Theoretically user should be able to consent to sending navigations that happened before consent, but I don't think we inform them sufficiently.)

src/app/components/AnalyticsConsent/index.tsx
```diff
@@ -0,0 +1,117 @@
...
+  const { t } = useTranslation()
+  const { isMobile } = useScreenSize()
+  const [hasAccepted, setHasAccepted] = useState<matomo.HasAccepted>('timed_out_matomo_not_loaded')
+
+  useEffect(() => {
+    const updateHasAccepted = async () => {
+      setHasAccepted(await matomo.hasAccepted({ timeout: 10_000 }))
+    }
+    updateHasAccepted()
+  }, [])
```
lukaw3d, line 38: Due to how we implemented layout+footer, this gets called a lot.

lubej: You could wrap this in a custom hook, and debounce it(singleton approach). Not sure how heavy `matomo.hasAccepted` is. One thing that is missing here is definitely clean up function(in case component gets destroyed in the meantime, `setHasAccepted` should not be called), if I understand correctly, you are saying that this gets called for every navigation - thats why it gets called a lot.

src/app/components/AnalyticsConsent/initializeMatomo.ts
lubej: In that case, seems fine.

src/app/components/AnalyticsConsent/index.tsx
lubej: Or does `matomo.hasAccepted({ timeout: 10_000 })`, `timeout` parameter mean that it has 10s debouce already? Then I don't see an issue.

lukaw3d: Issue is the whole component gets destroyed and recreated on every navigation

(I don't see any issue or need for a cleanup function)

lukaw3d:
i think this works now

PR #1296, opened 2024-02-24
Improve phrasing of error messages on reverted transactions

csillag:
As discussed in chat, for EVM errors, we want to add back the "Reverted:" prefix in case of error code 8.
Also, on missing messages, say "Reverted without e massage."

### DIFF
--- a/src/app/hooks/useTxErrorMessage.ts
+++ b/src/app/hooks/useTxErrorMessage.ts
@@ -3,5 +3,12 @@ import { TxError } from '../../oasis-nexus/api'
 
 export const useTxErrorMessage = (error: TxError | undefined): string | undefined => {
   const { t } = useTranslation()
-  return error ? `${error.message} (${t('errors.code')} ${error.code})` : undefined
+  if (!error) return undefined
+  if (error.module === 'evm' && error.code === 8 && !error.message) {
+    // EVM reverted, with missing error message
+    return `${t('errors.revertedWithoutMessage')} (${t('errors.code')} ${error.code})`
+  } else {
+    // Anything else
+    return `${error.message} (${t('errors.code')} ${error.code})`
+  }
 }
--- a/src/app/hooks/useTxErrorMessage.ts
+++ b/src/app/hooks/useTxErrorMessage.ts
@@ -1,0 +1,7 @@
+import { useTranslation } from 'react-i18next'
+import { TxError } from '../../oasis-nexus/api'
+
+export const useTxErrorMessage = (error: TxError | undefined): string | undefined => {
+  const { t } = useTranslation()
+  return error ? `${error.message} (${t('errors.code')} ${error.code})` : undefined
+}
--- a/src/locales/en/translation.json
+++ b/src/locales/en/translation.json
@@ -223,6 +223,7 @@
     "validateURL": "Please validate provided URL",
     "validateURLOrGoToFirstPage": "Please check the URL or <FirstPageLink />.",
     "invalidUrl": "Invalid URL",
+    "revertedWithoutMessage": "reverted without a message",
     "storage": "Access to browser storage denied"
   },
   "footer": {
--- a/.changelog/1296.trivial.md
+++ b/.changelog/1296.trivial.md
@@ -1,0 +1,1 @@
+Improve phrasing of error messages on reverted transactions
--- a/src/app/components/StatusIcon/index.tsx
+++ b/src/app/components/StatusIcon/index.tsx
@@ -8,6 +8,7 @@ import { COLORS } from '../../../styles/theme/colors'
 import HelpIcon from '@mui/icons-material/Help'
 import { TxError } from '../../../oasis-nexus/api'
 import Tooltip from '@mui/material/Tooltip'
+import { useTxErrorMessage } from '../../hooks/useTxErrorMessage'
 
 type TxStatus = 'unknown' | 'success' | 'failure'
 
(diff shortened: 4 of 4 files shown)


src/app/hooks/useTxErrorMessage.ts
```diff
@@ -0,0 +1,18 @@
+import { useTranslation } from 'react-i18next'
+import { TxError } from '../../oasis-nexus/api'
+
+export const useTxErrorMessage = (error: TxError | undefined): string | undefined => {
+  const { t } = useTranslation()
+  if (!error) return undefined
+  if (error.code === 8) {
```
lubej, line 7: This is a nice quick fix, but I think we really should discuss with BE, if they want to add support for this on Nexus.

lukaw3d: What does code 8 mean? Any docs?

lubej: > All the errors are defined per module in the oasis sdk repository 

this was posted by @kostko in Slack

csillag: > Evm code 8 is reverted :) 

> there is a large amount of error codes and modules. But one thing about evm/8 is that the contract can customize the revert error message. And it is also possible for it to be empty (which is the one that nexus/explorer don’t handle in the best way currently)

> All the errors are defined per module in the oasis sdk repository

> But most others don’t have custom error messages so you will already get the right thing :)

> So the error message for the above returned by the runtime is actually “reverted: ” because thr custom part comes after the colon, but nexus is using a custom decoder for that since when the custom message exists, it is evm abi and base64 encoded. So it is stripping the “reverted: “ part

kostko: Oh this is actually wrong, it shouldn't just match on code 8, it also needs to match the module. Error codes are namespaced per module, e.g. evm/8 is defined [here](https://github.com/oasisprotocol/oasis-sdk/blob/9fb148a52636d7c1e8f839d1caa97aceef449e04/runtime-sdk/modules/evm/src/lib.rs#L157-L159).

csillag: It has been discussed with @donouwens and @kostko that adding back the "Reverted" part, and saying "without a message" is the right course of action for now.

csillag: >  it shouldn't just match on code 8, it also needs to match the module. 

Updated accordingly.

src/locales/en/translation.json
```diff
@@ -223,6 +223,8 @@
     "validateURL": "Please validate provided URL",
     "validateURLOrGoToFirstPage": "Please check the URL or <FirstPageLink />.",
     "invalidUrl": "Invalid URL",
+    "reverted": "Reverted:",
```
kostko, line 226: Error messages are usually all lower-case:
```suggestion
    "reverted": "reverted:",
```

lukaw3d:
Is backend already trying to do this? https://github.com/oasisprotocol/nexus/blob/acfb224aecec87229d30fa5cf5fffe4e732ecb7f/storage/oasis/nodeapi/evm.go#L33

csillag:
> Is backend already trying to do this? https://github.com/oasisprotocol/nexus/blob/acfb224aecec87229d30fa5cf5fffe4e732ecb7f/storage/oasis/nodeapi/evm.go#L33

Apparently not.

Looking at the response [here](https://testnet.nexus.oasis.io/v1/sapphire/transactions/872d95e14bd497f35209539341958346a4e20ddf66d1ef2f282251294aa5d3d0), all we get for the error is:

```JSON
"error": {
    "code": 8,
    "module": "evm"
}
```

There is no message. So in this case, we have agreed to add a message (in the frontend) saying "reverted because of a message".

We probably don't have to do anything with any other cases, for now.

csillag:
Should be fine now.

src/app/hooks/useTxErrorMessage.ts
```diff
@@ -0,0 +1,14 @@
+import { useTranslation } from 'react-i18next'
+import { TxError } from '../../oasis-nexus/api'
+
+export const useTxErrorMessage = (error: TxError | undefined): string | undefined => {
+  const { t } = useTranslation()
+  if (!error) return undefined
+  if (error.module === 'evm' && error.code === 8 && !error.message) {
```
lubej, line 7: One alternative would be to just do this with translations. For example:

t(`errors.${error.module}.${error.code}${error.message ?? ''}`)

And the key would be `errors.evm.8` -> poiniting to custom error.

I imagine that there will be other edge cases, that would need this and it would look nicer with `switch` and direct translation implementation.

lubej (approved):
LGTM for now, but we should reconsider my suggestion, if there will be bunch of edge cases like this.

src/app/hooks/useTxErrorMessage.ts
csillag: The possibility of other corner cases has been discussed at length on Slack, and the conclusion was there is probably nothing else, and if there is, we want to handle it in Nexus.

mitjat:
> Is backend already trying to do this? https://github.com/oasisprotocol/nexus/blob/acfb224aecec87229d30fa5cf5fffe4e732ecb7f/storage/oasis/nodeapi/evm.go#L33

Yes but we have an bug (or possibly a conscious decision at the time) in the corner case where there is no message. We'll fix it eventually but it will need a reindex, so it's useful to have this patch on top.

PR #1307, opened 2024-02-28
Add support for displaying privacy policy when required

csillag:
Our deployment doesn't need a privacy page.
Some forks do. (For example, Pontus-X).

So here, we are just adding the config flag, the link and and (conditional) root.
Any content should be added on the fork.

For an example of actually using this, see https://github.com/oasisprotocol/explorer/pull/1247/commits/d184612b4ce61fc8c24d93b7ddebf309e71e5eb1


lubej (commented):
I don't know if this really makes sense, why wouldn't a fork add their own route? Since as you mentioned in the PR description, we don't need it. I don't see any benefit for it.

csillag:
> I don't know if this really makes sense, why wouldn't a fork add their own route? Since as you mentioned in the PR description, we don't need it. I don't see any benefit for it.

The benefit is that the routing file doesn't have to be changed on the fork, only the insides of the Privacy page file.

I think that keeping the changes of the fork minimal and localized, and hence easy to rebase, is valuable.

lubej:
> > I don't know if this really makes sense, why wouldn't a fork add their own route? Since as you mentioned in the PR description, we don't need it. I don't see any benefit for it.
> 
> The benefit is that the routing file doesn't have to be changed on the fork, only the insides of the Privacy page file.
> 
> I think that keeping the changes of the fork minimal and localized, and hence easy to rebase, is valuable.

Ok I agree on the rebasing thing. How about we comment out the code in this PR? So it stays inside the codebase, but gets filtered out when making a build. I really don't see why we should include this in our runtime(I know it is not heavy), but if we are not going to use `PrivacyPage`, I don't see a reason why it should be included.

csillag:
> How about we comment out the code in this PR? So it stays inside the codebase, but gets filtered out when making a build.

The problem with commented out code is that we don't notice when we break it with refactorings.

> If we are not going to use `PrivacyPage`, I don't see a reason why it should be included.

It's included but disabled. (Route is not active, link is not shown.)

* Current JS bundle size _with_ privacy page stub: 11,220,914 bytes
* Current JS bundle size _without_ privacy page stub: 11,216,530 bytes

So the extra load is  ~4400 bytes (_before_ compression), adding 0,04% of size.

I think that is a good price to pay for reducing the scope of difference between the different branches / forks.

(But this is not the hill I want to die on.)

src/routes.tsx
```diff
@@ -264,6 +266,14 @@ export const routes: RouteObject[] = [
           },
         ],
       },
+      ...(showPrivacyPolicy
```
buberdds, line 269: This does not belong to our code base. If we know that they want to add custom pages:
1) they do everything in a fork
2) we can limit routes merge conflicts by moving custom routes to separate file. Instead of this we can have sth like
 `...custom3rdPartyRoutesSlot,`

custom3rdPartyRoutesSlot.tsx 

```
// Used by 3rd parties to add custom routes to the app
export const custom3rdPartyRoutesSlot = []
```
      
their fork

```
// Used by 3rd parties to add custom routes to the root of the app
import { withDefaultTheme } from './app/components/ThemeByNetwork'
import { PrivacyPage } from './app/pages/PrivacyPage'

export const customRoutesSlots = [
   {
     path: '/privacy',
     element: withDefaultTheme(<PrivacyPage />),
   },
]
```

either way this repo should not be aware of Pontus-X custom routes.

csillag: > we can limit routes merge conflicts by moving custom routes to separate file. 

I have done that in #1345.

csillag:
Superseded by #1345.

PR #1308, opened 2024-02-29
Adjust title levels: use h2 for main title on each page

csillag:
@donouwens says:

>  I’m totally fine with turning the current H3 into H2.

### DIFF
--- a/src/styles/theme/defaultTheme.ts
+++ b/src/styles/theme/defaultTheme.ts
@@ -143,8 +143,8 @@ export const defaultTheme = createTheme({
     },
     h2: {
       fontSize: '24px',
-      fontWeight: 500,
-      lineHeight: '140%',
+      fontWeight: 600,
+      lineHeight: '32px',
       color: COLORS.brandExtraDark,
     },
     h3: {
--- a/.changelog/1308.trivial.md
+++ b/.changelog/1308.trivial.md
@@ -1,0 +1,1 @@
+Adjust title levels: use h2 for main title on each page
--- a/src/app/components/Snapshots/Snapshot.tsx
+++ b/src/app/components/Snapshots/Snapshot.tsx
@@ -26,7 +26,7 @@ export const Snapshot: FC<SnapshotProps> = ({ children, header, scope, title }) 
           <AppendMobileSearch scope={scope}>
             <Box sx={{ display: 'flex', flexDirection: isMobile ? 'column' : 'row', mb: 2 }}>
               <Typography
-                variant="h3"
+                variant="h2"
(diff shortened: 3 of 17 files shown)


src/styles/theme/defaultTheme.ts
```diff
@@ -136,21 +136,21 @@ export const defaultTheme = createTheme({
...
       color: COLORS.brandExtraDark,
     },
     h3: {
       fontSize: '24px',
-      fontWeight: 600,
-      lineHeight: '32px',
+      fontWeight: 'bold',
+      lineHeight: '140%',
       color: COLORS.brandExtraDark,
     },
```
lubej, line 155: Are you sure, there were no previous h2 components?

csillag:
> I have just opened a random view. Is this really what we want ?

So, following what @donouwens said, I have 

* Replaced all H3 with H2
* Updated the styles for H1, H2 and H3, including changing the size of H2 from 24 to 32.

This means that things that used to be size H3 24px are now H2 32px.

This might cause some unwanted effects, that's why I asked @donouwens to check.

I think he did, but of course more 👀 eyes can find more problems. I will ask him to check again, specifically [this page](https://71f5d01c.oasis-explorer.pages.dev/mainnet/sapphire/token/0x998633BDF6eE32A9CcA6c9A247F428596e8e65d8)

csillag:
OK this has been completely retooled.

Explanation for the latest version:
 - We don't change the font sizes of any of the existing styles (h1, h2 or h3)
 - h2 and h3 were already looking pretty much the same (same font size), now they are completely the same. 
 - Some of the titles that used to be h3 are not h2, but that doesn't matter since they look the same
   - For this purpose, `SubPageCard` got an optional prop called `mainTitle`, which instructs it to render as h2 instead of h2. This prop is added for the first / main card on each page.

Again, the look doesn't change, just the semantics: now there is a valid h2 title on most of the pages.

csillag:
> I have just opened a random view. Is this really what we want ?

Totally now. Fixed. Now there is no further changes in font sizes.

lukaw3d (approved):

csillag:
@lukaw3d you have approved this, but we have also discussed earlier that we actually want h1 for those titles, not h2. This PR is currently changing them from h3 to h2.

So, before merging should I:

1. Change it from h2 to h1?
2. Make the logo _not_ h1 (as discussed)?
3. Update the wording, ie. append something like "On Testnet Sapphire" to the titles, as discussed?

lukaw3d:
i'd merge this improvement and put those into separate PRs

PR #1317, opened 2024-03-06
Show unencrypted contract creation data

csillag:
(As part of Runtime TX details)

### DIFF
--- a/src/app/pages/RuntimeTransactionDetailPage/index.tsx
+++ b/src/app/pages/RuntimeTransactionDetailPage/index.tsx
@@ -364,6 +364,15 @@ export const RuntimeTransactionDetailView: FC<{
             </>
           )}
 
+          {!transaction.encryption_envelope && transaction.body?.init_code && (
+            <>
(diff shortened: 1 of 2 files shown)


src/app/pages/RuntimeTransactionDetailPage/index.tsx
```diff
@@ -364,6 +364,15 @@ export const RuntimeTransactionDetailView: FC<{
             </>
           )}
 
+          {!transaction.encryption_envelope && transaction.body?.init_code && (
```
buberdds, line 367: Can we end up with two `transaction.rawData` rows or it's impossible to have `!!transaction.body?.data` and `!transaction.encryption_envelope && transaction.body?.init_code` in one tx ?

buberdds (approved):

src/app/pages/RuntimeTransactionDetailPage/index.tsx
csillag: Great question. I fwd-ed it to the Nexus guys for clarification.

PR #1325, opened 2024-03-13
Show instructions if analytics is adblocked and user clicks privacy settings

lukaw3d:
image

PR #1327, opened 2024-03-14
Add e2e tests for analytics

lukaw3d:
Tests https://github.com/oasisprotocol/explorer/pull/1325 and https://github.com/oasisprotocol/explorer/pull/1294


src/app/components/AnalyticsConsent/index.tsx
```diff
@@ -12,12 +12,15 @@ import { Network } from '../../../types/network'
 import { AnalyticsIsBlocked } from './AnalyticsIsBlocked'
 import { AnalyticsDialogLayout } from './AnalyticsDialogLayout'
 
+// Allow e2e tests to pre-define it in runtime
+window.REACT_APP_ENABLE_OASIS_MATOMO_ANALYTICS ??= process.env.REACT_APP_ENABLE_OASIS_MATOMO_ANALYTICS as any
```
lukaw3d, line 16: Thoughts on this as a workaround to inject env vars during e2e tests?
(similar to our discussion for setting REACT_APP_API after build during deploy)

lubej: This looks dangerous to me, and I would avoid setting it here. The reason for it, is that you are introducing race condition IMO. Image you  access `window.REACT_APP_ENABLE_OASIS_MATOMO_ANALYTICS` variable in any other src file, you can't guarantee that the `window.REACT_APP_ENABLE_OASIS_MATOMO_ANALYTICS` will be available(sure, will be available in components, but not top level). 

So the only right place do define this, before this .js bundle, which is in `index.html`. Or adding another script, that would set this window ENV variable only, but you need to make sure it runs before the app bundle.

lubej: Actually, disregard the above, if you add this line in entry file `src/index.tsx` it should work fine.

src/types/global.d.ts
```diff
@@ -5,7 +5,7 @@ declare global {
       REACT_APP_BUILD_DATETIME: string
       REACT_APP_BUILD_SHA: string
       REACT_APP_BUILD_VERSION: string
-      REACT_APP_ENABLE_OASIS_MATOMO_ANALYTICS: 'true' | 'false'
+      REACT_APP_ENABLE_OASIS_MATOMO_ANALYTICS: 'read from window.REACT_APP_ENABLE_OASIS_MATOMO_ANALYTICS instead (needed for e2e tests)'
```
lubej, line 8: Why would you change the type? The type is still `'true' | 'false'`

src/app/components/AnalyticsConsent/index.tsx
lubej: You could add `env.ts` as first import(but we need to be sure, that the order won't change), and set the `window.REACT_APP_ENABLE_OASIS_MATOMO_ANALYTICS` there. 

There is a prettier plugin I used before `https://www.npmjs.com/package/@trivago/prettier-plugin-sort-imports`, where you can define a regex, so that the order is always set. I think we should include it anyway. 'env files', 'node_modules packages' and then 'local relative imports'.

src/app/components/AnalyticsConsent/index.tsx
```diff
@@ -1,4 +1,4 @@
-/* eslint-disable react-hooks/rules-of-hooks -- REACT_APP_ENABLE_OASIS_MATOMO_ANALYTICS can't change in runtime */
+/* eslint-disable react-hooks/rules-of-hooks -- REACT_APP_ENABLE_OASIS_MATOMO_ANALYTICS won't change in runtime */
```
lubej, line 1: Is this line still necessary after `window.REACT_APP_ENABLE_OASIS_MATOMO_ANALYTICS` ?

lukaw3d: yep

src/app/components/AnalyticsConsent/index.tsx
lukaw3d: I'll rather keep it in html

Import sorting plugin: I think import order so rarely matters that I wouldn't add it. So far import order mattered with CSS imports, polyfills, and this env workaround.

However, import sorting could reduce conflicts:
- if two PRs add the same import in different positions, moving them to the same line would not cause a conflict
- if two PRs add different imports to the same line, moving them apart (e.g. sort by hash(lineContent)) would automatically resolve the conflict

lubej: Yeah, exactly it does cause less headaches with conflicts. Also it does matter if you have circular dependency 😅 but anyway, yeah I always used it for .css imports(that reminds me I should add it to dapp-voting & dapp-wrose)

If you want to keep it in html, thats fine. Would be nice to document it somewhere(README) - for that matter, documenting every env variable would be nice.

Will add it to backlog, to add the above mentioned plugin for prettier.

lubej (approved):

src/app/components/AnalyticsConsent/index.tsx
lukaw3d: Can we avoid documentation by using better env variable names? (and just link to global.d.ts for a list of all options)

lubej: Sure, using JDoc would work, just some one line description of the variable would help. Sometimes it is not trivial. 

For example setting `REACT_APP_FIXED_NETWORK` or `REACT_APP_FIXED_LAYER` is not trivial, what it excludes/disables in the app at the first glance.

PR #1329, opened 2024-03-15
Fix oasis address links even when toggled to ETH addresses

### LINKED ISSUE #1220: [Bug]: Can't open non-EVM transactions from vertical list of transactions

- [X] I have searched the existing issues

### SUMMARY

_No response_

### STEPS TO REPRODUCE

- open https://explorer.dev.oasis.io/mainnet/sapphire/tx on mobile
- switch to vertical layout
- 301523890-5f3e16f3-015c-4334-8e36-53a1bd3da48c
- Consensus Deposit transactions don't link to transaction details page

### EXPECTED BEHAVIOR

_No response_

lukaw3d:
Makes "plain" links still link to the address. User might not have known they needed to toggle the address format before they could open a "plain" address.

Fixes #1220: oasis addresses in e.g. consensus deposit transactions in EVM runtimes were not clickable in mobile vertical layout of transactions because it is implicitly toggled to ETH addresses (and can't be toggled to Oasis).

| Before | After |
| --- | --- |
| localhost_1234_mainnet_sapphire_tx_2c997a87b4f57c0266a780cab8bb1f39b35d37d6c894c79e2a43395b3323ad85 (1) | image |
| localhost_1234_mainnet_sapphire_tx_2c997a87b4f57c0266a780cab8bb1f39b35d37d6c894c79e2a43395b3323ad85 (3) | image |

### DIFF
--- a/src/app/components/Account/AccountLink.tsx
+++ b/src/app/components/Account/AccountLink.tsx
@@ -11,16 +11,15 @@ export const AccountLink: FC<{
   scope: SearchScope
   address: string
   alwaysTrim?: boolean
-  plain?: boolean
-}> = ({ scope, address, alwaysTrim, plain }) => {
+}> = ({ scope, address, alwaysTrim }) => {
   const { isTablet } = useScreenSize()
   const to = RouteUtils.getAccountRoute(scope, address)
   return (
     <Typography variant="mono" component="span">
       {alwaysTrim || isTablet ? (
-        <TrimLinkLabel label={address} to={to} plain={plain} />
+        <TrimLinkLabel label={address} to={to} />
       ) : (
-        <Link component={RouterLink} to={to} sx={{ fontWeight: plain ? 400 : undefined }}>
+        <Link component={RouterLink} to={to}>
           {address}
         </Link>
       )}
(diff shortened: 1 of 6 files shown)


src/app/components/Account/AccountLink.tsx
```diff
@@ -17,21 +16,11 @@ export const AccountLink: FC<{
...
-    >
+    <Typography variant="mono" component="span">
       {alwaysTrim || isTablet ? (
-        <TrimLinkLabel label={address} to={to} />
-      ) : plain ? (
-        address
+        <TrimLinkLabel label={address} to={to} plain={plain} />
       ) : (
-        <Link component={RouterLink} to={to}>
+        <Link component={RouterLink} to={to} sx={{ fontWeight: plain ? 400 : undefined }}>
```
buberdds, line 23: all links should be 700 according to Don

lukaw3d: > all links should be 700 according to Don (same for TransactionLink). If that is true, why we even need plain?

fixed

> I am a bit confused what plain is doing now (in master not this PR). We render (...)Link component and we use that prop to not render a link?

in most cases it displayed just text instead of a link, I guess so that toggling to ETH addresses would not contain confusing oasis1 addresses :shrug:

buberdds (approved):

PR #1344, opened 2024-03-26
Reuse useFreshness logic and enable offline banner in Consensus

buberdds:
getStatus is what we need here, right?


lukaw3d (approved):
Yep

Note: we don't have RuntimeOfflineBanner equivalent for consensus

PR #1345, opened 2024-03-27
Add extension slots for custom routes and footer links

csillag:
Following the [suggestion](https://github.com/oasisprotocol/explorer/pull/1307#discussion_r1507204787) of @buberdds, this PR adds extension slots for adding custom links to the footer, and routes to the router, so that if a 3rd party (i.e. a fork) wants to add custom pages, it can do it in a way that minimizes the change of merge conflicts.

### DIFF
--- a/src/app/components/PageLayout/Footer.tsx
+++ b/src/app/components/PageLayout/Footer.tsx
@@ -11,6 +11,7 @@ import { AppendMobileSearch } from '../AppendMobileSearch'
 import { SearchScope } from '../../../types/searchScope'
 import { api, github } from '../../utils/externalLinks'
 import { ReopenAnalyticsConsentButton } from 'app/components/AnalyticsConsent'
+import { CustomFooter } from '../../config/CustomFooter'
 
 const FooterBox = styled(Box)(({ theme }) => ({
   display: 'flex',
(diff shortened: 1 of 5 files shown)


src/app/components/PageLayout/Footer.tsx
```diff
@@ -129,6 +130,7 @@ export const Footer: FC<FooterProps> = ({ scope, mobileSearchAction }) => {
                 </Typography>
               </StyledLinksGroup>
             </StyledBox>
+            <CustomFooter />
```
lukaw3d, line 133: Doesn't appear on mobile

csillag: Also added to mobile.
(And added commented-out example code for proper formatting, which is a bit tricky to get right)

csillag:
See #1354 for an alternative of this

csillag:
> See #1354 for an alternative of this

#1354 has been merged, but according to @lukaw3d, we probably [still want](https://github.com/oasisprotocol/explorer/pull/1354#pullrequestreview-1972521953) this.

So I have rebased, and it's ready for review again.

lukaw3d:
Update: if pontusx doesn't need it, I would skip

buberdds:
when I was thinking about slots I was under impression that customization is controlled by env variables. looking at their fork it's not true. Slots won't change anything because there is already a big problem with potential conflicts. Anyway let's discuss during sync meeting.

csillag:
As we have discussed this at the dev meeting, we no longer need this for now. So, closing. (We can revisit this if PontusX will want to have more custom links in the footer)

PR #1357, opened 2024-03-31
Network proposal votes: add search by validator name

csillag:
This adds the search feature for the network proposal vote list.

### DIFF
--- a/src/app/components/Search/TableSearchBar.tsx
+++ b/src/app/components/Search/TableSearchBar.tsx
@@ -6,7 +6,6 @@ import InputAdornment from '@mui/material/InputAdornment'
 import { COLORS } from '../../../styles/theme/colors'
 import IconButton from '@mui/material/IconButton'
 import { useScreenSize } from '../../hooks/useScreensize'
-import Box from '@mui/material/Box'
 import WarningIcon from '@mui/icons-material/WarningAmber'
 import { typingDelay } from '../../../styles/theme'
 import Typography from '@mui/material/Typography'
@@ -25,16 +24,16 @@ export interface TableSearchBarProps {
 export const TableSearchBar: FC<TableSearchBarProps> = ({ value, onChange, placeholder, warning }) => {
   const { isTablet } = useScreenSize()
 
-  const [isProblemFresh, setIsProblemFresh] = useState(false)
+  const [isWarningFresh, setIsWarningFresh] = useState(false)
 
   useEffect(() => {
     if (warning) {
       const timeout = setTimeout(() => {
-        setIsProblemFresh(false)
+        setIsWarningFresh(false)
       }, typingDelay)
       return () => clearTimeout(timeout)
     } else {
-      setIsProblemFresh(true)
+      setIsWarningFresh(true)
     }
   }, [warning])
 
@@ -61,7 +60,7 @@ export const TableSearchBar: FC<TableSearchBarProps> = ({ value, onChange, place
     </InputAdornment>
   )
 
-  const helperText = isProblemFresh ? undefined : (
+  const helperText = isWarningFresh ? undefined : (
     <Typography
       component="span"
       sx={{
--- a/src/app/components/Search/TableSearchBar.tsx
+++ b/src/app/components/Search/TableSearchBar.tsx
@@ -1,0 +1,134 @@
+import { FC, useEffect, useState } from 'react'
+import TextField from '@mui/material/TextField'
+import SearchIcon from '@mui/icons-material/Search'
+import HighlightOffIcon from '@mui/icons-material/HighlightOff'
+import InputAdornment from '@mui/material/InputAdornment'
+import { COLORS } from '../../../styles/theme/colors'
+import IconButton from '@mui/material/IconButton'
+import { useScreenSize } from '../../hooks/useScreensize'
+import Box from '@mui/material/Box'
+import WarningIcon from '@mui/icons-material/WarningAmber'
+import { typingDelay } from '../../../styles/theme'
+import Typography from '@mui/material/Typography'
+import { useTranslation } from 'react-i18next'
+import Button from '@mui/material/Button'
+import { CardEmptyState } from '../CardEmptyState'
+import { inputBaseClasses } from '@mui/material/InputBase'
+
+export interface TableSearchBarProps {
+  placeholder: string
+  warning?: string
+  value: string
+  onChange: (value: string) => void
+}
+
+export const TableSearchBar: FC<TableSearchBarProps> = ({ value, onChange, placeholder, warning }) => {
+  const { isTablet } = useScreenSize()
+
+  const [isProblemFresh, setIsProblemFresh] = useState(false)
+
+  useEffect(() => {
+    if (warning) {
+      const timeout = setTimeout(() => {
+        setIsProblemFresh(false)
+      }, typingDelay)
+      return () => clearTimeout(timeout)
+    } else {
+      setIsProblemFresh(true)
+    }
+  }, [warning])
+
+  const startAdornment = (
+    <InputAdornment
+      position="start"
+      disablePointerEvents // Pass clicks through, so it focuses the input
+    >
+      <SearchIcon sx={{ color: COLORS.grayDark, ml: -3 }} />
+    </InputAdornment>
+  )
+
+  const onClearValue = () => onChange('')
+
+  const endAdornment = (
+    <InputAdornment position="end">
+      {value ? (
+        <IconButton color="inherit" sx={{ mt: -3, mb: -3, mr: -1 }} onClick={onClearValue}>
+          <HighlightOffIcon />
+        </IconButton>
+      ) : (
+        <span style={{ width: '38px' }} />
+      )}
+    </InputAdornment>
+  )
+
+  const helperText = isProblemFresh ? undefined : (
+    <Typography
+      component="span"
+      sx={{
+        display: 'inline-flex',
+        color: COLORS.warningColor,
+        fontSize: 12,
+        lineHeight: 2,
+        alignItems: 'center',
+        verticalAlign: 'middle',
+        mt: 3,
+        mb: 4,
+        width: isTablet ? '160px' : undefined,
+      }}
+    >
+      <WarningIcon sx={{ mr: 3 }} />
+      {warning}
+    </Typography>
+  )
+
+  return (
+    <TextField
+      sx={{
+        backgroundColor: COLORS.white,
+        marginLeft: 4,
+        marginRight: helperText ? '25px' : '25px',
+        '&:focus-within': {
+          boxShadow: '3px 3px 3px 3px rgb(0, 0, 98, 0.25) !important',
+        },
+        [`.${inputBaseClasses.root}`]: {
+          border: '1px solid',
+          borderColor: COLORS.inactiveStroke,
+        },
+        ...(helperText
+          ? {
+              border: '1px solid',
+              borderColor: COLORS.inactiveStroke,
+              marginBottom: isTablet ? '-99px' : '-50px',
+            }
+          : {}),
+        zIndex: 10,
+      }}
+      variant={'outlined'}
+      value={value}
+      onChange={e => onChange(e.target.value)}
+      InputProps={{
+        inputProps: {
+          sx: {
+            p: 0,
+            width: isTablet ? 110 : 300,
+            margin: 2,
+          },
+        },
+        startAdornment,
+        endAdornment,
+      }}
+      placeholder={placeholder}
+      helperText={helperText}
+    />
+  )
+}
+
+export const NoMatchingDataMaybeClearFilters: FC<{ clearFilters: () => void }> = ({ clearFilters }) => {
+  const { t } = useTranslation()
+  const clearButton = (
+    <Button variant={'text'} onClick={() => clearFilters()}>
+      {t('tableSearch.clearFilters')}
+    </Button>
+  )
+  return <CardEmptyState label={t('tableSearch.noMatchingResults')} action={clearButton} />
+}
--- a/src/app/pages/ProposalDetailsPage/ProposalVotesCard.tsx
+++ b/src/app/pages/ProposalDetailsPage/ProposalVotesCard.tsx
@@ -6,13 +6,16 @@ import { TablePaginationProps } from '../../components/Table/TablePagination'
 import { useTranslation } from 'react-i18next'
 import { Table, TableCellAlign, TableColProps } from '../../components/Table'
 import { ExtendedVote, ProposalVoteValue } from '../../../types/vote'
-import { PAGE_SIZE, useAllVotes, useDisplayedVotes, useWantedVoteType } from './hooks'
+import { useVotes, useVoteFiltering } from './hooks'
 import { ProposalVoteIndicator } from '../../components/Proposals/ProposalVoteIndicator'
 import { DeferredValidatorLink } from '../../components/Validators/DeferredValidatorLink'
 import { CardHeaderWithResponsiveActions } from '../../components/CardHeaderWithResponsiveActions'
 import { VoteTypeFilter } from '../../components/Proposals/VoteTypeFilter'
 import { AppErrors } from '../../../types/errors'
 import { ErrorBoundary } from '../../components/ErrorBoundary'
+import Box from '@mui/material/Box'
+import { NoMatchingDataMaybeClearFilters, TableSearchBar } from '../../components/Search/TableSearchBar'
+import { CardEmptyState } from '../../components/CardEmptyState'
 
 type ProposalVotesProps = {
   isLoading: boolean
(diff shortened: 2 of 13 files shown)


src/app/hooks/useTypedSearchParam.ts
```diff
@@ -0,0 +1,44 @@
...
+  deleteParams?: string[]
+}
+
+type SetterFunction<T> = (value: T, options?: UrlSettingOptions) => void
+
+export function useTypedSearchParam<T = string>(
+  paramName: string,
+  defaultValue: T,
+  defaultSetterOptions: UrlSettingOptions = {},
+): [T, SetterFunction<T>] {
```
lukaw3d, line 30: Instead of `T = string`, lets do something like https://github.com/oasisprotocol/explorer/blob/61c87ee24567997f684ddf23314492b1e318e436/src/app/utils/storage.ts#L18 in the future

src/app/pages/ProposalDetailsPage/hooks.ts
```diff
@@ -0,0 +1,179 @@
...
+export const useVoterSearchPattern = () => {
+  const [voterSearchInput] = useVoterSearch()
+  return voterSearchInput.length < 3 ? undefined : voterSearchInput
+}
+
+export const useVoterSearchError = (t: TFunction) => {
+  const [input] = useVoterSearch()
+  const pattern = useVoterSearchPattern()
+  return !!input && !pattern ? t('tableSearch.error.tooShort') : undefined
+}
```
lukaw3d, line 136: This is bad design: developer basically can not know which of these 3 to use and what it does without reading the code of all three of these functions.

What would be your ideal interface / function shape for consuming this?

csillag: > developer basically can not know which of these 3 to use and what it does without reading the code of all three of these functions.

True.

> What would be your ideal interface / function shape for consuming this?

That's a non-trivial question. 
- Usually I prefer returning objects as opposed to arrays, because with field names there is no confusion, and also it's easier to selectively pick the field I want. 
- However, because or `React.useState()`, the `[value, setValue()]` pattern is very widespread.

Given all, that,
 - For ease of use, I kind of want to keep using this pattern for the obvious things (value and setter),
 - For anything,else, something else. Either individual hooks, or a bigger hook that provides everything in a object.

But how to have both side by side, and not be confusing? 

(Use longer names that exactly describes the functionality? Have only a single hook, with everything there, and let go of the ease of use of `[value, setValue()]` ?)

lubej:
Highlight part looks kind of weird, I was searching for "mainnet-upgrade" and it highlights half of "-" as well?

image

Should we add proposal example in the suggestion bar bellow the search bar:

image

src/app/components/Search/TableSearchBar.tsx
```diff
@@ -0,0 +1,135 @@
...
+  error?: string
+  value: string
+  onChange: (value: string) => void
+}
+
+export const TableSearchBar: FC<TableSearchBarProps> = ({ value, onChange, placeholder, error, warning }) => {
+  const { isTablet } = useScreenSize()
+
+  const [isProblemFresh, setIsProblemFresh] = useState(false)
+  const hasProblem = !!error || !!warning
```
lubej, line 28: Maybe it is just me, but I don't like `Problem` wording. Maybe `issue` or maybe even `exception`.

lubej:
Also search looks weird:

image

- shadow doesn't look right
- message bellow(its background) is somehow wider than the search box itself

csillag:
> Highlight part looks kind of weird, I was searching for "mainnet-upgrade" and it highlights half of "-" as well?

When we have first introduced the highlights, someone (probably @donouwens ?) has requested me to increase the padding, so we always highlight half a character extra, on both sides. It's all in the CSS.

Should we adjust it, @donouwens ?

lubej:
> > Highlight part looks kind of weird, I was searching for "mainnet-upgrade" and it highlights half of "-" as well?
> 
> When we have first introduced the highlights, someone (probably @donouwens ?) has requested me to increase the padding, so we always highlight half a character extra, on both sides. It's all in the CSS.
> 
> Should we adjust it, @donouwens ?

I mean, if it were consistent on the left side, it would make sense to me. But in my case it is not. That's why it looks inconsistent(left and right side).

csillag:
> Should we add proposal example in the suggestion bar bellow the search bar:

Good question. Technically it would make sense, but it's a very small niche, most users don't care about it. I think it would just draw focus away from more important things, like txs, accounts, blocks, etc.

But I defer this to @donouwens , too.

src/app/components/Search/TableSearchBar.tsx
csillag: Since we are not using the error prop for everything, I have just removed it, and then we don't even need this aggregated problem flag, because we can just check `!!warning`. So, removed it.

csillag:
> I mean, if it were consistent on the left side, it would make sense to me. But in my case it is not. That's why it looks inconsistent(left and right side).

Normally it's consistent. Searching for upgrade gets me this: 
kép

The problem is only on the beginning of the whole label. I guess the highlight can't expand there.

But in any case, this PR is about searching among the votes _within_ a given network change proposals, and not searching _for_ network change proposals.

So I suggest that we try to solve the generic issue of highlighting staff at the beginning of the labels separately.

csillag:
> Also search looks weird:
> 
> image
> 
>     * shadow doesn't look right
> 
>     * message bellow(its background) is somehow wider than the search box itself

This has been extensively reviewed by @donouwens . Everything has been adjusted as he requested. I am open to changes though, if there are any specific requests.

csillag:
> It does not seem logical to highlight sections that are not part of the search query

OK, reduced padding from 4px to 2px. 
Before: 
kép
After: 
kép

csillag:
> I would say the search bar here is self-explanatory. Users are 'deep' into the Explorer so we can reasonably expect them to know what they'd be searching for (unlike with the main search bar of the Explorer). My vote would be to not include suggestions.

@donouwens , I think @lubej was actually speaking about the main search bar of the Explorer. Right?

(However, Don, your interpretation is also understandable, since PR is about adding another search bar besides that one...)

csillag:
> The width and radius of the dropdown do not seem to match in the screenshot indeed. Could you match the width and radius to match the input field (visually), please?

OK, updated to: 
kép

csillag:
> We can add it, on desktop it's no problem, but mobile might make it tricky with the width (I would not want this to go over 2 lines tbh).

On further discussion, @donouwens 's verdict is:

> I [...] prefer to keep it out. [...]
> do not incorporate the suggestion in the dropdown

So I guess we don't need to do anything about this, for now.

src/app/components/Search/TableSearchBar.tsx
```diff
@@ -0,0 +1,133 @@
...
+      <SearchIcon sx={{ color: COLORS.grayDark, ml: -3 }} />
+    </InputAdornment>
+  )
+
+  const onClearValue = () => onChange('')
+
+  const endAdornment = (
+    <InputAdornment position="end">
+      {value ? (
+        <IconButton color="inherit" sx={{ mt: -3, mb: -3, mr: -1 }} onClick={onClearValue}>
```
lubej, line 54: Why do we need negative offset from top & bottom?

src/app/components/Search/TableSearchBar.tsx
```diff
@@ -0,0 +1,133 @@
...
+          <HighlightOffIcon />
+        </IconButton>
+      ) : (
+        <span style={{ width: '38px' }} />
```
lubej, line 58: Do we need empty span here? Can it not just be null?

src/app/components/Search/TableSearchBar.tsx
```diff
@@ -0,0 +1,133 @@
...
+        },
+        [`.${inputBaseClasses.root}`]: {
+          border: '1px solid',
+          borderColor: COLORS.inactiveStroke,
+        },
+        ...(helperText
+          ? {
+              border: '1px solid',
+              borderColor: COLORS.inactiveStroke,
+              marginBottom: isTablet ? '-99px' : '-50px',
```
lubej, line 100: Might look weird on devices with different DPI than yours.

src/app/utils/vote.ts
```diff
@@ -11,3 +12,6 @@ const voteFilters: Record<VoteType, VoteFilter> = {
 }
 
 export const getFilterForVoteType = (voteType: VoteType): VoteFilter => voteFilters[voteType]
+
+export const getFilterForVoterNameFragment = (fragment: string | undefined) =>
+  fragment ? (vote: ExtendedVote) => hasTextMatch(vote.validator?.media?.name, [fragment]) : () => true
```
lubej, line 17: Bit difficult to read.

Maybe:

```suggestion
export const getFilterForVoterNameFragment = (fragment: string | undefined) => {
   if(!fragment) { return () => true }
   return (vote: ExtendedVote) => hasTextMatch(vote.validator?.media?.name, [fragment])
 }
```

lubej (approved):
Apart from few small nits, LGTM

src/app/components/Search/TableSearchBar.tsx
```diff
@@ -0,0 +1,133 @@
...
+  useEffect(() => {
+    if (warning) {
+      const timeout = setTimeout(() => {
+        setIsProblemFresh(false)
+      }, typingDelay)
+      return () => clearTimeout(timeout)
+    } else {
+      setIsProblemFresh(true)
+    }
+  }, [warning])
```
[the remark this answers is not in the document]
lubej: I mean like, you would fire validation after debounce, and then show it straight away.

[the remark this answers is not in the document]
csillag: This has been copied verbatim from the existing big (main) search bar. When we added that, this has been reviewed, and various approaches tried... I wouldn't want to touch this again now, unless there is a real problem.

src/app/components/Search/TableSearchBar.tsx
lubej: No, that makes it inconsistent with what we are using(majority units defined in px). But yeah, we should have gone with `em` type units from the start.

src/app/components/Search/TableSearchBar.tsx
csillag: We want a placeholder, so that width of the input field stays the same, independently of the presence or lack of the `X` in the end.

src/app/components/Search/TableSearchBar.tsx
lubej: I mean, converting px -> em, does not belong in this PR. We can bring it up on next FE sync to discuss.

csillag: OK, I agree.

src/app/components/Search/TableSearchBar.tsx
lukaw3d: `const [isProblemFresh, setIsProblemFresh] = useState(false)`

src/app/components/Search/TableSearchBar.tsx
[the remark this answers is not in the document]
lukaw3d: Related https://github.com/oasisprotocol/explorer/pull/689#discussion_r1255676146

I still don't like this pattern of input validation

src/app/pages/ProposalDetailsPage/hooks.ts
```diff
@@ -116,28 +117,54 @@ export const useVoteStats = (network: Network, proposalId: number): VoteStats =>
   }
 }
 
-export const useWantedVoteType = () =>
-  useTypedSearchParam<VoteType>('vote', 'any', {
+export const useVoteFiltering = (t: TFunction) => {
```
lukaw3d, line 120: How come you're pushing t through hooks instead of using useTranslation? Is there some limitation?

src/app/pages/ProposalDetailsPage/ProposalVotesCard.tsx
```diff
@@ -74,41 +80,61 @@ const ProposalVotes: FC<ProposalVotesProps> = ({ isLoading, votes, rowsNumber, p
...
   const { network } = useRequiredScopeParam()
   const proposalId = parseInt(useParams().proposalId!, 10)
 
-  const { isLoading } = useAllVotes(network, proposalId)
-  const displayedVotes = useDisplayedVotes(network, proposalId)
+  const { clearFilters } = useVoteFiltering(t)
+  const { results, isLoading, isOnNonexistentPage, hasNoResultsBecauseOfFilters, hasNoResultsWhatsoever } =
+    useVotes(network, proposalId, t)
+
+  if (isOnNonexistentPage) throw AppErrors.PageDoesNotExist
```
lukaw3d, line 91: ("hasNoResultsOnSelectedPage" would match others better)

src/app/pages/ProposalDetailsPage/hooks.ts
csillag: There are no limitations that would apply here. We are sometimes doing this (passing `t` instead of _using_ it) in other situations, when the _rules of hooks_ might get in the way. (Ie. if the number of function calls might dynamically change while withing the render function if a given component.)  But it's not an issue here, so I have simplified the code..

At some point we could go over similar situations of the code, there might be other places that also can be simplified the same way.

src/app/pages/ProposalDetailsPage/hooks.ts
```diff
@@ -116,28 +117,54 @@ export const useVoteStats = (network: Network, proposalId: number): VoteStats =>
...
+  const [wantedType, setWantedType] = useTypedSearchParam<VoteType>('vote', 'any', {
     deleteParams: ['page'],
   })
+  const [wantedNameInput, setWantedNameInput] = useTypedSearchParam('voter', '', { deleteParams: ['page'] })
+  const wantedNamePattern = wantedNameInput.length < 3 ? undefined : wantedNameInput
+  const nameError = !!wantedNameInput && !wantedNamePattern ? t('tableSearch.error.tooShort') : undefined
+  const hasFilters = wantedType !== 'any' || !!wantedNamePattern
+  const clearFilters = () => {
+    setWantedType('any', { deleteParams: ['voter', 'page'] })
```
lukaw3d, line 130: This is a bit unintuitive: why is `vote` missing in `deleteParams: ['voter', 'page']`? It's because `setWantedType` sets `vote` to the default value and that deletes it.

I think I would choose:
```ts
  setWantedType('any')
  setWantedNameInput('')

// or

  setSearchParams((searchParams) => {
    searchParams.delete('vote')
    searchParams.delete('voter')
    searchParams.delete('page')
    return searchParams
  })
```

lukaw3d: Unexpectedly `setWantedType('assignsFirst-deletesSecond', { deleteParams: ['vote', 'voter', 'page'] })` would delete it too

src/app/components/Search/TableSearchBar.tsx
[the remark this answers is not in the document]
csillag: OK I will create an issue

lukaw3d (approved):

src/app/components/Search/TableSearchBar.tsx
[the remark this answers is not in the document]
csillag: See https://github.com/oasisprotocol/explorer/issues/1418.

src/app/pages/ProposalDetailsPage/ProposalVotesCard.tsx
csillag: OK renamed.

src/app/pages/ProposalDetailsPage/hooks.ts
csillag: > This is a bit unintuitive: why is `vote` missing in `deleteParams: ['voter', 'page']`? It's because `setWantedType` sets `vote` to the default value and that deletes it.

Exactly.

> 
> I think I would choose:

> ```ts
>   setWantedType('any')
>   setWantedNameInput('')

Yes, that is the natural choice, but unfortunately it doesn't work. I think the issue is with react router's implementation. Setting multiple search query parameter in quick succession causes race conditions.

 
> // or
> 
>   setSearchParams((searchParams) => {
>     searchParams.delete('vote')
>     searchParams.delete('voter')
>     searchParams.delete('page')
>     return searchParams
>   })
> ```

Yes, that would work. The problem is that if we are going that way, we can't reap any of the benefits of the already defined hook for working with the vote type.  By going around the existing hook, we are hurting encapsulation a little bit.

OTOH we are hurting it already (by manually deleting `page` and `voter`, so it doesn't really matter. I will go with this versino.

> Unexpectedly `setWantedType('assignsFirst-deletesSecond', { deleteParams: ['vote', 'voter', 'page'] })` would delete it too

Yes that works too, but I don't think it's any more intuitive to read.

csillag: Updated accordingly

src/app/components/Search/TableSearchBar.tsx
csillag: Renamed to use warning instead of problem.

PR #1358, opened 2024-03-31
Improve the interface of the comprehensive pagination engine

csillag:
Currently, we have two interfaces for pagination:
 - There is `SimplePaginationEngine`. One such engine is returned by `useSearchParamsPagination()`.
 - There is `ComprehensivePaginationEngine`. We have two implementations for this: 
   - `useComprehensiveSearchParamsPagination()` and
   - `useClientSidePagination()`

As you can guess from the name, the "Simple" engine is only used to provide parameters for queries, and driving the pagination ui. The "Comprehensive" engine has more responsibilities: it sits between the data source and the UI, completely encapsulating the data source, and therefore it is able to do various transformations (like sorting, filtering, etc).

**Currently we only using this comprehensive engine on the token dashboard (plus on some feature branches, mostly where we want client-side pagination), so it's easy to introduce breaking API changes.**

This PR polishes the interface of the comprehensive engine a little bit:
 * Adds lots of comments, explaining functionality
 * Renames some fields to better describe the actual meaning
 * Add some fields that are required for cleaner data encapsulation
 * Add support for arbitrary data transformation (besides the already existing filtering)

### DIFF
--- a/src/app/components/Table/useClientSidePagination.ts
+++ b/src/app/components/Table/useClientSidePagination.ts
@@ -1,14 +1,58 @@
 import { To, useSearchParams } from 'react-router-dom'
 import { AppErrors } from '../../../types/errors'
-import { ComprehensivePaginationEngine } from './PaginationEngine'
+import { ComprehensivePaginatedResults, ComprehensivePaginationEngine } from './PaginationEngine'
 import { List } from '../../../oasis-nexus/api'
 import { TablePaginationProps } from './TablePagination'
 
-type ClientSizePaginationParams<Item> = {
+type Filter<Item> = (item: Item) => boolean
+
+type ClientSizePaginationParams<Item, QueryResult extends List, ExtractedData = typeof undefined> = {
+  /**
+   * How should we call the query parameter (in the URL)?
+   */
   paramName: string
+
+  /**
+   * The pagination page size from the POV of the data consumer component
+   */
   clientPageSize: number
+
+  /**
+   * The pagination page size used for actually loading the data from the server.
+   *
+   * Please note that currently this engine doesn't handle when the data consumer requires data which is not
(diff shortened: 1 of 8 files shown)


src/app/components/Table/useClientSidePagination.ts
```diff
@@ -57,30 +109,51 @@ export function useClientSidePagination<Item, QueryResult extends List>({
     return { search: newSearchParams.toString() }
   }
 
-  const limit = serverPageSize
-  const offset = (selectedServerPage - 1) * clientPageSize
+  // From the server, we always want to load the first batch of data, with the provided (big) window.
+  // In theory, we could move this window as required, but currently this is not implemented.
+  const selectedServerPage = 1
```
lukaw3d, line 114: Is batch a different concept from window and page?

src/app/components/Table/useClientSidePagination.ts
```diff
@@ -57,30 +109,51 @@ export function useClientSidePagination<Item, QueryResult extends List>({
...
+
+      // The data window from the POV of the data consumer component
       const offset = (selectedClientPage - 1) * clientPageSize
       const limit = clientPageSize
       const dataWindow = filteredData ? filteredData.slice(offset, offset + limit) : undefined
 
+      // The control interface for the data consumer component (i.e. Table)
       const tableProps: TablePaginationProps = {
         selectedPage: selectedClientPage,
         linkToPage,
```
lukaw3d, line 159: Can you improve the comment below?
```
          ? // This correction is here to simulate the bogus behavior of server-side pagination
            // Remove this when server-side pagination is fixed,
            // and so the work-around in the pagination widget is fixed.
```

it doesn't say TODO, it doesn't say what's wrong, and repeats last line?

PR #1365, opened 2024-04-05
Simplify converting EVM addresses to oasis1

lukaw3d:


src/app/components/Search/search-utils.ts
```diff
@@ -121,27 +128,21 @@ export function isSearchValid(searchTerm: string) {
...
-  if (!!network && !RouteUtils.getEnabledNetworks().includes(network as Network)) {
+export const useParamSearch = () => {
+  const network = useNetworkParam()
+
+  if (!!network && !RouteUtils.getEnabledNetworks().includes(network)) {
     throw new AppError(AppErrors.InvalidUrl)
   }
 
-  const searchTerm = getSearchTermFromRequest(request)
+  const searchTerm = useSearchParams()[0].get('q')?.trim() ?? ''
```
lubej, line 138: Not sure I get this change, as `getSearchTermFromRequest` is still available.

src/routes.tsx
```diff
@@ -104,7 +103,6 @@ export const routes: RouteObject[] = [
             {
               path: '/search', // Global search
               element: withDefaultTheme(<SearchResultsPage />),
-              loader: searchParamLoader,
```
lubej, line 107: Why are we removing the `searchParamLoader` here? I would rather see that app fails in a loader if invalid URL, instead of in component.

src/app/components/Search/search-utils.ts
```diff
@@ -121,27 +128,21 @@ export function isSearchValid(searchTerm: string) {
...
   const normalized = Object.fromEntries(
     Object.entries(validateAndNormalize).map(([key, fn]) => [key, fn(searchTerm)]),
   ) as { [Key in keyof typeof validateAndNormalize]: string | undefined }
   return {
     searchTerm,
     ...normalized,
-    // TODO: remove conversion when API supports querying by EVM address
-    // TODO: without async conversion, this won't need to even be a loader
-    evmBech32Account: normalized.evmAccount ? await getEvmBech32Address(normalized.evmAccount) : undefined,
```
lubej, line 139: So EVM support querying was added to the API?

src/app/components/Search/search-utils.ts
lukaw3d: But request isn't available in a hook

src/routes.tsx
lukaw3d: I don't think searchParamLoader was validating the URL. We just needed to parse the query in URL, but async code forced us to use a loader (or useEffects)

src/app/components/Search/search-utils.ts
lukaw3d: I don't think so. I moved this code to its buddies at https://github.com/oasisprotocol/explorer/blob/14de628f5304bed3c53c5c193976072d38393195/src/app/components/Search/search-utils.ts#L105-L111

lukaw3d: Now that it isn't async

src/app/components/Search/search-utils.ts
lubej: Ah, this is a different function, overlooked rename `seachParamLoader` -> `useParamSearch`

lubej (approved):
LGTM, just fix the linting issues

src/routes.tsx
lubej: Should have linked to master, either way, this code was previously in `searchParamLoader`

lukaw3d: Checked more thoroughly

Before:
- /:_network/:_layer/search:
  - /:_network/:_layer/: assertEnabledScope validated network and scope
  - /:_network/:_layer/search: searchParamLoader validated network again
- /search: searchParamLoader skipped the validation because !!network is false

After:
- /:_network/:_layer/search:
  - /:_network/:_layer/: assertEnabledScope validates network and scope
- /search: nothing to validate

lukaw3d: (found https://github.com/oasisprotocol/explorer/pull/1366 while checking)

PR #1366, opened 2024-04-05
Fix root error element crashing "must be used within AnalyticsContext"

lukaw3d:
https://explorer.dev.oasis.io/err
crashes with "Error: must be used within AnalyticsContext"

http://localhost:1234/err
shows Not Found

### DIFF
--- a/src/routes.tsx
+++ b/src/routes.tsx
@@ -85,7 +85,11 @@ const RedirectToDashboard: FC = () => {
 
 export const routes: RouteObject[] = [
   {
-    errorElement: <RoutingErrorPage />,
+    errorElement: (
+      <AnalyticsConsentProvider>
+        <RoutingErrorPage />
+      </AnalyticsConsentProvider>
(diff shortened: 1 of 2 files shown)


src/routes.tsx
```diff
@@ -86,7 +86,11 @@ const RedirectToDashboard: FC = () => {
 
 export const routes: RouteObject[] = [
   {
-    errorElement: <RoutingErrorPage />,
+    errorElement: (
+      <AnalyticsConsentProvider>
+        <RoutingErrorPage />
+      </AnalyticsConsentProvider>
+    ),
```
lubej, line 93: Wouldn't this make more sense to be one level above `<RouterProvider />`?

csillag (approved):
Reviewed. LGTM.

src/routes.tsx
lukaw3d: Doesn't work.

Issue #1367, opened 2024-04-08
[Bug]: Gas fee paying address on Sapphire testnet showing wrong amount


lubej:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### SUMMARY

Correct balance not being reflected on Sapphire testnet gas paying address

### STEPS TO REPRODUCE

Check account [oasis1qp3r8hgsnphajmfzfuaa8fhjag7e0yt35cjxq0u4](https://explorer.dev.oasis.io/testnet/sapphire/address/oasis1qp3r8hgsnphajmfzfuaa8fhjag7e0yt35cjxq0u4), upon refreshing the page, you can replicate the negative balance on the account.

![image](https://github.com/oasisprotocol/explorer/assets/9722540/455c3209-057a-444e-85c7-c095822a3963)

### EXPECTED BEHAVIOR

Expected that the gas paying address `oasis1qp3r8hgsnphajmfzfuaa8fhjag7e0yt35cjxq0u4` would show the correct balance.

PR #1374, opened 2024-04-10
Make sure to clean RPC balance when required

csillag:
We always need to forget the previously loaded
RPC balance info, even when the newly
loaded account doesn't have a new RPC balance.

Fixes #1369.

PR #1073, opened 2023-12-06
Fetch account balance from RPC node

lubej:
Account that this can be tested on: `<explorer-url>/testnet/sapphire/address/0x2B95834655499B3C3817f379e1C8E132baa965F4`

PR #1384, opened 2024-04-15
Fetch balances from Oasis gRPC instead of Web3 RPC

lukaw3d:
Fixes https://github.com/oasisprotocol/explorer/issues/1367

Differences in behaviour:
- Fetches all balances, not just the native one
- Fetches balances for contracts
- Fetches balances for all accounts, not just EVM ones
- Starts the request in parallel with Nexus API call

Related to https://github.com/oasisprotocol/explorer/pull/1073 https://github.com/oasisprotocol/explorer/pull/1338 https://github.com/oasisprotocol/explorer/pull/1374


buberdds (approved):

src/app/utils/getRPCAccountBalances.ts
```diff
@@ -0,0 +1,47 @@
...
+): Promise<RuntimeSdkBalance[]> {
+  const paratimeConfig = paraTimesConfig[scope.layer][scope.network]
+  if (!paratimeConfig.runtimeId) throw new Error('Paratime is not configured')
+
+  const nic = new oasis.client.NodeInternal(grpcUrls[scope.network])
+  const accountsWrapper = new oasisRT.accounts.Wrapper(oasis.misc.fromHex(paratimeConfig.runtimeId))
+
+  const rawBalances = (
+    await accountsWrapper
+      .queryBalances()
```
lubej, line 28: I think native balance querying should be enough, as ERC-20 tokens for example are showing correct balances. But since it is going to be removed in the future, I guess it doesn't matter that much.

src/app/utils/getRPCAccountBalances.ts
```diff
@@ -0,0 +1,47 @@
...
+      .setArgs({ address: oasis.staking.addressFromBech32(oasisAddress) })
+      .query(nic)
+  ).balances
+
+  return [...rawBalances.entries()]
+    .map(([denomination, amount]) => {
+      return {
+        balance: oasis.quantity.toBigInt(amount).toString(),
```
lubej, line 36: Why is conversion to bigint and back to string necessary?

src/app/utils/getRPCAccountBalances.ts
```diff
@@ -0,0 +1,47 @@
...
+        token_symbol: oasis.misc.toStringUTF8(denomination) || getTokensForScope(scope)[0].ticker,
+        token_decimals: paraTimesConfig[scope.layer].decimals, // TODO: differentiate by token
```
lubej, line 38: That doesn't look right, if this is querying ERC-20 tokens as well, each of the ERC-20 token, can have custom decimals specified. So defaulting to paratime decimals here, is wrong.

src/app/utils/getRPCAccountBalances.ts
```diff
@@ -0,0 +1,47 @@
...
+      }
+    })
+    .map(token => {
+      return {
+        ...token,
+        balance: fromBaseUnits(token.balance, token.token_decimals),
+      }
+    })
```
lubej, line 46: This is iterating twice over an array, for no good reason. This should happen in the previous `map`.

src/app/utils/getRPCAccountBalances.ts
lukaw3d: It's from Uint8Array to bigint to string

src/app/utils/getRPCAccountBalances.ts
lukaw3d: It's only querying e.g. TEST and EUROe; evm_balances are a different field in Nexus API.
But we don't differentiate decimals between TEST and EUROe - related: https://github.com/oasisprotocol/explorer/pull/1265#discussion_r1498078526

src/app/utils/getRPCAccountBalances.ts
lukaw3d: I prefer this, and it mirrors https://github.com/oasisprotocol/explorer/blob/26079807ff9a4e91cc42d106d75a0c824533caba/src/oasis-nexus/api.ts#L361-L366. 

(I also think https://github.com/oasisprotocol/explorer-frontend/blob/26079807ff9a4e91cc42d106d75a0c824533caba/src/oasis-nexus/api.ts#L709-L752 should be split into multiple iterations over an array)

lubej: I know it is not a large array, but for me it makes no sense, to iterate over the same array multiple times.

src/app/utils/getRPCAccountBalances.ts
lubej: Ok, I verified that both Ocean and EUROe use 18 decimals places, so coincidentally this works as expected.

Looking at the documentation at https://docs.pontus-x.eu/docs/Development/quick_start#setup-metamask the native token has been renamed to GX @csillag are we aware of this?

PR #1389, opened 2024-04-24
Migrate to Storybook 8

buberdds:
- mockService is autogenerated file 
- one of our test relies on Storybook 7 dep, so needed to add node fetch types to our package.json

side note: Storybook does not support Parcel builds, so moving to Vite will reduce node_module
```
https://github.com/storybookjs/storybook/tree/next/code/frameworks
https://github.com/parcel-bundler/parcel/issues/8668
```


lukaw3d (approved):

lubej:
> side note: Storybook does not support Parcel builds, so moving to Vite will reduce node_module

But Storybook relies on Webpack? At least on first glance at the config. Don't see a connection with Vite here, unless it supports vite plugin.

PR #1396, opened 2024-04-30
Add dynamically resizing label able to fill up horizontal space

csillag:
Improve how we use horizontal space. When working with small-ish vertical space, instead of always shortening text to a very short predefined value, try to fill up the available space.

Changes cleared by @donouwens (a while back).

See https://github.com/oasisprotocol/explorer/pull/1290 for the background and context.

This implementation is using JS code for the adaptive dynamic resizing.
We might want to use a pure-CSS implementation instead if this, for simplicity.

### DIFF
--- a/src/app/components/Account/AccountLink.tsx
+++ b/src/app/components/Account/AccountLink.tsx
@@ -14,7 +14,8 @@ const WithTypographyAndLink: FC<{
   to: string
   mobile?: boolean
   children: ReactNode
-}> = ({ children, to, mobile }) => {
+  labelOnly?: boolean
+}> = ({ children, to, mobile, labelOnly }) => {
   return (
     <Typography
       variant="mono"
@@ -28,9 +29,13 @@ const WithTypographyAndLink: FC<{
           : {}),
       }}
     >
-      <Link component={RouterLink} to={to}>
-        {children}
-      </Link>
+      {labelOnly ? (
+        children
+      ) : (
+        <Link component={RouterLink} to={to}>
+          {children}
+        </Link>
+      )}
     </Typography>
   )
 }
@@ -50,9 +55,16 @@ interface Props {
    * (Besides the content necessary because of potential shortening)
    */
   extraTooltip?: ReactNode
+
+  /**
+   * Should we display this as a simple label, with no link?
+   *
+   * (Used for own address)
+   */
+  labelOnly?: boolean
 }
 
-export const AccountLink: FC<Props> = ({ scope, address, alwaysTrim, extraTooltip }) => {
+export const AccountLink: FC<Props> = ({ scope, address, alwaysTrim, extraTooltip, labelOnly }) => {
   const { isTablet } = useScreenSize()
   const to = RouteUtils.getAccountRoute(scope, address)
 
@@ -68,7 +80,7 @@ export const AccountLink: FC<Props> = ({ scope, address, alwaysTrim, extraToolti
     // In a table, we only ever want a short line
 
     return (
-      <WithTypographyAndLink to={to}>
+      <WithTypographyAndLink to={to} labelOnly={labelOnly}>
         <MaybeWithTooltip title={address}>{trimLongString(address, 6, 6)}</MaybeWithTooltip>
       </WithTypographyAndLink>
     )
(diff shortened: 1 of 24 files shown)


csillag:
(Code extracted from the Pontus-X branch)

src/app/components/Transactions/ConsensusTransactions.tsx
```diff
@@ -94,19 +91,7 @@ export const ConsensusTransactions: FC<ConsensusTransactionsProps> = ({
                     pr: 3,
                   }}
                 >
-                  {!!ownAddress && transaction.sender === ownAddress ? (
```
lukaw3d, line 97: Is this intentional / any idea how to re-implement marking the current address?

https://93f0c866.oasis-explorer.pages.dev/mainnet/consensus/address/oasis1qp533xjqg7rjn9xglk6fgp3pykdscsk6vs3m2tuz

| Before | After |
| --- | --- |
|   explorer dev oasis io_mainnet_consensus_address_oasis1qp533xjqg7rjn9xglk6fgp3pykdscsk6vs3m2tuz     |    93f0c866 oasis-explorer pages dev_mainnet_consensus_address_oasis1qp533xjqg7rjn9xglk6fgp3pykdscsk6vs3m2tuz    |

csillag: Not intentional on my part. I have a vague memory seeing this chance coming in from somewhere else, I being surprised, but I can't recall the specifics. I'll just re-introduce this now.

csillag: Restored in https://74d1e081.oasis-explorer.pages.dev/mainnet/consensus/address/oasis1qp533xjqg7rjn9xglk6fgp3pykdscsk6vs3m2tuz

lukaw3d: Only fixed consensus

csillag: Oops. Now fixed on paratimes, too.

lukaw3d (approved):

src/app/components/Account/AccountLink.tsx
```diff
@@ -1,30 +1,107 @@
...
+   */
+  labelOnly?: boolean
 }
 
-export const AccountLink: FC<Props> = ({ scope, address, alwaysTrim }) => {
+export const AccountLink: FC<Props> = ({ scope, address, alwaysTrim, extraTooltip, labelOnly }) => {
   const { isTablet } = useScreenSize()
   const to = RouteUtils.getAccountRoute(scope, address)
+
+  const tooltipPostfix = extraTooltip ? (
```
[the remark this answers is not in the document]
csillag: Renamed accordingly

src/app/components/Account/AccountLink.tsx
```diff
@@ -1,30 +1,107 @@
...
+  /**
+   * Should we display this as a simple label, with no link?
+   *
+   * (Used for own address)
+   */
+  labelOnly?: boolean
 }
 
-export const AccountLink: FC<Props> = ({ scope, address, alwaysTrim }) => {
+export const AccountLink: FC<Props> = ({ scope, address, alwaysTrim, extraTooltip, labelOnly }) => {
```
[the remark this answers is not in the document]
csillag: The problem with the `from` and `to` fields is that the auto-trimming only works if  no overflow / scrolling is happening in the DOM _above_ the component that is displaying the address. If the overflow is happening in a higher level, then we can't detect it in the adaptive trimmer.

The way to ensure this is to go through the dom level by level, and boxing everything with `max-width` or similar CSS features. However, here we are inside a `table`, which doesn't support `max-width`, as far as I know.

On the short term, I added a switch called `alwaysTrimOnTable`, which we can set when using the `AccountLink` inside a table (or similar environments) to disable the adaptive behavior, and fall back to fixed length. 

This gets us: 
image

Trimming too much is still better then hiding the stuff in an overflow, I guess.

This could be improved, but I would rather return to this after the new event-related stuff has landed.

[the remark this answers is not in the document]
csillag: > * "Emitting contract" is misbehaving a bit

That one, we can solve easily.

image

src/app/components/AdaptiveTrimmer/MaybeWithTooltip.tsx
```diff
@@ -0,0 +1,65 @@
...
+  tooltipWanted = true,
+  title,
+  children,
+  spanSx,
+}) =>
+  tooltipWanted && !!title ? (
+    <Tooltip
+      placement="top"
+      title={
+        <Box
```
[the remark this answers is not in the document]
csillag: Simplified accordingly

PR #1397, opened 2024-04-30
Add dynamically resizing label with highlight, for adaptively  filling up horizontal space

csillag:
## This PR adds:

 - **Level 1**: `cutAroundMatch()`: utility function for getting a fixed-length a part of a text string which contains the match to the pattern (if any)
 - **Level 2**: `HighlightedTrimmedText`: component that takes a text and a search pattern, and displays the text, with the matching part (if any) highlighted, the whole thing trimmed to a specific length around the highlight. (Or the beginning if there is no match.) This uses `cutAroundMatch()`.
- **Utility**: `AdaptiveDynamicTrimmer`: a component which can display content, dynamically filling up the available space until scrolling/overflow happens. The dynamic part in name references to the fact that it expects a function prop to provide a version of the content shortened for any specific length. This function will be called with various length, to find the best version of the content that fits. Note that this function can do smart things to get the best representation for the given length, which might be more complex than simply getting the beginning, the end or the middle.
- **Level 3**: `AdaptiveHighlightedText`: a component that displays text with a part highlighted, adaptively trimmed around the highlight to fit the maximum available length without overflow. This uses `HighlightedTrimmedText` and `AdaptiveDynamicTrimmer`.
 
## Changes:
- Teach `TrimEndLinkLabel` component to support highlighting parts within the labels, and show the relevant part instead of always showing the beginning of the label.

## Testing

None of these changes are immediately visible on the UI. To see this in action, check out some other PRs that use this:
 - #1357, which uses this for displaying votes for network proposals (with highlights for search results). This showcases Level 2: No adaptive expansion, just smart trimming around the highlight. - [example](https://ec1a5160.oasis-explorer.pages.dev/mainnet/consensus/proposal/3?voter=ell) (check with 630px width)
 - #1398, which uses this for displaying account names (with highlights for search results). Showcases Level 3 - [example](https://1089a21b.oasis-explorer.pages.dev/search?q=tor) (check with 630px width)


csillag:
Code extracted from the Pontus-X branch

src/app/components/HighlightedText/text-cutting.ts
```diff
@@ -0,0 +1,74 @@
...
+  // We will start before the start of the match, by buffer / 2 chars
+  const startPos = Math.max(Math.min(matchStart - Math.floor(buffer / 2), corpus.length - fragmentLength), 0)
+  const endPos = Math.min(startPos + fragmentLength, corpus.length)
+
+  // compile the result
+  const part =
+    (startPos ? '…' : '') + corpus.substring(startPos, endPos) + (endPos < corpus.length - 1 ? '…' : '')
+
+  return {
+    hasMatch: true,
```
lukaw3d, line 71: Why is this true

src/app/components/HighlightedText/text-cutting.ts
```diff
@@ -0,0 +1,74 @@
...
+ * an adequately sized part from the beginning of the corpus is returned.
+ *
+ * If there is a match, but the corpus is at most as long as the desired fragment length,
+ * the whole corpus is returned.
+ *
+ * If there is a match, and the corpus is longer than the desired fragment length,
+ * then a part of a corpus is returned, so that the match is within the returned part,
+ * around the middle.
+ */
+export function cutAroundMatch(
```
lukaw3d, line 28: Rename to "trim"

src/app/components/HighlightedText/text-cutting.ts
```diff
@@ -0,0 +1,74 @@
...
+  if (!corpus) {
+    // there is nothing to see here
+    return {
+      hasMatch: false,
+      part: undefined,
+    }
+  }
+
+  // do we have a match?
+  const match = pattern ? findTextMatch(corpus, [pattern], matchOptions) : undefined
```
lukaw3d, line 47: searching by capitalized pattern is broken (based on the example in PR description):
- https://ec1a5160.oasis-explorer.pages.dev/mainnet/consensus/proposal/3?voter=bam
  image
- https://ec1a5160.oasis-explorer.pages.dev/mainnet/consensus/proposal/3?voter=Bam
  image

src/app/components/HighlightedText/text-cutting.ts
```diff
@@ -0,0 +1,74 @@
...
+  }
+
+  // how much extra space do we have?
+  const buffer = fragmentLength - (pattern || '').length
+
+  const matchStart = match?.startPos ?? 0
+
+  // We will start before the start of the match, by buffer / 2 chars
+  const startPos = Math.max(Math.min(matchStart - Math.floor(buffer / 2), corpus.length - fragmentLength), 0)
+  const endPos = Math.min(startPos + fragmentLength, corpus.length)
```
lukaw3d, line 64: Something doesn't behave well if match is longer than fragmentLength:
image

src/app/components/HighlightedText/AdaptiveHighlightedText.tsx
```diff
@@ -0,0 +1,66 @@
...
+          text={text}
+          pattern={pattern}
+          options={options}
+        />
+      )}
+      extraTooltip={
+        extraTooltip ? (
+          <>
+            <InfoIcon />
+            {extraTooltip}
```
lukaw3d, line 60: It's confusing that AdaptiveHighlightedText's `extraTooltip` prepends the icon and AdaptiveDynamicTrimmer's `extraTooltip` doesn't.

lukaw3d (approved):
those things can be fixed as separate PR, if it adds too many conflicts

src/app/components/AdaptiveTrimmer/AdaptiveDynamicTrimmer.tsx
```diff
@@ -0,0 +1,151 @@
...
+}
+
+/**
+ * Display content, potentially shortened as needed.
+ *
+ * This component will do automatic detection of available space,
+ * and determine the best way to display content accordingly.
+ *
+ * The difference compared to AdaptiveTrimmer is that this component
+ * expects a function to provide a shortened version of the components.
```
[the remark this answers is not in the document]
csillag: >  Why isn't AdaptiveTrimmer calling AdaptiveDynamicTrimmer?

I considered doing that, but I didn't want to achieve the more simple thing by relying on the more complex implementation. But I'm OK with that approach, let's see...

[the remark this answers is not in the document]
csillag: > (in this case the first commit should be renaming AdaptiveTrimmer into AdaptiveDynamicTrimmer, then modifying AdaptiveDynamicTrimmer and creating a new wrapper named AdaptiveTrimmer)

Done.

src/app/components/HighlightedText/text-cutting.ts
csillag: Done

src/app/components/HighlightedText/text-cutting.ts
csillag: Actually we don't know if it's true. It should have been `!!match`. On the other hand we are not using this part of the result anywhere, so I have now removed `hasMatch` completely.

src/app/components/HighlightedText/text-cutting.ts
csillag: Good catch.

Here, the problem was that the referenced example PR didn't use the correct routine (`findTextMatch()`) to look for the match. (It _was_ using the highlighting functionality from here.)

Fixed this on the other PR by calling the correct function.

csillag:
> > [example](https://ec1a5160.oasis-explorer.pages.dev/mainnet/consensus/proposal/3?voter=ell) (check with 630px width)
> 
> ec1a5160 oasis-explorer pages dev_mainnet_consensus_proposal_3_voter=ell (1) 
>
> @donouwens is this the expected behavior

You mean the fact that the highlight has some margin, so we are selecting half a character to the left and to the right?
If that's what you mean, then yes, I think this is what we wanted. This is also why we made the yellow highlight class slightly transparent, so that it's not an issue that we partially cover other characters.  (But @donouwens can confirm/correct.)

src/app/components/HighlightedText/text-cutting.ts
csillag: Yeah, another food catch. Here, the problem was that the no information has been passed between the trimming function and the highlighting function. First the trimming happened, and then the highlighter independently tried to identify the part to highlight. But since the pattern is longer than the trimmed text, in this case the highlighter can't find the match.

I solved this by channeling the original match info from the trimmer function to the highlighter.

kép

src/app/components/HighlightedText/AdaptiveHighlightedText.tsx
csillag: True. Will think about this later.

PR #1398, opened 2024-04-30
Add named accounts

csillag:
- Load metadata about Oasis accounts (available via Nexus).
- Load metadata about Pontus-X accounts 
- Displaying the names
- Support searching for the names.

Some accounts to test with are listed [here](https://github.com/oasisprotocol/nexus/blob/main/account-names/mainnet_paratime.json).

### DIFF
--- a/src/app/components/Account/AccountLink.tsx
+++ b/src/app/components/Account/AccountLink.tsx
@@ -6,8 +6,12 @@ import { RouteUtils } from '../../utils/route-utils'
 import InfoIcon from '@mui/icons-material/Info'
 import Typography from '@mui/material/Typography'
 import { SearchScope } from '../../../types/searchScope'
+import { useAccountMetadata } from '../../hooks/useAccountMetadata'
 import { trimLongString } from '../../utils/trimLongString'
 import { MaybeWithTooltip } from '../AdaptiveTrimmer/MaybeWithTooltip'
+import Box from '@mui/material/Box'
+import { HighlightedText } from '../HighlightedText'
+import { AdaptiveHighlightedText } from '../HighlightedText/AdaptiveHighlightedText'
 import { AdaptiveTrimmer } from '../AdaptiveTrimmer/AdaptiveTrimmer'
 
 const WithTypographyAndLink: FC<{
@@ -24,7 +28,7 @@ const WithTypographyAndLink: FC<{
         ...(mobile
           ? {
               maxWidth: '100%',
-              overflowX: 'hidden',
+              overflow: 'hidden',
             }
           : {}),
       }}
@@ -52,7 +56,12 @@ interface Props {
   /**
    * Should we always trim the text to a short line when on mobile or Tablet?
    */
-  alwaysTrimOnTable?: boolean
+  alwaysTrimOnTablet?: boolean
+
+  /**
+   * What part of the name should be highlighted (if any)
+   */
+  highlightedPartOfName?: string | undefined
 
   /**
    * Any extra tooltips to display
@@ -73,47 +82,87 @@ export const AccountLink: FC<Props> = ({
   scope,
   address,
   alwaysTrim,
-  alwaysTrimOnTable,
+  alwaysTrimOnTablet,
+  highlightedPartOfName,
   extraTooltip,
   labelOnly,
 }) => {
   const { isTablet } = useScreenSize()
+  const {
+    metadata: accountMetadata,
+    // isError, // Use this to indicate that we have failed to load the name for this account
+  } = useAccountMetadata(scope, address)
+  const accountName = accountMetadata?.name // TODO: we should also use the description
+
   const to = RouteUtils.getAccountRoute(scope, address)
 
   const extraTooltipWithIcon = extraTooltip ? (
-    <>
+    <Box
+      sx={{
+        display: 'inline-flex',
+        alignItems: 'center',
+        verticalAlign: 'middle',
+        gap: 2,
+      }}
+    >
       <InfoIcon />
       {extraTooltip}
-    </>
+    </Box>
   ) : undefined
 
   // Are we in a situation when we should always trim?
-  if (alwaysTrim || (alwaysTrimOnTable && isTablet)) {
+  if (alwaysTrim || (alwaysTrimOnTablet && isTablet)) {
     // In a table, we only ever want a short line
 
     return (
       <WithTypographyAndLink to={to} labelOnly={labelOnly}>
-        <MaybeWithTooltip title={address}>{trimLongString(address, 6, 6)}</MaybeWithTooltip>
+        <MaybeWithTooltip
+          title={
+            <div>
+              {accountName && <Box sx={{ fontWeight: 'bold' }}>{accountName}</Box>}
+              <Box sx={{ fontWeight: 'normal' }}>{address}</Box>
+              {extraTooltipWithIcon}
+            </div>
+          }
+        >
+          {accountName ? trimLongString(accountName, 12, 0) : trimLongString(address, 6, 6)}
+        </MaybeWithTooltip>
       </WithTypographyAndLink>
     )
   }
 
   if (!isTablet) {
     // Details in desktop mode.
-    // We want one long line
+    // We want one long line, with name and address.
 
     return (
       <WithTypographyAndLink to={to} labelOnly={labelOnly}>
-        <MaybeWithTooltip title={extraTooltipWithIcon}>{address} </MaybeWithTooltip>
+        <MaybeWithTooltip title={extraTooltipWithIcon}>
+          {accountName ? (
+            <span>
+              <HighlightedText text={accountName} pattern={highlightedPartOfName} /> ({address})
+            </span>
+          ) : (
+            address
+          )}
+        </MaybeWithTooltip>
       </WithTypographyAndLink>
     )
   }
 
   // We need to show the data in details mode on mobile.
-  // Line adaptively shortened to fill available space
+  // We want two lines, one for name (if available), one for address
+  // Both lines adaptively shortened to fill available space
   return (
     <WithTypographyAndLink to={to} mobile labelOnly={labelOnly}>
-      <AdaptiveTrimmer text={address} strategy="middle" extraTooltip={extraTooltip} />
+      <>
+        <AdaptiveHighlightedText
+          text={accountName}
+          pattern={highlightedPartOfName}
+          extraTooltip={extraTooltip}
+        />
+        <AdaptiveTrimmer text={address} strategy="middle" extraTooltip={extraTooltip} />
+      </>
     </WithTypographyAndLink>
   )
 }
--- a/src/app/data/named-accounts.ts
+++ b/src/app/data/named-accounts.ts
@@ -1,0 +1,51 @@
+import { Network } from '../../types/network'
+import { Account, Layer, Runtime, RuntimeAccount } from '../../oasis-nexus/api'
+
+export type AccountMetadata = {
+  address: string
+  name?: string
+  description?: string
+}
+
+export type AccountMetadataInfo = {
+  metadata?: AccountMetadata
+  isLoading: boolean
+  isError: boolean
+}
+
+export type AccountMap = Map<string, AccountMetadata>
+
+export type AccountData = {
+  map: AccountMap
+  list: AccountMetadata[]
+}
+
+export type AccountNameSearchMatch = {
+  network: Network
+  layer: Layer
+  address: string
+}
+
+export type AccountNameSearchRuntimeMatch = {
+  network: Network
+  layer: Runtime
+  address: string
+}
+
+export type AccountNameSearchConsensusMatch = {
+  network: Network
+  layer: typeof Layer.consensus
+  address: string
+}
+
+export type AccountNameSearchResults = {
+  results: (Account | RuntimeAccount)[] | undefined
+  isLoading: boolean
+  isError: boolean
+}
+
+export type AccountNameSearchRuntimeResults = {
+  results: RuntimeAccount[] | undefined
+  isLoading: boolean
+  isError: boolean
+}
--- a/src/app/data/oasis-account-names.ts
+++ b/src/app/data/oasis-account-names.ts
@@ -1,0 +1,147 @@
+import axios from 'axios'
+import { useQuery } from '@tanstack/react-query'
+import {
+  Layer,
+  useGetConsensusAccountsAddresses,
+  useGetRuntimeAccountsAddresses,
+} from '../../oasis-nexus/api'
+import { Network } from '../../types/network'
+import {
+  AccountData,
+  AccountMap,
+  AccountMetadata,
+  AccountMetadataInfo,
+  AccountNameSearchConsensusMatch,
+  AccountNameSearchMatch,
+  AccountNameSearchResults,
+  AccountNameSearchRuntimeMatch,
+} from './named-accounts'
+import { hasTextMatch } from '../components/HighlightedText/text-matching'
+
+const dataSources: Record<Network, Partial<Record<Layer, string>>> = {
+  [Network.mainnet]: {
+    [Layer.consensus]:
+      'https://raw.githubusercontent.com/oasisprotocol/nexus/main/account-names/mainnet_consensus.json',
+    [Layer.emerald]:
+      'https://raw.githubusercontent.com/oasisprotocol/nexus/main/account-names/mainnet_paratime.json',
+    [Layer.sapphire]:
+      'https://raw.githubusercontent.com/oasisprotocol/nexus/main/account-names/mainnet_paratime.json',
+  },
+  [Network.testnet]: {
+    [Layer.consensus]:
+      'https://raw.githubusercontent.com/oasisprotocol/nexus/main/account-names/testnet_consensus.json',
+    [Layer.emerald]:
+      'https://raw.githubusercontent.com/oasisprotocol/nexus/main/account-names/testnet_paratime.json',
+    [Layer.sapphire]:
+      'https://raw.githubusercontent.com/oasisprotocol/nexus/main/account-names/testnet_paratime.json',
+  },
+}
+
+const getOasisAccountsMetadata = async (network: Network, layer: Layer): Promise<AccountData> => {
+  const url = dataSources[network][layer]
+  if (!url) throw new Error('No data available for this layer')
+  const response = await axios.get(url)
+  if (response.status !== 200) throw new Error("Couldn't load names")
+  if (!response.data) throw new Error("Couldn't load names")
+  // console.log('Response data is', response.data)
+  const map: AccountMap = new Map()
+  const list: AccountMetadata[] = []
+  Array.from(response.data).forEach((entry: any) => {
+    const metadata: AccountMetadata = {
+      address: entry.Address,
+      name: entry.Name,
+      description: entry.Description,
+    }
+    // Register the metadata in its native form
+    list.push(metadata)
+    map.set(metadata.address, metadata)
+  })
+  return {
+    map,
+    list,
+  }
+}
+
+export const useOasisAccountsMetadata = (
+  network: Network,
+  layer: Layer,
+  queryOptions: { enabled: boolean },
+) => {
+  return useQuery(['oasisAccounts', network, layer], () => getOasisAccountsMetadata(network, layer), {
+    enabled: queryOptions.enabled,
+    staleTime: Infinity,
+  })
+}
+
+export const useOasisAccountMetadata = (
+  network: Network,
+  layer: Layer,
+  address: string,
+  queryOptions: { enabled: boolean },
+): AccountMetadataInfo => {
+  const { isLoading, isError, error, data: allData } = useOasisAccountsMetadata(network, layer, queryOptions)
+  if (isError) {
+    console.log('Failed to load Oasis account metadata', error)
+  }
+  return {
+    metadata: allData?.map.get(address),
+    isLoading,
+    isError,
+  }
+}
+
+export const useSearchForOasisAccountsByName = (
+  network: Network,
+  layer: Layer,
+  nameFragment: string,
+  queryOptions: { enabled: boolean },
+): AccountNameSearchResults => {
+  const {
+    isLoading: isMetadataLoading,
+    isError: isMetadataError,
+    error: metadataError,
+    data: namedAccounts,
+  } = useOasisAccountsMetadata(network, layer, queryOptions)
+  if (isMetadataError) {
+    console.log('Failed to load Oasis account metadata', metadataError)
+  }
+
+  const textMatcher =
+    nameFragment && queryOptions.enabled
+      ? (account: AccountMetadata) => hasTextMatch(account.name, [nameFragment])
+      : () => false
+
+  const matches =
+    namedAccounts?.list.filter(textMatcher).map(
+      (account): AccountNameSearchMatch => ({
+        network,
+        layer,
+        address: account.address,
+      }),
+    ) ?? []
+
+  const consensusMatches = layer === Layer.consensus ? (matches as AccountNameSearchConsensusMatch[]) : []
+  const runtimeMatches = layer === Layer.consensus ? [] : (matches as AccountNameSearchRuntimeMatch[])
+
+  const {
+    isLoading: areConsensusAccountsLoading,
+    isError: areConsensusAccountsError,
+    data: consensusResults,
+  } = useGetConsensusAccountsAddresses(consensusMatches, {
+    enabled: queryOptions.enabled && !isMetadataLoading && !isMetadataError,
+  })
+
+  const {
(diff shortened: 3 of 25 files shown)


src/app/data/oasis-account-names.ts
```diff
@@ -0,0 +1,120 @@
...
+): AccountMetadataInfo => {
+  // When running jest tests, we don't want to load from Pontus-X.
+  if (process.env.NODE_ENV === 'test') {
+    return {
+      metadata: {
+        name: undefined,
+      },
+      loading: false,
+    }
+  }
```
lubej, line 82: This would have been better if it was mocked in tests(`useOasisAccountMetadata` function), instead of being bundled to production code.

src/app/data/pontusx-account-names.ts
```diff
@@ -0,0 +1,84 @@
...
+}
+
+export const usePontusXAccountMetadata = (address: string, enabled: boolean): AccountMetadataInfo => {
+  // When running jest tests, we don't want to load from Pontus-X.
+  if (process.env.NODE_ENV === 'test') {
+    return {
+      metadata: undefined,
+      loading: false,
+    }
+  }
```
lubej, line 47: Same here. Mock globally in tests.

src/app/data/oasis-account-names.ts
csillag: OK, removed mocking from production code.

src/app/data/pontusx-account-names.ts
csillag: OK, removed mocking from production code.

src/app/data/oasis-account-names.ts
```diff
@@ -0,0 +1,110 @@
...
+
+  const textMatcher =
+    nameFragment && enabled
+      ? (entry: AccountEntry): boolean => {
+          return hasTextMatch(entry.name, [nameFragment])
+        }
+      : () => false
+  return {
+    results: (allData?.list || []).filter(textMatcher).map(entry => ({
```
lukaw3d, line 103: This is returning an empty array while loading and while disabled.
Related https://github.com/oasisprotocol/explorer/pull/1356#discussion_r1547059928

src/app/pages/SearchResultsPage/SearchResultsList.tsx
```diff
@@ -94,6 +98,35 @@ export const SearchResultsList: FC<{
...
+                tokenPrices={tokenPrices}
+                highlightedPartOfName={searchTerm}
+                showLayer={true}
+              />
+            )
+          }
+          link={acc => RouteUtils.getAccountRoute(acc, acc.address)}
+          linkLabel={t('search.results.accounts.viewLink')}
+        />
+
```
lukaw3d, line 129: Why do we need to reimplement the accounts results group?

src/app/components/Account/AccountLink.tsx
```diff
@@ -54,6 +58,11 @@ interface Props {
    */
   alwaysTrimOnTable?: boolean
```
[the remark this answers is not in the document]
csillag: fixed typo

src/app/data/oasis-account-names.ts
```diff
@@ -0,0 +1,110 @@
...
+          })
+          map.set(address, metadata)
+        })
+        resolve({
+          map,
+          list,
+        })
+      }, reject)
+    }
+  })
```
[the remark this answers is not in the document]
csillag: Replace Promise with async / await

src/app/data/named-accounts.ts
```diff
@@ -0,0 +1,18 @@
...
+}
+
+export type AccountMetadataInfo = {
+  metadata?: AccountMetadata
+  loading: boolean
+}
+
+export type AccountMap = Map<string, AccountMetadata>
+
+export type AccountEntry = { address: string } & AccountMetadata
```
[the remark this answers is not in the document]
csillag: done

src/app/data/oasis-account-names.ts
```diff
@@ -0,0 +1,110 @@
...
+
+export const useOasisAccountsMetadata = (network: Network, layer: Layer, enabled: boolean) => {
```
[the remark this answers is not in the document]
csillag: done

src/app/data/oasis-account-names.ts
```diff
@@ -0,0 +1,110 @@
...
+}
+
+export const useOasisAccountMetadata = (
+  network: Network,
+  layer: Layer,
+  address: string,
+  enabled: boolean,
+): AccountMetadataInfo => {
+  // This is not a condition that can change while the app is running, so it's OK.
+  // eslint-disable-next-line react-hooks/rules-of-hooks
```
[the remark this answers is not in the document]
csillag: Yeah, code comment was obsolete, removed. (There was no condition here whatsoever, so the problem the comment has originally been described has since been solved a different way.)

src/app/pages/SearchResultsPage/SearchResultsList.tsx
csillag: We don't necessarily have to reimplement the group, but we need to extend it at least, because this kind of search result can't (immediately) be displayed like other accounts, since we only have the address at this point. (That is what we have the `DeferrentRuntimeAccountDetails` component and it's consensus counterpart.) So, possible approaches here:

- Extend the existing accounts group with a renderer function that can handle both type of results (full accounts and addresses to be loaded later)
- Modify the search function itself so it also loads the accounts and returns them. This means we don't need to touch the displaying part. Problems with this approach:
   - It would make the whole search slower (no results would be displayed for a longer time)
   - It's not easy to conditionally call multiple hooks from within a hook, thanks to the rules of hooks. The best way to do that is to actually render multiple components, and have each component call it's own hook to load that data, just as we are doing now with the `DeferredRuntimeAccountDetails`   component (and friends)
   
I would probably go with the first approach.

src/app/data/oasis-account-names.ts
csillag: Yes it does. And why is it a problem?

src/app/data/pontusx-account-names.ts
```diff
@@ -0,0 +1,79 @@
...
+import { hasTextMatch } from '../components/HighlightedText/text-matching'
+import { AccountData, AccountMetadata, AccountMap, AccountMetadataInfo } from './named-accounts'
+
+const DATA_SOURCE_URL = 'https://raw.githubusercontent.com/deltaDAO/mvg-portal/main/pontusxAddresses.json'
+
+const getPontusXAccountsMetadata = () =>
+  new Promise<AccountData>((resolve, reject) => {
+    axios.get(DATA_SOURCE_URL).then(response => {
+      if (response.status !== 200) reject("Couldn't load names")
+      if (!response.data) reject("Couldn't load names")
```
lukaw3d, line 14: can be modernized with async await

src/app/data/oasis-account-names.ts
```diff
@@ -0,0 +1,108 @@
...
+    const metadata: AccountMetadata = {
+      address,
+      name: entry.Name,
+      description: entry.Description,
+    }
+    // Register the metadata in its native form
+    list.push({
+      ...metadata,
+    })
+    map.set(address, metadata)
```
lukaw3d, line 48:
```suggestion
    const metadata: AccountMetadata = {
      address: entry.Address,
      name: entry.Name,
      description: entry.Description,
    }
    // Register the metadata in its native form
    list.push(metadata)
    map.set(metadata.address, metadata)
```

src/app/data/pontusx-account-names.ts
csillag: Yes, I have already done that with the Oasis counterpart, just had to step out a little bit..

src/app/pages/SearchResultsPage/SearchResultsList.tsx
lukaw3d: Second approach - getting full account objects makes a lot more sense to me:
- it's how an API endpoint would/will behave if we would have it. I expect API to search validators by name will return validator objects too
- I think it can be implemented in an easier to maintain way
- no need to call multiple hooks, just one conditional

csillag: > no need to call multiple hooks, just one conditional

OK, so let's assume that the client-side search (in the address-name mappings) have determined the address of the 3 accounts that have a match for the given name fragment. So we have the 3 addresses. How do we lead the 3 accounts with a single nexus API call?

lukaw3d: It would make multiple API requests, but in a single useQuery hook

csillag: Let's see if we can do something like [this](https://tanstack.com/query/latest/docs/framework/react/guides/dependent-queries?from=reactQueryV3#usequeries-dependent-query) in the nexus api  wrapper layer...

lukaw3d: oh I haven't seen that before, looks good too

src/app/data/pontusx-account-names.ts
csillag: replaced with aync await

src/app/data/oasis-account-names.ts
csillag: Simplified

src/app/data/pontusx-account-names.ts
```diff
@@ -0,0 +1,79 @@
...
+      const map: AccountMap = new Map()
+      const list: AccountMetadata[] = []
+      Object.entries(response.data).forEach(([address, name]) => {
+        map.set(address, { address, name: name as string })
+        const normalizedEntry: AccountMetadata = {
+          name: name as string,
+          address,
+        }
+        list.push(normalizedEntry)
```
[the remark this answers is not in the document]
csillag: Simplified

src/app/data/pontusx-account-names.ts
csillag: replaced with async await

src/app/data/oasis-account-names.ts
```diff
@@ -0,0 +1,110 @@
...
+    if (!url) {
+      reject('No data available for this layer')
+    } else {
+      axios.get(url).then(response => {
+        if (response.status !== 200) reject("Couldn't load names")
+        if (!response.data) reject("Couldn't load names")
+        // console.log('Response data is', response.data)
+        const map: AccountMap = new Map()
+        const list: AccountEntry[] = []
+        Array.from(response.data).forEach((entry: any) => {
```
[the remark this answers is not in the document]
csillag: Should we validate every element in the incoming JSON data every single time we load it?

src/app/data/oasis-account-names.ts
```diff
@@ -0,0 +1,110 @@
...
+  // This is not a condition that can change while the app is running, so it's OK.
+  // eslint-disable-next-line react-hooks/rules-of-hooks
+  const { isLoading, error, data: allData } = useOasisAccountsMetadata(network, layer, enabled)
+  if (error) {
+    console.log('Failed to load Oasis account metadata', error)
+  }
+  return {
+    metadata: allData?.map.get(address),
+    loading: isLoading,
+  }
```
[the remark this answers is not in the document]
csillag: I am not aware of it.
My thinking was that the data is valid with or without these names attached to the addresses, so we don't need to try so hard. Also, I don't think we have created (or want to create) UI design for the situation when we failed to load the names, and need to display an address, and we are not sure if there is a name attached to it...

So eventually we will ignore the missing data.

On the other hand, this is probably not the best layer to drop it. We could channel it up properly to the UI layer, and drop it there.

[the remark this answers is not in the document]
csillag: OK, I have now channeled the errors into the related UI components (AccountLink and various search result pages). We are still ignoring the errors eventually, but at least now the data is there, if we want to indicate it somehow.

src/app/data/oasis-account-names.ts
```diff
@@ -0,0 +1,110 @@
...
+
+const dataSources: Record<Network, Partial<Record<Layer, string>>> = {
+  [Network.mainnet]: {
+    [Layer.consensus]:
+      'https://raw.githubusercontent.com/oasisprotocol/nexus/main/account-names/mainnet_consensus.json',
+    [Layer.emerald]:
+      'https://raw.githubusercontent.com/oasisprotocol/nexus/main/account-names/mainnet_paratime.json',
+    [Layer.sapphire]:
+      'https://raw.githubusercontent.com/oasisprotocol/nexus/main/account-names/mainnet_paratime.json',
+  },
```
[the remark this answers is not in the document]
csillag: Yeah. The problem is that they have created their format first, so it's not so nice to ask them to re-implement it now.

src/app/data/oasis-account-names.ts
csillag: Removed unnecessary fallback values.

src/app/data/pontusx-account-names.ts
```diff
@@ -0,0 +1,76 @@
...
+      if (!response.data) reject("Couldn't load names")
+      const map: AccountMap = new Map()
+      const list: AccountEntry[] = []
+      Object.entries(response.data).forEach(([address, name]) => {
+        map.set(address, { name: name as string })
+        const normalizedEntry: AccountEntry = {
+          name: name as string,
+          address,
+        }
+        list.push(normalizedEntry)
```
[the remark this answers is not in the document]
csillag: I have done that now. (But I still don't want to unite the (trivial) lookup code, because it's entirely possible that some other layers will want an API-based lookup for addresses, so that must be kept separate.)

src/app/pages/SearchResultsPage/SearchResultsList.tsx
csillag: > It would make multiple API requests, but in a single useQuery hook

How would you do that?

The problem seems to be that we already have multiple layers of wrappers around the actual Nexus API call, and we have added come logic in each layers. Even the auto-generated layer adds useful stuff, and then there is the data transformation logic in the wrapper layer, plus we want to mix in the results of the RPC calls. Extracting all that logic, and re-applying it after we call the Nexus API another way (ie. avoiding our current wrappers) would be a pain. So I would stick to using the already existing hooks... but then already have `useQuery` hooks, and it's not easy to dynamically call them multiple times from within a single hook.

> it's how an API endpoint would/will behave if we would have it. I expect API to search validators by name will return validator objects too

Yeah, if we had that API, it would be a different matter ... but we don't.

For now, it seems to be a lot easier to just render more components, and have them load their own data using the existing APIs.

    * * *
 
As a first step I'll just unite the two "Accounts" section in the results list. (Although I don't think we can have both of them at the same time, given that you would have to search for completely different things to get results in the two sections.)

csillag: > As a first step I'll just unite the two "Accounts" section in the results list. 

I have done that now.

csillag: How about we limit the number of results from this specific sub-search? (Accounts listed because of matching name segment.) If we can agree to show at most 20 such result (or any fixed number), then we can launch the subsequent account loader sub-queries using a simple `for` cycle of fixed length, which doesn't break the rules of hooks. (And then  use `enabled` to selectively enable the ones we need, based on the number of matching accounts found.)

That way we can emulate that we have the missing API, and then all the higher layers (UI etc) can be kept simple. (We can drop the DefferedRuntimeAccount component and friends, no need for any expansion in the result list account section,etc.)

Maybe we can indicate at the end of the search results list if there are more matches that haven't been displayed. @donouwens would that be acceptable?

csillag: >  where is the cut off

Wherever we want it to be. It can be anything, but it has to be a fixed number. Bigger numbers come with a bigger performance penalty, when we try to load all the account data for all matching accounts _all at once_, it as many parallel requests as we have accounts to display.

> would there be a way for users to view to additional results?

Not an easy one.

   * * *

Please note that we also have other (currently unclear an undefined) limits for all other results, too. I don't know what those are, but probably no more than 100 or 1000 for each kind of objects.

csillag: > Second approach - getting full account objects makes a lot more sense to me

I have changed implementation so that we are now getting the full account objects with the hooks now. 
Accordingly, eliminated the new result type.

Please note that this should only be reviewed after #1422.

csillag: As we have discussed at today's meeting, this is all we can do here, for now.

src/app/data/oasis-account-names.ts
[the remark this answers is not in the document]
lukaw3d: that's an option too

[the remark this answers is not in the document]
lukaw3d: I was thinking a cronjob:
- fetch json
- write a typescript file: `const data: Type = <json>`
- yarn checkTs

src/app/data/oasis-account-names.ts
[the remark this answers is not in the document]
lukaw3d: nice

src/app/pages/SearchResultsPage/SearchResultsView.tsx
```diff
@@ -31,12 +32,14 @@ export const SearchResultsView: FC<{
...
           searchTerm={searchTerm}
           searchResults={searchResults.filter(getFilterForLayer(wantedScope.layer))}
+          isPotentiallyIncomplete={isPotentiallyIncomplete}
           tokenPrices={tokenPrices}
         />
       ) : (
         <GlobalSearchResultsView
           searchTerm={searchTerm}
           searchResults={searchResults}
+          isPotentiallyIncomplete={isPotentiallyIncomplete}
```
lukaw3d, line 42: Does this get set if there are more results than MASS_LOAD_INDEXES

src/oasis-nexus/api.ts
```diff
@@ -110,17 +111,20 @@ declare module './generated/api' {
...
-      available === '0' &&
-      debonding_delegations_balance === '0' &&
-      delegations_balance === '0' &&
-      escrow === '0' &&
-      total === '0' &&
-      nonce === 0 &&
-      size === 'XXS'
-    )
+    // TODO: find a sane way to recognize an important consensus account.
+    // The heuristics below is clearly insufficient, because it would indicate even named accounts like "Governance Escrow" to be empty.
```
lukaw3d, line 115: +1

src/oasis-nexus/api.ts
```diff
@@ -304,6 +308,10 @@ export const useGetConsensusAccountsAddress: typeof generated.useGetConsensusAcc
   const ticker = getTokensForScope({ network, layer: Layer.consensus })[0].ticker
   return generated.useGetConsensusAccountsAddress(network, address, {
     ...options,
+    query: {
+      ...options?.query,
+      queryKey: [network, Layer.consensus, 'accounts', 'byAddress', address],
```
lukaw3d, line 313: How come this is needed?

src/app/pages/SearchResultsPage/useRedirectIfSingleResult.ts
```diff
@@ -34,7 +34,7 @@ export function useRedirectIfSingleResult(
         redirectTo = RouteUtils.getTransactionRoute(item, item.eth_hash || item.hash)
         break
       case 'account':
-        redirectTo = RouteUtils.getAccountRoute(item, (item as RuntimeAccount).address_eth ?? item.address)
+        redirectTo = `${RouteUtils.getAccountRoute(item, (item as RuntimeAccount).address_eth ?? item.address)}?q=${searchTerm}`
```
lukaw3d, line 37: For future improvement:
https://92fdad12.oasis-explorer.pages.dev/mainnet/consensus/address/oasis1qp000ff4weh6t6083hazdpw0r902leffp5gjgu55?q=oasis1qp000ff4weh6t6083hazdpw0r902leffp5gjgu55

src/app/pages/ConsensusAccountDetailsPage/ConsensusAccountDetailsCard.tsx
```diff
@@ -13,18 +13,25 @@ type ConsensusAccountDetailsCardProps = {
...
+  highlightedPartOfName,
 }) => {
   const { t } = useTranslation()
 
   return (
     <SubPageCard featured isLoadingTitle={isLoading} title={t('account.title')} mainTitle>
-      <ConsensusAccountDetailsView isError={isError} isLoading={isLoading} account={account} />
+      <ConsensusAccountDetailsView
+        isError={isError}
+        isLoading={isLoading}
```
lukaw3d, line 31: Something isn't handling loading state correctly:
reload [an account](https://92fdad12.oasis-explorer.pages.dev/mainnet/consensus/address/oasis1qp000ff4weh6t6083hazdpw0r902leffp5gjgu55) a few times to see an warning before it gets loaded
[screen-recording.webm](https://github.com/oasisprotocol/explorer/assets/3758846/839e3302-9df5-4828-8fa8-b8f7cf084a73)

src/oasis-nexus/api.ts
```diff
@@ -433,6 +443,64 @@ export const useGetRuntimeAccountsAddress: typeof generated.useGetRuntimeAccount
...
+      target?.address ?? '',
+      {
+        query: { enabled: queryOptions.enabled && !!target?.address },
+      },
+    ),
+  )
+  return {
+    isLoading: !!targets?.length && queries.some(query => query.isLoading && query.fetchStatus !== 'idle'),
+    isError: queries.some(query => query.isError),
+    isFetched: !queries.some(query => !query.isFetched),
```
lukaw3d, line 474: This doesn't ignore disabled MASS_LOAD queries. Is it always false?

src/app/data/oasis-account-names.ts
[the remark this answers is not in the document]
csillag: Yes, that sounds nice. Maybe we could make this into a GitHub action, which can create a GitHub issue (in the nexus repo) if it finds a problem.

src/app/pages/SearchResultsPage/SearchResultsView.tsx
csillag: Now it does.

src/oasis-nexus/api.ts
csillag: There is already an ongoing with the Nexus team to acquire the missing data. Will follow up on this when appropriate.

src/oasis-nexus/api.ts
csillag: It isn't. Removed.

src/app/pages/SearchResultsPage/useRedirectIfSingleResult.ts
csillag: Up to now, the idea was that we only do highlighting when (partial) text matching is involved. It didn't occur to me that we could just as well highlight fields when equality was for testing, such as block number, tx hash or address. Indeed, we could do that. I will discuss this with @donouwens.

src/oasis-nexus/api.ts
csillag: We don't actually use or need the values calculated here. Removed.

src/app/pages/ConsensusAccountDetailsPage/ConsensusAccountDetailsCard.tsx
csillag: I can reproduce the problem, but unfortunately this is also happening on master branch (at least since merging #1422), so the issue is independent of the current PR. (I will fix this in a follow-up PR.)

csillag: See fix in #1423.

src/app/pages/SearchResultsPage/useRedirectIfSingleResult.ts
lukaw3d: I didn't mean that :D I meant: I prefer the old URLs when searching for an exact address

lukaw3d (approved):

src/app/components/Account/AccountLink.tsx
```diff
@@ -24,7 +28,7 @@ const WithTypographyAndLink: FC<{
         ...(mobile
           ? {
               maxWidth: '100%',
-              overflowX: 'hidden',
+              overflow: 'hidden',
```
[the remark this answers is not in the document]
lukaw3d: (I don't think this behaves different from overflow-x)

src/app/pages/SearchResultsPage/useRedirectIfSingleResult.ts
csillag: > I prefer the old URLs when searching for an exact address

Oh, that's not difficult to change. Fixed that now.

PR #1401, opened 2024-05-08
Improve Consensus on the metaverse graph

csillag:
Task: https://app.clickup.com/t/8693kb88a

- Make it so that that clicking on the consensus circle
  selects the layer, instead if opening it immediately,
  like we have for the other layers
- Add online/out-of-date status displays everywhere

**Update**:

Feedback from @donouwens:

> I think it works great. 
>
> I did notice two other things though  [...] 
> The opacity of the popups should be increased (from (0, 0, 98, 0.48); to (0, 0, 98, 0.75);)

### DIFF
--- a/src/app/pages/HomePage/Graph/Graph/index.tsx
+++ b/src/app/pages/HomePage/Graph/Graph/index.tsx
@@ -159,7 +159,7 @@ const handleHover: (
   },
 })
 
-const GraphParaTimeStatus: FC<
+const GraphLayerStatus: FC<
   PropsWithChildren<{
     fillText: string
     iconX: number
@@ -565,7 +565,7 @@ const GraphCmp: ForwardRefRenderFunction<SVGSVGElement, GraphProps> = (
           {...handleHover(Layer.emerald, setHoveredLayer)}
         >
           {(isMobile || hoveredLayer !== Layer.emerald) && (
-            <GraphParaTimeStatus
+            <GraphLayerStatus
               iconX={201}
               iconY={102}
               textX={174}
@@ -574,7 +574,7 @@ const GraphCmp: ForwardRefRenderFunction<SVGSVGElement, GraphProps> = (
               outOfDate={outOfDateMap.emerald}
             >
               {t('common.emerald')}
-            </GraphParaTimeStatus>
+            </GraphLayerStatus>
           )}
 
           {!isMobile && hoveredLayer === Layer.emerald && !disabledMap[Layer.emerald] && (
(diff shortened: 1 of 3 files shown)


src/app/pages/HomePage/Graph/Graph/index.tsx
```diff
@@ -699,9 +699,16 @@ const GraphCmp: ForwardRefRenderFunction<SVGSVGElement, GraphProps> = (
           {...handleHover(Layer.consensus, setHoveredLayer)}
         >
           {(isMobile || hoveredLayer !== Layer.consensus) && (
-            <text x="161" y="212" fill={graphTheme.text} fontSize="12px">
+            <GraphLayerStatus
```
buberdds, line 702: 1) it's misaligned for me can we adjust `textY` ?
2) Currently it works the same way as removed line. It never shows online/offline badge after clicking Consensus circle.

.changelog/1401.feature.md
```diff
@@ -0,0 +1 @@
+Handle the consensus layer in the metaverse graph the same way as other layers
```
buberdds, line 1: It's rather trivial task not a feature

src/app/pages/HomePage/Graph/Graph/index.tsx
csillag: > It never shows online/offline badge after clicking Consensus circle.

image

csillag: > it's misaligned for me can we adjust textY ?

OK, after adjustment:

image

Is it OK now?

csillag: > It never shows online/offline badge after clicking Consensus circle.

The trick is that it only show the badge when we are zoomed in, but clicking the Consensus circle doesn't really zoom in, since it still selects the same network. The only case we actually see this badge if we zoom in by clicking on another circle, and the panning to the consensus circle.

I was not sure if this was the desired behavior, so I asked @donouwens to check, he did, and said that it works OK the way it is.

.changelog/1401.feature.md
csillag: OK, fixed

src/app/pages/HomePage/Graph/Graph/index.tsx
buberdds: ahh if it's by design ignore my comment

buberdds (approved):

src/app/pages/HomePage/Graph/Graph/index.tsx
buberdds: Don says it has to zoom in :shrug:

csillag: See follow-up PR https://github.com/oasisprotocol/explorer/pull/1404

PR #1404, opened 2024-05-09
Actually zoom in on mobile when consensus circle is selected

csillag:
When selecting the "consensus" circle in the graph browser (on mobile), it should zoom it.

For context, see [here](https://github.com/oasisprotocol/explorer/pull/1401/files#r1595173633).

### DIFF
--- a/src/app/pages/HomePage/Graph/Graph/graph-utils.ts
+++ b/src/app/pages/HomePage/Graph/Graph/graph-utils.ts
@@ -38,7 +38,11 @@ export abstract class GraphUtils {
           y: 1.1 * height,
         }
       case Layer.consensus:
-        return initialValue
+        return {
+          scale: 2.4,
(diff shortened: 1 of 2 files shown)


buberdds (approved):

src/app/pages/HomePage/Graph/Graph/graph-utils.ts
```diff
@@ -38,7 +38,11 @@ export abstract class GraphUtils {
           y: 1.1 * height,
         }
       case Layer.consensus:
-        return initialValue
+        return {
```
buberdds, line 41: Looks like this broke `zoom out` button on desktop.

csillag: See fix in https://github.com/oasisprotocol/explorer/pull/1413

PR #1406, opened 2024-05-10
Update Explorer logotype

buberdds:
https://app.clickup.com/t/8694cukt5


src/app/components/CustomIcons/ExplorerIcon.tsx
```diff
@@ -1,23 +1,31 @@
 import { FC } from 'react'
 import SvgIcon, { SvgIconProps } from '@mui/material/SvgIcon'
 
+export const viewBoxAspectRatio = 782.02 / 146.19
```
lubej, line 4: This is a little tricky to see in code. But I ok I guess.

lubej (approved):

src/app/components/CustomIcons/ExplorerIcon.tsx
buberdds: yeah maybe it;s not worth doing. I will bring back hardcoded value

PR #1411, opened 2024-05-14
Fix mobile search issues

buberdds:
https://app.clickup.com/t/8694h28bx

current master:

Screenshot from 2024-05-14 09-13-15
Screenshot from 2024-05-14 09-12-45

### DIFF
--- a/src/app/components/Search/index.tsx
+++ b/src/app/components/Search/index.tsx
@@ -52,6 +52,9 @@ const SearchForm = styled('form', {
               display: 'none',
             },
             [`.${inputAdornmentClasses.positionEnd}`]: {
+              'button[type="button"]': {
+                display: 'none',
+              },
               marginLeft: 0,
             },
           },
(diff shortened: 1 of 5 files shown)


lubej:
Should we just remove the search in footer on HomePage for mobile? I don't see the point, because you have the search on the top of the page.

buberdds:
I have asked the same question to Don, but an answer was to fix z-index. I will validate this one more time with him.

buberdds:
cc @donouwens

lubej:
> The reason we have got the search in the footer is that inner pages, like block detail or account pages, often get very long. This way the user has an option to access search from the bottom of the page as well. A sticky navigation was considered but deemed to take up too much screen estate.

Yes, makes sense on other pages. But not on home page, where the page's height isn't large, and you can almost seen both search bars at the same time.

src/app/components/Search/index.tsx
```diff
@@ -52,6 +52,9 @@ const SearchForm = styled('form', {
               display: 'none',
             },
             [`.${inputAdornmentClasses.positionEnd}`]: {
+              'button[type="button"]': {
+                display: 'none',
+              },
```
lubej, line 57: Why is this change necessary?

lubej (approved):

src/app/components/Search/index.tsx
buberdds: (the first screenshot in the description) 
When mobile footer search input is collapsed, but has a value, a part of it is not fully hidden. Search icon positioning is affected too and only half of icon is visible.

PR #1413, opened 2024-05-14
Fix the "zoom out" button

csillag:
Earlier, the "zoom out" functionality (of the metaverse graph, in desktop mode)  worked by selecting the consensus layer, which had the whole graph as designated area, so zooming to it actually zoomed out.

But then we wanted to make the consensus circle selectable and zoomable, so in #1404 the designated zoom are has been changed. This made is possible to zoom in on selection (on mobile), but has broken the zoom out button (on desktop). (see https://github.com/oasisprotocol/explorer/pull/1404#discussion_r1600031198 )

So now we are fixing this be introducing a special "fake" layer, just for the components in/around the layer selector, for zooming out: the "universe layer". This layer will have the whole graph as designated zoom area, so selecting this layer will zoom out, while the consensus layer can have it's small area.

Also, since it's no longer always a layer, we are now calling the selection on the graph "area". All related props and functions have been renamed accordingly.

### DIFF
--- a/src/app/pages/HomePage/Graph/ParaTimeSelector/index.tsx
+++ b/src/app/pages/HomePage/Graph/ParaTimeSelector/index.tsx
@@ -152,11 +152,16 @@ interface ParaTimeSelectorProps extends ParaTimeSelectorBaseProps {
 
 const localStore = storage()
 
-// This is a special layer used to indicate that we should zoom out,
+// This is a special area used to indicate that we should zoom out,
 // and see the whole universe.
-export const UniverseLayer = 'Universe'
+export const UniverseArea = 'Universe'
 
-export type SelectorLayer = Layer | typeof UniverseLayer
+export type SelectorArea =
+  | typeof UniverseArea
+  | typeof Layer.consensus
+  | typeof Layer.cipher
+  | typeof Layer.emerald
+  | typeof Layer.sapphire
 
 const ParaTimeSelectorCmp: FC<ParaTimeSelectorProps> = ({
   disabled,
@@ -174,13 +179,11 @@ const ParaTimeSelectorCmp: FC<ParaTimeSelectorProps> = ({
   const exploreBtnTextTranslated = t('home.exploreBtnText')
   const { network, setNetwork } = useSearchQueryNetworkParam()
 
-  // Using object here to force side effect trigger when setting to the same layer
-  const [selectedLayer, setSelectedLayer] = useState<{ current: SelectorLayer }>()
-  const [activeMobileGraphTooltip, setActiveMobileGraphTooltip] = useState<{ current: SelectorLayer | null }>(
-    {
-      current: null,
-    },
-  )
+  // Using object here to force side effect trigger when setting to the same area
+  const [selectedArea, setSelectedArea] = useState<{ current: SelectorArea }>()
+  const [activeMobileGraphTooltip, setActiveMobileGraphTooltip] = useState<{ current: SelectorArea | null }>({
+    current: null,
+  })
 
   const [scale, setScale] = useState<number>(1)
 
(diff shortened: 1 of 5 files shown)


buberdds:
something is wrong after zoom out

Screenshot from 2024-05-14 16-55-35

src/app/pages/HomePage/Graph/ParaTimeSelector/index.tsx
```diff
@@ -160,6 +160,12 @@ interface ParaTimeSelectorProps extends ParaTimeSelectorBaseProps {
 
 const localStore = storage()
 
+// This is a special layer used to indicate that we should zoom out,
+// and see the whole universe.
+export const UniverseLayer = 'Universe'
+
+export type SelectorLayer = Layer | typeof UniverseLayer
```
lubej, line 167: Would be nice to exclude `pontusx` here, as it is not shown on the graph. And it probably won't be.

csillag: Why do we need to exclude it? What happens with layers that are not excluded explicitly, but we don't have specific zoom areas defined for them? Nothing, as far as I tell....

lubej: Nothing, as they can not be clicked.

src/app/pages/HomePage/Graph/ParaTimeSelector/index.tsx
```diff
@@ -160,6 +160,12 @@ interface ParaTimeSelectorProps extends ParaTimeSelectorBaseProps {
 
 const localStore = storage()
 
+// This is a special layer used to indicate that we should zoom out,
+// and see the whole universe.
+export const UniverseLayer = 'Universe'
```
lukaw3d, line 165: Don't name it "layer"

lukaw3d: It impacts downstream things like `setSelectedLayer` too

csillag: OK, what naming would y'all suggest, instead of layer? `section` ? `region` ? `part` ? `segment` ? Something else entirely?

src/app/pages/HomePage/Graph/ParaTimeSelector/index.tsx
csillag: Yeah anyway, I think I'll just explicitly list list the ones that we want, instead of excluding stuff.

src/app/pages/HomePage/Graph/ParaTimeSelector/index.tsx
lukaw3d: I would name it area :shrug:

A similar concept in SVG: https://developer.mozilla.org/en-US/docs/Web/SVG/Element/view
and HTML https://developer.mozilla.org/en-US/docs/Web/HTML/Element/area

src/app/pages/HomePage/Graph/ParaTimeSelector/index.tsx
lubej: Works for me.

csillag: OK, now using:

```
export type SelectorArea =
  | typeof UniverseArea
  | typeof Layer.consensus
  | typeof Layer.cipher
  | typeof Layer.emerald
  | typeof Layer.sapphire
```

src/app/pages/HomePage/Graph/ParaTimeSelector/index.tsx
csillag: Ok, calling it area now. (And recursively renamed all relevant arguments and functions)

lukaw3d (approved):

lukaw3d (changes requested):
oh, it is a strangely broken after zoom out
https://github.com/oasisprotocol/explorer/pull/1413#issuecomment-2110463001

csillag:
> something is wrong after zoom out

Yeah, thanks, good catch. Fixed.

buberdds (approved):

Issue #1410, opened 2024-05-14
Refactor usage of React deprecated API


buberdds:
Get rid of `.defaultProps` usage

PR #1414, opened 2024-05-14
Remove usage of deprecated React API

buberdds:
Fixes https://github.com/oasisprotocol/explorer/issues/1410

- needed for incoming version of React
- removed a few buttons from Storybook as they are regular MUI buttons

### DIFF
--- a/src/app/components/Search/index.tsx
+++ b/src/app/components/Search/index.tsx
@@ -3,7 +3,7 @@ import { useNavigate, useSearchParams } from 'react-router-dom'
 import TextField, { textFieldClasses } from '@mui/material/TextField'
 import InputAdornment, { inputAdornmentClasses } from '@mui/material/InputAdornment'
 import { styled } from '@mui/material/styles'
-import Button from '@mui/material/Button'
+import Button, { ButtonProps } from '@mui/material/Button'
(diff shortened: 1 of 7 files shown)


src/app/components/Search/index.tsx
```diff
@@ -90,10 +90,10 @@ export const SearchButton = styled(Button, {
...
     : {}),
 }))
-SearchButton.defaultProps = {
-  variant: 'contained',
-  color: 'primary',
-}
+
+export const SearchButton = (props: ButtonProps & SearchButtonProps) => (
+  <StyledSearchButton variant="contained" color="primary" {...props} />
+)
```
lubej, line 96: I would rather see something like this:

```suggestion
const PlainTextButton = ({variant = 'text', color: 'inherit', ...restProps}: ButtonProps) => (
  <StyledPlainTextButton variant={variant} color={color} {...restProps} />
)
```

But not a big deal.

lukaw3d: I vote for original

lukaw3d: some hope for a nicer solution in the future https://github.com/emotion-js/emotion/issues/2573#issuecomment-1925000208

PR #1422, opened 2024-05-20
Support searching for consensus accounts

csillag:
Up to now, we could only search for runtime accounts by address.

This PR fixes that.

Note to the reviewer: the first too commits only contain refactoring, the new functionality is in the 3rd one.

### DIFF
--- a/src/app/components/Account/index.tsx
+++ b/src/app/components/Account/index.tsx
@@ -5,7 +5,7 @@ import { useScreenSize } from '../../hooks/useScreensize'
 import { StyledDescriptionList, StyledListTitleWithAvatar } from '../../components/StyledDescriptionList'
 import { CopyToClipboard } from '../../components/CopyToClipboard'
 import { TextSkeleton } from '../../components/Skeleton'
-import { EvmToken, type RuntimeAccount } from '../../../oasis-nexus/api'
+import { Account, EvmToken, type RuntimeAccount } from '../../../oasis-nexus/api'
 import { TokenPills } from './TokenPills'
 import { AccountLink } from './AccountLink'
 import { RouteUtils } from '../../utils/route-utils'
@@ -23,6 +23,9 @@ import { RuntimeBalanceDisplay } from '../Balance/RuntimeBalanceDisplay'
 import { calculateFiatValue } from '../Balance/hooks'
 import { FiatMoneyAmount } from '../Balance/FiatMoneyAmount'
 import { getFiatCurrencyForScope, getTokensForScope, showFiatValues } from '../../../config'
+import Box from '@mui/material/Box'
+import { AccountSizeBadge } from '../AccountSizeBadge'
+import { StyledListTitle } from '../../pages/ConsensusAccountDetailsPage/ConsensusAccountDetailsCard'
 
 type RuntimeAccountDataProps = {
   account?: RuntimeAccount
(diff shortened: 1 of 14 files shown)


src/app/pages/SearchResultsPage/useRedirectIfSingleResult.ts
```diff
@@ -33,7 +34,7 @@ export function useRedirectIfSingleResult(
         redirectTo = RouteUtils.getTransactionRoute(item, item.eth_hash || item.hash)
         break
       case 'account':
-        redirectTo = RouteUtils.getAccountRoute(item, item.address_eth ?? item.address)
+        redirectTo = RouteUtils.getAccountRoute(item, (item as RuntimeAccount).address_eth ?? item.address)
```
lubej, line 37: Why is this cast necessary?

src/app/components/Account/index.tsx
```diff
@@ -160,3 +169,85 @@ export const Account: FC<AccountProps> = ({ account, token, isLoading, tokenPric
...
+  account,
+  isLoading,
+  showLayer,
+  standalone,
+}) => {
+  const { t } = useTranslation()
+  const { isMobile } = useScreenSize()
+
+  if (isLoading) return <TextSkeleton numberOfRows={7} />
+  if (!account) return null
```
lubej, line 190: We should probably better handle this? There is no account with this address, or something like that?

src/app/pages/ConsensusAccountDetailsPage/ConsensusAccountDetailsView.tsx
```diff
@@ -0,0 +1,35 @@
...
+import { CardEmptyState } from '../../components/CardEmptyState'
+import { ConsensusAccountData } from '../../components/Account'
+
+type ConsensusAccountDetailsViewProps = {
+  account: Account | undefined
+  isError?: boolean | undefined
+  isLoading?: boolean | undefined
+  showLayer?: boolean
+  standalone?: boolean
+}
```
lubej, line 13: Can we extend this interface from `ConsensusAccountDataProps`, looks to me pretty similar.

src/app/pages/ConsensusAccountDetailsPage/ConsensusAccountDetailsView.tsx
```diff
@@ -0,0 +1,35 @@
...
+  account,
+  isError,
+  isLoading,
+  showLayer,
+  standalone,
+}) => {
+  const { t } = useTranslation()
+
+  if (isError) return <CardEmptyState label={t('account.cantLoadDetails')} />
+  if (!account) return null
```
lubej, line 25: We should probably better handle this? There is no account with this address, or something like that?

src/oasis-nexus/api.ts
```diff
@@ -107,15 +108,30 @@ declare module './generated/api' {
...
+    const { available, size, nonce, debonding_delegations_balance, delegations_balance, escrow, total } =
+      account as Account
+    return (
+      available === '0' &&
+      debonding_delegations_balance === '0' &&
+      delegations_balance === '0' &&
+      escrow === '0' &&
+      total === '0' &&
+      nonce === 0 &&
+      size === 'XXS'
```
lubej, line 122: This looks confusing to me, everything is necessary to check if the account is "empty" on consensus layer?

src/app/pages/SearchResultsPage/useRedirectIfSingleResult.ts
csillag: Because the data can be either (consensus) `Account` or `RuntimeAccount`, and only `RuntimeAccount` has this field. This is not a problem, since we can use `address` as a fallback, but TS is complaining when accessing a field that might not be defined at all (in case of `Account`).

src/app/components/Account/index.tsx
csillag: 1. I didn't write this code now, only moved this around.
2. I am not even sure if this ever happens.
   - If the address is invalid, we get a different error.
   - If we can't load because of network issues, we get a different error
   - If the data is still showing, we get a different error.

So I think this is probably a dead codepath. However, if we remove it, then TS will complain about `account` being possibly undefined later on...

src/app/pages/ConsensusAccountDetailsPage/ConsensusAccountDetailsView.tsx
csillag: Done

src/oasis-nexus/api.ts
csillag: It turns out that until we get a field about the number of transactions, we can't reliably detect and "empty" consensus account. (I have seen [cases](https://github.com/oasisprotocol/nexus/blob/main/account-names/mainnet_consensus.json#L13) where everything looks empty, but it's actually an important system account with lots of TXs.)

I think hiding a real account in search is worse than showing an extra empty one, I decided that until this is fixed, no consensus accounts should be considered to be empty.

src/app/pages/ConsensusAccountDetailsPage/ConsensusAccountDetailsView.tsx
csillag: See https://github.com/oasisprotocol/explorer/pull/1422#discussion_r1607801383

lubej: @buberdds any reason why we are returning `null` here? I would expect a better explanation for this edge case. Or does it really never happens?

src/oasis-nexus/api.ts
lubej: Ok, but for example, how is then "important system account" marked as `size === 'XXS'`?

src/app/pages/ConsensusAccountDetailsPage/ConsensusAccountDetailsView.tsx
csillag: I think when the query returns undefined for the data, then it also sets either `isLoading` or `isError` to true.
So I don't think actually happens.

src/oasis-nexus/api.ts
csillag: > Ok, but for example, how is then "important system account" marked as `size === 'XXS'`?

The size is calculated based on the current total balance. The account in question has a total balance of 0, because it doesn't hold any ROSE currently, but it's still an important system account, which many transactions.
(Which we fail to load because of a DB indexing issue, but that is being taken care of by the Nexus team .)

src/app/pages/ConsensusAccountDetailsPage/ConsensusAccountDetailsView.tsx
csillag: In any case, I have updated this to still show the "can't load details" empty filler screen.

src/oasis-nexus/api.ts
csillag: I have raised this issue on with the Nexus guys. I don't think where is much more we can do ATM.

lubej: Ok, please follow up, if there is a simpler way of doing this.

lubej (approved):

src/app/components/Account/index.tsx
```diff
@@ -160,3 +169,84 @@ export const Account: FC<AccountProps> = ({ account, token, isLoading, tokenPric
...
 }
+
+export type ConsensusAccountDataProps = {
+  account?: Account
+  isLoading?: boolean
+  showLayer?: boolean
+  standalone?: boolean
+}
+
+export const ConsensusAccountData: FC<ConsensusAccountDataProps> = ({
```
lukaw3d, line 180: What are components with "*Data" suffix?

csillag: Unfortunately, we don't have any consistent convention for naming all our components (that I am aware of).

In case of these two components, they simply display the data in the fields of the object. `RuntimeAccountData` displays a `RuntimeAccount`. Earlier this React component used to be called `Account`, but that's not really a good name since we also have an Account type in the data layer (for Consensus accounts.) `ConsensusAccountData` used to be implemented as part of the page code for displaying consensus account details; it has been factored out in #1422.

lukaw3d: ah idk :shrug:
- we use "*View" for mostly stateless subsets of components and "*Page" for non-reusable
- "*Data" for component reminds me of renderless components (before hooks existed)
- out of RuntimeAccount and RuntimeAccountData I would expect RuntimeAccountData to be the data layer type

csillag: Yeah, it is a true that "*Data" invokes the expectation of finding logic and data storage. The reason I didn't go with "*View" in this case is that we already have some components called `RuntimeAccountView`, so I wanted something else. On the other hand, we can probably do with one less layer of wrappers (between the page and the actual data display), and in that case, the only remaining class could still be called "*View". Let me try that in a separate follow-up PR.

csillag: See fix in https://github.com/oasisprotocol/explorer/pull/1423.

PR #1423, opened 2024-05-21
Consensus account fixup

csillag:
This PR fixes some problems introduced by #1422 

- Inconsistent displaying of loading/error state while loading consensus account details
- Inconsistent component class naming (for displaying consensus and runtime account details)


.changelog/1423.internal.md
```diff
@@ -0,0 +1 @@
+Follow-up / clean-up after #1422
```
lukaw3d, line 1: If fragment text and type are the same as #1422 it will get merged in the changelog

src/app/components/Account/ConsensusAccountDetailsView.tsx
```diff
@@ -0,0 +1,103 @@
...
+  isError,
+  isLoading,
+  showLayer,
+  standalone,
+}) => {
+  const { t } = useTranslation()
+  const { isMobile } = useScreenSize()
+
+  if (!account || isLoading) return <TextSkeleton numberOfRows={7} />
+  if (isError || (!isLoading && !account)) return <CardEmptyState label={t('account.cantLoadDetails')} />
```
lukaw3d, line 40: Why is this not `if (isLoading)` and `if (isError)`?

.changelog/1423.internal.md
csillag: Oh, sweet. Did that now.

src/app/components/Account/ConsensusAccountDetailsView.tsx
csillag: > Why is this not `if (isLoading)` and `if (isError)`?

Updated to `if (isLoading)` and `if (isError || !account)`

Testing for account in the latter is required to handle the (hypothetical) situation when `account` is undefined.

Before #1422, this was handled by adding ` if (!account) return null`, after the loading/error logic, but @lubej didn't like returning null instead of a proper explanation, so it has been changed.

lukaw3d: This sounds like inaccurate types. Account should be defined after loading without an error

lukaw3d: useQuery types for comparison https://github.com/TanStack/query/blob/59eef10c1afa5b16ffcac4cd805d86975640f927/packages/query-core/src/types.ts#L675-L718

lukaw3d (approved):

src/app/components/Account/ConsensusAccountDetailsView.tsx
csillag: OK, the type definitions basically confirm that the data will be defined except when loading or on error. 
So, adding it as a condition here doesn't cause any change in behavior, but it satisfies the TS compiler.

So I think we are OK.

PR #1424, opened 2024-05-22
Assemble changes for 1.11.0 release

buberdds:


lubej (approved):

buberdds:
Layer is disabled on prod, why we need it for a release?

csillag:
> Layer is disabled on prod, why we need it for a release?

Good point. Never mind,then.

PR #1425, opened 2024-05-22
Get all deployed Nexus API changes

csillag:
I have updated the code step by stap (nexus API revision by revision), so that it's easy to see what is changing.

Please note that Nexus v0.3.0 has been deployed to production, but not to staging. Currently Explorer is using the staging deployment.

This means some of the new features will only be available if 
- the new Nexus version is deployed everywhere, or
- the backed servers are switched from staging to production. (Please also note that the domain names for this are reversed 😕  )

There were some breaking changes, but our code has been updated to handle it.

PR #1426, opened 2024-05-22
Get latest Nexus changes (not yet deployed)

csillag:
Please see #1425 for all the Nexus API changes up to v0.3.0 (currently deployed).

This PR depends on that, and added the remaining API changes, which are not yet deployed.
There is a breaking change, and out code is adapted accordingly.


csillag:
No, scratch that. Actually, this latest change is also part of Nexus v0.3.0, so there is no need for this separate PR.

PR #1427, opened 2024-05-23
Fix account redirection logic for checksum capitalized addresses

lukaw3d:
Fixes https://github.com/oasisprotocol/explorer/commit/a59aae37d64c1b208c62f85fc3f5066a7d2e0be9

### DIFF
--- a/.changelog/1427.trivial.md
+++ b/.changelog/1427.trivial.md
@@ -1,0 +1,1 @@
+Fix account redirection logic for checksum capitalized addresses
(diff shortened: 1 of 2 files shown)


lubej (approved):

csillag (approved):
Oh. I was thinking about lowercasing the addresses before comparison, but then I assumed that Nexus is already lowercasing it during the indexing. I forgot about checksum accounts.

LGTM.

buberdds:
@lukaw3d do we want to include it in a release? if that's a case let's switch to bugfix

csillag:
> @lukaw3d do we want to include it in a release? if that's a case let's switch to bugfix

I don't think it's critical.

The worst think that can happen without this fix is that when someone searches for an account by a checksum capitalizd address, and there is only one result, and we redirect to it, then the URL will have an unnecessary query param, which won't cause any technical problem.

lukaw3d:
yeah not critical

PR #1435, opened 2024-05-29
Set up Pontus-X devnet and testnet

csillag:
This PR adds / fixes support for Pontus-X devnet and testnet.

They are both hidden layers, so not available in the layer selector, only available via URL.

Things left to do:
 - Configure search suggestions for Pontus-X testnet

### DIFF
--- a/src/app/utils/faucet-links.ts
+++ b/src/app/utils/faucet-links.ts
@@ -9,8 +9,11 @@ const faucetLinks: Partial<Record<Network, Partial<Record<Layer, Partial<Record<
     [Layer.emerald]: { [Ticker.TEST]: `${faucets.oasisTestnet}?paratime=emerald` },
     [Layer.sapphire]: { [Ticker.TEST]: `${faucets.oasisTestnet}?paratime=sapphire` },
     [Layer.cipher]: { [Ticker.TEST]: `${faucets.oasisTestnet}?paratime=cipher` },
+    [Layer.pontusxdev]: {
+      [Ticker.EUROe]: `mailto:contact@delta-dao.com?subject=${encodeURIComponent('Request test tokens for Pontus-X Devnet')}`,
+    },
     [Layer.pontusx]: {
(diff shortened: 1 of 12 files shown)


src/config.ts
```diff
@@ -134,6 +134,26 @@ const pontusxConfig: LayerConfig = {
   type: RuntimeTypes.Evm,
 }
 
+const pontusxTestConfig: LayerConfig = {
```
buberdds, line 137: I am not super familiar with PontusX, but why we cannot have one layer with mainnet, testnet and devenet ?

csillag: > The decision (between deltaDao and Jernej) was that they want to run both their devnet and testnet on our testnet.

Also, they have already deployed their paratimes (both on testnet) accordingly. 

So really, there is not much we can do about this in Explorer. (How we implement the layer selection is another matter, but that will handle that on the pontusx branch.)

src/app/utils/faucet-links.ts
```diff
@@ -9,6 +9,9 @@ const faucetLinks: Partial<Record<Network, Partial<Record<Layer, Partial<Record<
     [Layer.emerald]: { [Ticker.TEST]: `${faucets.oasisTestnet}?paratime=emerald` },
     [Layer.sapphire]: { [Ticker.TEST]: `${faucets.oasisTestnet}?paratime=sapphire` },
     [Layer.cipher]: { [Ticker.TEST]: `${faucets.oasisTestnet}?paratime=cipher` },
+    [Layer.pontusxdev]: {
+      [Ticker.EUROe]: `mailto:contact@delta-dao.com?subject=${encodeURIComponent('Request test tokens')}`,
+    },
     [Layer.pontusx]: {
       [Ticker.EUROe]: `mailto:contact@delta-dao.com?subject=${encodeURIComponent('Request test tokens')}`,
     },
```
lubej, line 17: How would they know, which Layer to send the test tokens to? Wouldn't be nice to include it in the subject, for which Layer it is?

lubej (approved):

PR #1437, opened 2024-05-30
Migrate from grpc.oasis.dev to grpc.oasis.io

lukaw3d:


.changelog/1437.bugfix.md
```diff
@@ -0,0 +1 @@
+Migrate from grpc.oasis.dev to grpc.oasis.io
```
buberdds, line 1: If dev is functional it's rather internal type.

buberdds (approved):

PR #1443, opened 2024-06-05
Avoid showing "NaN%" in the charts when there is no prev data

csillag:
We have some statistics up on the paratime dashboards. However, when comparing to the current data to the previous one (whatever that means for the current setting), if the previous one was 0, when calculating the change, we get a division by zero error, and hence, NaN.

As discussed with @donouwens , in these situations, we should simply hide the percentage value. Like this:

| Before:   | After: |
| ------------- | ------------- |
| image  | image |

### DIFF
--- a/src/app/components/PercentageGain/index.tsx
+++ b/src/app/components/PercentageGain/index.tsx
@@ -13,6 +13,9 @@ interface PercentageGainProps {
  * Negative percentage shows red box with down arrow.
  */
 export const PercentageGain: FC<PercentageGainProps> = ({ earliestValue, latestValue }) => {
+  // If the previous data was zero, we can't divide with it, it would just get us infinity or Nan...
+  if (earliestValue === 0) return
+
   const ratio = (latestValue - earliestValue) / earliestValue
 
   return (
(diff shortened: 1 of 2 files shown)


src/app/components/PercentageGain/index.tsx
```diff
@@ -13,6 +13,9 @@ interface PercentageGainProps {
  * Negative percentage shows red box with down arrow.
  */
 export const PercentageGain: FC<PercentageGainProps> = ({ earliestValue, latestValue }) => {
+  // If there is no previous data, we can't divide with it, it would just get us a Nan...
+  if (!earliestValue) return
```
lukaw3d, line 17: `earliestValue === 0` would make it easier to see "division by zero" issue

lukaw3d (approved):

lukaw3d (changes requested):
Wait this doesn't work like I thought. If it was division by zero it would print Infinity, not NaN

lukaw3d (approved):
Oh, 0/0 is not infinity

csillag:
> Wait this doesn't work like I thought. If it was division by zero it would print Infinity, not NaN

1/0 is `infinity`, but 0/0 is `NaN`. Bottom line, we just shouldn't divide by zero.

src/app/components/PercentageGain/index.tsx
csillag: OK updated.

PR #1444, opened 2024-06-05
Don't show TX error messages for empty error blocks

csillag:
Some transactions have an `error` block like this:

```
"error": {
  "code": 0,
  "module": ""
}
```

An example found by @matevz is available [here](https://explorer.oasis.io/mainnet/sapphire/tx/7224d59e6b74d9b2caf55ab40dfe0757cef4f71c66d3a7aba3de1d267d2ea409).

In these cases, no error message should be shown.

### DIFF
--- a/src/app/hooks/useTxErrorMessage.ts
+++ b/src/app/hooks/useTxErrorMessage.ts
@@ -3,7 +3,11 @@ import { TxError } from '../../oasis-nexus/api'
 
 export const useTxErrorMessage = (error: TxError | undefined): string | undefined => {
   const { t } = useTranslation()
-  if (!error || (!error.module && !error.code)) return undefined
+  if (
+    !error ||
+    (!error.module && !error.code) // TODO: This second check shouldn't be necessary. Remove when https://github.com/oasisprotocol/nexus/issues/704 is fixed.
+  )
+    return undefined
   if (error.module === 'evm' && error.code === 8 && !error.message) {
     // EVM reverted, with missing error message
     return `${t('errors.revertedWithoutMessage')} (${t('errors.code')} ${error.code})`
--- a/src/app/hooks/useTxErrorMessage.ts
+++ b/src/app/hooks/useTxErrorMessage.ts
@@ -3,7 +3,7 @@ import { TxError } from '../../oasis-nexus/api'
 
 export const useTxErrorMessage = (error: TxError | undefined): string | undefined => {
   const { t } = useTranslation()
-  if (!error) return undefined
+  if (!error || (!error.module && !error.code)) return undefined
   if (error.module === 'evm' && error.code === 8 && !error.message) {
     // EVM reverted, with missing error message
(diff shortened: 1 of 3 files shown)


csillag:
This is marked as draft until someone can verify that my interpretation of the situation is correct.

csillag:
> This is marked as draft until someone can verify that my interpretation of the situation is correct.

Yes: @kostko says that

> Yes code 0 is always success.

So, we don't have to show an error message here.

src/app/hooks/useTxErrorMessage.ts
```diff
@@ -3,7 +3,7 @@ import { TxError } from '../../oasis-nexus/api'
 
 export const useTxErrorMessage = (error: TxError | undefined): string | undefined => {
   const { t } = useTranslation()
-  if (!error) return undefined
+  if (!error || (!error.module && !error.code)) return undefined
```
lubej, line 6: Why check for a module? I think simplification should suffice:

```suggestion
  if (!error?.code) return undefined
```

csillag: > Why check for a module? 

Because I don't know if we can assume that all possible modules (i.e. sources of errors) are sane enough to use the 0 error code for success.

There might be some deranged modules that break the convention. 

When hiding error messages, I want to be as narrow an specific as possible. We know that no module and error code 0 is not a problem, so we can hide this, but nothing else. 

If we see more useless error messages later, we can also hide them later...

lukaw3d:
Frontend should not be fixing this

lubej:
> Frontend should not be fixing this

How about we create an issue upstream? 
And merge this quick fix for now. As it takes "ages" to propagate the change on the BE side to mainnet.

lubej (approved):

csillag:
> How about we create an issue upstream?

The discussion with the Nexus guys have been going on for 3 days now.

> And merge this quick fix for now. 

I agree, but waiting for @lukaw3d's decision.

> As it takes "ages" to propagate the change on the BE side to mainnet.

Exactly.

src/app/hooks/useTxErrorMessage.ts
lubej: Maybe add a TODO, if this ever gets resolved upstream.

lukaw3d:
:shrug: seems very rare, and frontend can't correctly fix the transaction anyway: method missing, and status missing

csillag:
Not much progress with [the Nexus issue](https://github.com/oasisprotocol/nexus/issues/704) for the last 3 weeks. I still thing we ought to merge the frontend workaround, at least in order to remove the spurious error message...

src/app/hooks/useTxErrorMessage.ts
csillag: > Maybe add a TODO, if this ever gets resolved upstream.

Added TODO note, referencing the upstream/backend issue.

PR #1445, opened 2024-06-13
Un-hide the Pontus-X Testnet layer

csillag:
This PR enabled showing the Pontus-X Testnet in the layer selector.

Please note that it's still not on the graph.

(I'm not sure how much of a problem that is.)


src/app/utils/route-utils.ts
```diff
@@ -37,6 +37,14 @@ export type SpecifiedPerEnabledLayer<T = any, ExcludeLayers = never> = {
 
 export type SpecifiedPerEnabledRuntime<T = any> = SpecifiedPerEnabledLayer<T, typeof Layer.consensus>
 
+export const hiddenLayers: Layer[] = [Layer.pontusxdev]
+
+console.log('Hidden layers are', hiddenLayers)
```
buberdds, line 42: remove

src/app/utils/route-utils.ts
```diff
@@ -37,6 +37,14 @@ export type SpecifiedPerEnabledLayer<T = any, ExcludeLayers = never> = {
 
 export type SpecifiedPerEnabledRuntime<T = any> = SpecifiedPerEnabledLayer<T, typeof Layer.consensus>
 
+export const hiddenLayers: Layer[] = [Layer.pontusxdev]
```
buberdds, line 40: what's a plan for pontusxdev? It's going to be accessible via direct url only?

src/app/utils/route-utils.ts
csillag: Done.

src/app/utils/route-utils.ts
csillag: Yes, in our release, it's only available via direct URL. (On the pontus-x branch, it's not hidden, so it's available in the layer selector.)

buberdds (approved):

PR #1451, opened 2024-06-17
Provide real data in Consensus Current Epoch snapshot card

buberdds:


src/app/pages/ConsensusDashboardPage/SnapshotEpoch.tsx
```diff
@@ -14,13 +15,27 @@ const StyledBox = styled(Box)(({ theme }) => ({
...
+  let percentageValue = undefined
+
+  if (epochs && epochs[1]?.end_height) {
+    const epochDiffHeight = epochs[1]?.end_height - epochs[1]?.start_height
+
+    if (blockHeight) {
+      const completedBlocksInCurrentEpoch = blockHeight - epochs[0]?.start_height
+      percentageValue = completedBlocksInCurrentEpoch / epochDiffHeight
+    }
+  }
```
lubej, line 38: As this won't really change that often can we memoize it(Due to re-renders)?

And maybe multiply percentageValue in here? As opposed to inside prop passing(`value={percentageValue * 100}`).

lubej (approved):

src/app/pages/ConsensusDashboardPage/SnapshotEpoch.tsx
lukaw3d: I wouldn't memoize. Seems very likely that it would hurt performance: caching and comparing memoization keys will take more CPU than three math operations. And it would hurt readability

PR #1459, opened 2024-06-25
Prepare API layer for merged staking view

buberdds:
I will stop here because there is something wrong with designs.

- account can stake to a validator multiple times and `delegations` endpoint will return one row - view is Ok in this case
- account can start debonding multiple times from a validator.  We can have multiple rows (one row per epoch)
- What `Shares` column represents? summing delegation_shares + debonding_delegation shares doesn't make sense (since these are actually different shares)

Screenshot from 2024-06-25 09-48-23

Impl details: 
- we won't have a dedicated endpoint to show Staking card so we need data from `delegations` and `debonding_delegations` 
- https://nexus.oasis.io/v1/consensus/accounts/oasis1qqnk4au603zs94k0d0n7c0hkx8t4p6r87s60axru/delegations
- https://nexus.oasis.io/v1/consensus/accounts/oasis1qqnk4au603zs94k0d0n7c0hkx8t4p6r87s60axru/debonding_delegations
- [x] group `debonding_delegations` by validator and sum shares and amounts
- [ ] merge `delegations` and `debonding_delegations` group by validator and return custom object with props `amount`, `debondingAmount`, `delegator`,`validator`, `debondingEpochs`, `shares(???)`    
- [ ] as we are losing backend pagination we will use frontend pagination
- [ ] ignore is_total_count_clipped for both endpoints 
- [ ] add additional variant to Delegations table component

Alternatives:
- add badges in the right top header of Staking card (Active, Debonding) instead of (How shares are calc link) and show whatever endpoints provide in both sub views
- show Staking as two tabs next to Transfers, Evetns etc and show whatever endpoints provide

### DIFF
--- a/src/oasis-nexus/api.ts
+++ b/src/oasis-nexus/api.ts
@@ -4,6 +4,7 @@ import axios, { AxiosResponse } from 'axios'
 import { consensusDecimals, getTokensForScope, paraTimesConfig } from '../config'
 import * as generated from './generated/api'
 import { QueryKey, UseQueryOptions, UseQueryResult, useQuery } from '@tanstack/react-query'
+import BigNumber from 'bignumber.js'
 import {
   Account,
   EvmToken,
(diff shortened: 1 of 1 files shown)


buberdds:
we will go with https://github.com/oasisprotocol/explorer/pull/1461

PR #1461, opened 2024-06-27
Consensus: add debonding delegations to Staking card

buberdds:
- still showing validator address instead a name. This will be resolved in the future when validator_media object will be added, or by named accounts
- debonding list can show `children with the same key` warning temporary. Validator address + epoch will be unique in debonding delegations once bugfix is deployed on Nexus side
- filter buttons use styles from other Consensus card not from mockups (not sure why they are different)
- last commit includes changes provided by Tadej


buberdds:
bump

src/app/components/Delegations/index.tsx
```diff
@@ -27,15 +27,18 @@ export const Delegations: FC<DelegationsProps> = ({
   const { t } = useTranslation()
 
   const tableColumns: TableColProps[] = [
-    { key: 'delegator', content: t('common.address') },
+    { key: 'delegator', content: linkType === 'validator' ? t('validator.title') : t('common.address') },
```
lukaw3d, line 30: `key: 'delegator'` is misleading (also in existing code in tableRows)

Ideally it would be specified like this
```ts
    { key: 'delegator', content: t('common.address'), hide: linkType === 'validator' },
    { key: 'validator', content: t('validator.title'), hide: linkType !== 'validator' },
```

src/app/pages/ConsensusAccountDetailsPage/Staking.tsx
```diff
@@ -1,64 +1,75 @@
...
-          <Link
-            href={docs.consensusStaking}
-            rel="noopener noreferrer"
-            target="_blank"
-            sx={{ color: COLORS.brandDark }}
-          >
-            {t('validator.sharesDocs')}
-          </Link>
-        }
+        action={<FilterButtons options={options} value={type} onSelect={type => setType(type)} />}
```
lukaw3d, line 46: This infers a weak type: `FilterButtons<string>` instead of `FilterButtons<DelegationStatus>`.

One thing you can do is
```diff
  type FilterButtonsProps<T> = {
    options: { label: string; value: T }[]
    onSelect: (value: T) => void
-   value?: T
+   value?: NoInfer<T>
  }
```
that then marks `value={type}` as invalid in this line, so:
```diff
- const [type, setType] = useState('staked')
+ const [type, setType] = useState<DelegationStatus>('staked')
```

src/app/components/FilterButtons/index.tsx
```diff
@@ -0,0 +1,42 @@
...
+}
+
+export const FilterButtons = <T,>({ options, onSelect, value }: FilterButtonsProps<T>) => {
+  return (
+    <Box sx={{ display: 'inline-flex' }}>
+      {options.map(option => {
+        const selected = option.value === value
+        return (
+          <Chip
+            key={String(option.value)}
```
lukaw3d, line 19: This is a bit risky if `T` can be anything, including objects. Restrict it with `T extends string`?

lukaw3d (approved):

src/app/components/Delegations/index.tsx
buberdds: hide prop added

buberdds:
will wait for an updated empty state message

buberdds:
will merge as is and update message in a following PR as we will change layout slightly

src/app/components/Table/index.tsx
```diff
@@ -112,18 +114,23 @@ export const Table: FC<TableProps> = ({
...
-                    width: column.width || 'auto',
-                    ...(stickyColumn && !index && isMobile ? stickyColumnStyles : {}),
-                  }}
-                >
-                  {column.content}
-                </TableCell>
-              ))}
+              {columns.map((column, index) => {
+                if (column.hide) {
+                  return null
```
lukaw3d, line 119: Note: if first column is hidden then stickyColumn doesn't work

buberdds: we can remove `stickyColumn` it is not used anywhere

PR #1466, opened 2024-06-28
Account details: add link to validator details. 

csillag:
(This is relevant when we are displaying a consensus account
that belongs to a validator.)


csillag:
Design validated by @donouwens

buberdds:
Designs include:
- some icon next to a validator name
- this should be added in multiple places. I am not sure what are these places after the latest changes in Figma: other than account details we need account list update. What about  search results or account list - vertical list?

PR #1469, opened 2024-07-03
Add gasPrice calculation in runtime tx detail view

lubej:
## Description

Quote from API docs: "For EVM transactions this is calculated as `charged_fee` = `gas_price * gas_used`, where `gas_price = fee / gas_limit`, for compatibility with Ethereum."

## Resources 

image


lubej:
@donouwens could you take a quick glance on how the Gas price is displayed. Particularly the `(0.0000001 ROSE)` part.

src/app/utils/number-utils.ts
```diff
@@ -0,0 +1,14 @@
+import BigNumber from 'bignumber.js'
+
+const GWEI_DENOMINATION = BigNumber(10).pow(9)
+
+export const getGasPrice = ({ fee = '', gasUsed = '' }): string | null => {
+  if (!fee || !gasUsed) {
+    return null
+  }
+
+  return BigNumber(fee).div(BigNumber(gasUsed)).toLocaleString()
```
lubej, line 10: Can `toLocaleString` brake `getPreciseNumberFormat`?

src/app/utils/number-utils.ts
```diff
@@ -0,0 +1,14 @@
+import BigNumber from 'bignumber.js'
+
+const GWEI_DENOMINATION = BigNumber(10).pow(9)
+
+export const getGasPrice = ({ fee = '', gasUsed = '' }): string | null => {
+  if (!fee || !gasUsed) {
```
lukaw3d, line 6: I'd add `gasUsed === '0'`

src/app/utils/number-utils.ts
lukaw3d: Maybe. It's odd that it even works
- `toLocaleString` comes from Object in BigNumber's prototype chain
- Object.toLocaleString natively calls `this.toString`
- BigNumber.toString sometimes returns exponential format (1e+28, 1.234e-35).

toFixed is the safest option
https://github.com/oasisprotocol/explorer/blob/ed691bb3a65963b1c8345dd279454cab56e70b72/src/app/utils/helpers.ts#L113-L119

src/app/pages/RuntimeTransactionDetailPage/index.tsx
```diff
@@ -311,6 +314,27 @@ export const RuntimeTransactionDetailView: FC<{
             })}
           </dd>
 
+          {gasPrice && (
+            <>
+              <dt>{t('common.gasPrice')}</dt>
+              <dd>
+                {t('common.valueInToken', {
+                  ...getPreciseNumberFormat(formatGwei(gasPrice)),
+                  ticker: 'Gwei',
```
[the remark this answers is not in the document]
lukaw3d: Not sure if we have some good established format for printing gas price. Ask internally

https://docs.oasis.io/general/manage-tokens/cli/addressbook/#add
oasis CLI prints "(gas price: 0.0000001 ROSE per gas unit)"

src/app/utils/number-utils.ts
lubej: Yes, `toFixed` looks good. Thanks 👍 

It's true, that `toString` returns exponent format, but looks like `Intl.NumberFormat` correctly parsed it.

src/app/pages/RuntimeTransactionDetailPage/index.tsx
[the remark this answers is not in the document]
lubej: Yes, it does not work for all ParaTimes. But Gwei is something I would expect on Sapphire chain, but "breaks" context on Cipher. I would definitely skip `per gas unit` wording, as that is a given.

Can we maybe display Gwei on ParaTimes that have `type === RuntimeTypes.Evm` in config and `18 decimal spaces` denomination?

src/app/utils/number-utils.ts
```diff
@@ -0,0 +1,19 @@
...
+
+export const getGasPrice = ({ fee = '', gasUsed = '' }): string | null => {
+  if (!fee || !gasUsed || gasUsed === '0') {
+    return null
+  }
+
+  return BigNumber(fee).div(BigNumber(gasUsed)).toFixed()
+}
+
+export const fromBaseUnitsToNano = (wei: string): string => fromBaseUnits(wei, -9)
```
lukaw3d, line 19: This is not taking in base units

lubej: True. But what would wei unit be called then?

lubej: Thinking about it now, you probably meant the function parameter, not the naming. Nevermind.

lukaw3d: > `export const fromBaseUnitsToNano = (valueInBaseUnits: string): string => fromBaseUnits(valueInBaseUnits, -9)`

Both. This takes in values like "0.0000001", clearly not in base units

src/app/pages/RuntimeTransactionDetailPage/index.tsx
```diff
@@ -311,6 +314,18 @@ export const RuntimeTransactionDetailView: FC<{
             })}
           </dd>
 
+          {gasPrice && (
+            <>
+              <dt>{t('common.gasPrice')}</dt>
+              <dd>
+                {t('common.valueInToken', {
+                  ...getPreciseNumberFormat(fromBaseUnitsToNano(gasPrice)),
+                  ticker: `n${tickerName}`,
```
lukaw3d, line 323: Add TODO based on Jernej's comments

src/app/utils/number-utils.ts
lubej: Ah right, for some reason I though on Sapphire 1 ROSE = baseUnit 🤦 

Well should I just name it `unit` then? Or maybe `exaBaseUnit` - but that one is also not correct, as it depends on the ParaTime?

lubej: Did generalize, I hope it's ok now.

lukaw3d: :D 
(maybe toBaseUnits)

lukaw3d (approved):

src/app/utils/number-utils.ts
lubej: `toBaseUnit` would imply that "conversion" is done by 10^18, not by 10^9. Either way, I am keeping it as is.

PR #1474, opened 2024-07-09
Add missing props to Consensus block details  

buberdds:
Sync with incoming Nexus update


src/app/pages/ConsensusBlockDetailPage/index.tsx
```diff
@@ -104,6 +122,20 @@ export const ConsensusBlockDetailView: FC<{
...
+        <>
+          <dt>{t('common.epoch')}</dt>
+          <dd>{block.epoch}</dd>
+        </>
+      )}
+
+      {block.gas_limit && (
+        <>
+          <dt>{t('common.gasLimit')}</dt>
+          <dd>{block.gas_limit.toLocaleString()}</dd>
```
lukaw3d, line 136: toLocaleString won't do anything? `gas_limit: TextBigInt;` is already a string

lukaw3d (approved):

buberdds:
Will wait for another backend update that will add block size limit. This will be more clear for users as there is still a limit for the block (not by gas, but size) when block.gas_limit = "0"

PR #1476, opened 2024-07-09
Fix typo in translation key

lukaw3d:


src/locales/en/translation.json
```diff
@@ -656,7 +656,7 @@
     "emptyDelegationsList": "No one is delegating to this validator.",
     "entityId": "Entity ID",
     "externalLink": "External Link",
-    "externalLinkDescriotion": "Oasis is not responsible for any content on external pages.",
+    "externalLinkDescription": "Oasis is not responsible for any content on\u2028external pages.",
```
buberdds, line 659: why unicode character was added?

lukaw3d: You added it here, but it's invisible: https://github.com/oasisprotocol/explorer/pull/1232
I thought it added a new line in some scenarios, but I actually don't see a difference

buberdds (approved):

PR #1478, opened 2024-07-11
Consensus UI tweaks and API sync

buberdds:
sync with incoming Nexus changes and ui fixes 
- handle "waiting" state in validators 
- add missing props to account list 
- tx types will be needed for iconography and tx list update 
- handle a new prop in tx details

### DIFF
--- a/src/app/pages/ConsensusDashboardPage/SnapshotValidators.tsx
+++ b/src/app/pages/ConsensusDashboardPage/SnapshotValidators.tsx
@@ -12,12 +12,12 @@ type ValidatorsStats = {
   value: number
 }
 
-// TODO: remove if backend will provide validators stats
 function countValidatorsState(t: TFunction, validators: Validator[] | undefined): ValidatorsStats[] {
   if (!validators) return []
 
   return [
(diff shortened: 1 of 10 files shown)


src/app/pages/ConsensusDashboardPage/SnapshotValidators.tsx
```diff
@@ -17,7 +17,8 @@ function countValidatorsState(t: TFunction, validators: Validator[] | undefined)
   if (!validators) return []
 
   return [
-    { label: t('validator.active'), value: validators.filter(v => v.active).length },
+    { label: t('validator.active'), value: validators.filter(v => v.in_validator_set).length },
+    { label: t('validator.waiting'), value: validators.filter(v => !v.in_validator_set && v.active).length },
```
lukaw3d, line 21: Sync API spec again - current description of `v.active` is the same as `v.in_validator_set`

buberdds: description updated

lukaw3d (approved):

PR #1480, opened 2024-07-15
Add missing props to Consensus tx details

buberdds:


src/app/utils/transaction.ts
```diff
@@ -0,0 +1,28 @@
...
+    case ConsensusTxMethod.stakingReclaimEscrow:
+      return transaction.body?.account
+    case ConsensusTxMethod.stakingTransfer:
+      return transaction.body?.to
+    default:
+      return undefined
+  }
+}
+
+export const getConsensusTransactionAmount = (transaction: Transaction) => {
```
lukaw3d, line 18: Maybe this should be used in api.ts and extend Transaction definition; so various lists and details views can simply use it

lukaw3d (approved):

PR #1493, opened 2024-08-08
Automatically throw on 5xx error responses

lukaw3d:
These are currently throwing `{"msg":"storage error: can't scan into dest[5]: cannot scan NULL into *string"}`
https://explorer.dev.oasis.io/testnet/sapphire/address/oasis1qp66ryj9caek77kewkxxvjvkzypljhsdgvm5q34d
https://explorer.dev.oasis.io/search?q=8d5a30f1e586dce3ae7839c5f20c846d6d71b27d185adbacd32778d6c60f55eb
but frontend is ignoring the error

| Before | After |
| --- | --- |
| explorer dev oasis io_testnet_sapphire_address_oasis1qp66ryj9caek77kewkxxvjvkzypljhsdgvm5q34d | localhost_1234_testnet_sapphire_address_oasis1qp66ryj9caek77kewkxxvjvkzypljhsdgvm5q34d |
| explorer dev oasis io_search_q=8d5a30f1e586dce3ae7839c5f20c846d6d71b27d185adbacd32778d6c60f55eb | localhost_1234_search_q=8d5a30f1e586dce3ae7839c5f20c846d6d71b27d185adbacd32778d6c60f55eb |

### DIFF
--- a/src/app/data/oasis-account-names.ts
+++ b/src/app/data/oasis-account-names.ts
@@ -1,5 +1,5 @@
 import axios from 'axios'
-import { useQuery } from '@tanstack/react-query'
+import { useQuery, UseQueryOptions } from '@tanstack/react-query'
 import {
   Layer,
   useGetConsensusAccountsAddresses,
@@ -65,12 +65,11 @@ const getOasisAccountsMetadata = async (network: Network, layer: Layer): Promise
 export const useOasisAccountsMetadata = (
   network: Network,
   layer: Layer,
-  queryOptions: { enabled: boolean; useErrorBoundary?: boolean },
+  queryOptions: UseQueryOptions<AccountData, unknown, AccountData, string[]>,
 ) => {
   return useQuery(['oasisAccounts', network, layer], () => getOasisAccountsMetadata(network, layer), {
-    enabled: queryOptions.enabled,
     staleTime: Infinity,
-    useErrorBoundary: queryOptions.useErrorBoundary,
+    ...queryOptions,
   })
 }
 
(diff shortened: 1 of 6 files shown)


src/app/data/oasis-account-names.ts
```diff
@@ -65,19 +65,20 @@ const getOasisAccountsMetadata = async (network: Network, layer: Layer): Promise
...
+    useErrorBoundary: queryOptions.useErrorBoundary,
   })
 }
 
 export const useOasisAccountMetadata = (
   network: Network,
   layer: Layer,
   address: string,
-  queryOptions: { enabled: boolean },
+  queryOptions: { enabled: boolean; useErrorBoundary?: boolean },
```
lubej, line 81: Can we somehow use underlying `QueryObserverOptions` interface? Looks painful, if we want to add any additional fields to `queryOptions`, so why not just expose everything?

src/index.tsx
```diff
@@ -14,6 +14,11 @@ const queryClient = new QueryClient({
     queries: {
       refetchOnWindowFocus: false,
       retry: false,
+      useErrorBoundary: (error: any) => {
+        // Automatically throw on 5xx errors. Components that want to handle
+        // errors should set `useErrorBoundary: false` in their queries.
+        return error.response?.status >= 500
```
lubej, line 20: Are we sure that every query is covered? In a sense, I see status queries are disabled, which we don't want to crash the app. But could for example one query on account page "crash" the page to show the error boundary?

lubej (approved):

src/index.tsx
lukaw3d: Yeah some 5xx failures could hit page-level error boundary

PR #1495, opened 2024-08-09
Harden github workflow against injection

lukaw3d:


.github/workflows/ci-renovate.yml
```diff
@@ -32,20 +32,21 @@ jobs:
...
-          if [[ ! -f "${{ steps.vars.outputs.FILE_NAME }}" ]]; then
-            echo "Update dependencies" > ${{ steps.vars.outputs.FILE_NAME }}
+          if [[ ! -f "$FILE_NAME" ]]; then
+            echo "Update dependencies" > "$FILE_NAME"
             # Set git user email and name to match author of the last commit.
             git config --local user.email "$(git log --pretty='%ae' -1)"
             git config --local user.name "$(git log --pretty=format:'%an' -1)"
-            git add ${{ steps.vars.outputs.FILE_NAME }}
+            git add "$FILE_NAME"
             git commit --amend --no-edit
```
lubej, line 49: Not sure who is the owner of git add here, but I would prefer it was another commit. As who ever the owner is, will be the co-author of the commit - and it would be just cleaner, to see which changes were made by this workflow - the downside is useless commit though.

lubej (approved):

.github/workflows/ci-renovate.yml
lukaw3d: this only runs against renovate-bot's commit; and it is currently always failing. I'll leave it :D

PR #1523, opened 2024-08-28
Use runtimes tables to render Consensus tx lists

buberdds:
During last sync call we decided to: 

```
For Explorer Consensus MVP, just use the same Txn tables as for Sapphire, 
consider that there will be a lot of empty "To" columns for the global consensus Txn 
```

list context (no `to` and `amount` columns)  
https://60cad6a5.oasis-explorer.pages.dev/mainnet/consensus
account context
https://60cad6a5.oasis-explorer.pages.dev/mainnet/consensus/address/oasis1qqnk4au603zs94k0d0n7c0hkx8t4p6r87s60axru

### DIFF
--- a/src/app/components/Transactions/ConsensusTransactions.tsx
+++ b/src/app/components/Transactions/ConsensusTransactions.tsx
@@ -1,6 +1,10 @@
 import { FC } from 'react'
 import { useTranslation } from 'react-i18next'
+import Box from '@mui/material/Box'
+import { styled } from '@mui/material/styles'
+import ArrowForwardIcon from '@mui/icons-material/ArrowForward'
 import { Transaction } from '../../../oasis-nexus/api'
+import { COLORS } from '../../../styles/theme/colors'
 import { Table, TableCellAlign, TableColProps } from '../../components/Table'
 import { RoundedBalance } from '../../components/RoundedBalance'
 import { TablePaginationProps } from '../Table/TablePagination'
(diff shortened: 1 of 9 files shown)


src/app/components/Transactions/ConsensusTransactions.tsx
```diff
@@ -59,27 +72,64 @@ export const ConsensusTransactions: FC<ConsensusTransactionsProps> = ({
...
+                ) : null,
+                key: 'to',
+              },
+              {
+                align: TableCellAlign.Right,
+                content: getConsensusAmount(t, transaction),
+                key: 'value',
+              },
+            ]
+          : []),
```
lubej, line 128: `verbose` naming is a bit confusing for it, isn't it? Wouldn't it be better to display the columns based on layer? And drop `verbose` flag all together.

lukaw3d: This matches other code. Could refactor in a next PR, but basing it on layer probably won't be enough in matching code elsewhere

lukaw3d:
Was the intention to move "To" and "Amount" into "Details" column in the future?

lukaw3d (approved):

buberdds:
> Was the intention to move "To" and "Amount" into "Details" column in the future? 

`To` - yes, it would be part of details 
`Amount` - no, this data is a separate column. Issue is that in most cases amount prop does not exist in Consensus methods. Oasis Scan is not showing amount column as well because of that. Do you think it's worth adding it back ?

lukaw3d:
Amount is worth adding back into details at least

PR #1530, opened 2024-09-06
Align Consensus tx details endpoint with runtimes API

buberdds:
Can be tested with prod instance. Note: instance is in pretty bad shape atm, but some old txs are indexed already.

sample url: https://pr-1530.oasis-explorer.pages.dev/testnet/consensus/tx/78dbc2d0f43062e44e13d8e3d3d17b2338fb7986f13dd57d3c2045f4cceec2fc

### DIFF
--- a/src/oasis-nexus/api.ts
+++ b/src/oasis-nexus/api.ts
@@ -279,26 +279,30 @@ export const useGetConsensusTransactionsTxHash: typeof generated.useGetConsensus
       ...options?.request,
       transformResponse: [
         ...arrayify(axios.defaults.transformResponse),
-        (transaction: generated.Transaction, headers, status) => {
-          if (status !== 200) return transaction
-          const amount = getConsensusTransactionAmount(transaction)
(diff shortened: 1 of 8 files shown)


src/oasis-nexus/api.ts
```diff
@@ -279,26 +279,30 @@ export const useGetConsensusTransactionsTxHash: typeof generated.useGetConsensus
...
-                ? fromBaseUnits(transaction.body.amount, consensusDecimals)
-                : undefined,
-              amount_change: transaction.body?.amount_change
-                ? fromBaseUnits(transaction.body.amount_change, consensusDecimals)
-                : undefined,
-            },
-            layer: Layer.consensus,
-            network,
-            ticker,
+            ...data,
```
lukaw3d, line 285: Should we support both for a while?
```ts
if (!('transactions' in data)) normalized = {transactions: [data]}
...transform...
if (!('transactions' in data)) return transformed.transactions[0]
return transformed
```

lukaw3d (approved):

PR #1533, opened 2024-09-10
Add tooltip for transaction gas price explanation

lubej:


src/locales/en/translation.json
```diff
@@ -459,7 +459,8 @@
       "txEncrypted": "This transaction is encrypted.",
       "txNotEncrypted": "This transaction is <strong>not</strong> encrypted.",
       "senderTooltipUnavailable": "Sender not available in selected address format",
-      "recipientTooltipUnavailable": "Recipient not available in selected address format"
+      "recipientTooltipUnavailable": "Recipient not available in selected address format",
+      "gasPriceExplanation": "For each transaction, you're charged a certain amount for the gas you use and is measured in nano ROSE(Gwei)."
```
buberdds, line 463: It should be `...measured in nano {{ ticker }}`

Why we say it is gwei? nanoROSE and GWEI are not interchangeable or equal.

PR #1538, opened 2024-09-12
Detect ERC-1167 Minimal Proxy contracts

lukaw3d:
Part of https://github.com/oasisprotocol/explorer/issues/1536

https://explorer.dev.oasis.io/testnet/sapphire/token/0xbA7F3eE55b7939294caFA0F1e096E79722e1BD88

image

This adds some support. Alternatively we could also refactor large parts of verification icon display, but frontend can not override verification icons on token lists without backend changing.

Issue #1536, opened 2024-09-11
[Feature]: Support for minimal proxy contracts (EIP-1167)


CedarMist:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### SUMMARY

Detect EIP-1167 proxies, show information about the target contract (and whether it's verified)

### DESCRIPTION

[EIP-1167](https://eips.ethereum.org/EIPS/eip-1167) minimal proxy contracts have a well known bytecode format, unfortunately they can also not be verified using Sourcify.

This makes it difficult to find information about a contract if it was deployed using minimal proxies (such as clone factories).

For example, the following contract is an EIP-1167 proxy: https://explorer.oasis.io/testnet/sapphire/address/0xbA7F3eE55b7939294caFA0F1e096E79722e1BD88/code#code

The runtime bytecode is `0x363d3d373d3d3d363d73243e9b626bc3ca61795600ed5295377423cd4e695af43d82803e903d91602b57fd5bf3`, the target of this contract is [`0x243e9b626bc3ca61795600ed5295377423cd4e69`](https://explorer.oasis.io/testnet/sapphire/address/0x243e9b626bc3ca61795600ed5295377423cd4e69/code#code)

For more details on the minimal proxy: https://blog.openzeppelin.com/deep-dive-into-the-minimal-proxy-contract

So, while the proxy contract isn't verified, the target of the proxy is verified.

---------

### Suggestion:

 * In the 'Verification' row, if it's an EIP-1167 proxy, show that it's a proxy, and extract the destination contract with a link.

CedarMist:
Etherscan also supports a variety of proxy contracts, including EIP-1967 and EIP-1167.

### EIP-1167
![image](https://github.com/user-attachments/assets/71616c76-7bd0-4404-ad3f-19b49849e77d)

### EIP-1967
![image](https://github.com/user-attachments/assets/357178fc-e13e-4ec0-9d93-8a55f7a4bfe6)

PR #1540, opened 2024-09-12
Link to abi-playground from ERC-1167 Minimal Proxy contracts

lukaw3d:
Part of https://github.com/oasisprotocol/explorer/issues/1536
Extends https://github.com/oasisprotocol/explorer/pull/1538

image

Support inside abi-playground:
image


CedarMist (approved):
Nice :D

PR #1542, opened 2024-09-12
Cleanup

lukaw3d:


src/app/components/RoundedBalance/index.tsx
```diff
@@ -27,7 +26,7 @@ export const RoundedBalance: FC<RoundedBalanceProps> = ({
   compactLargeNumbers,
   scope,
   showSign,
-  ticker,
+  ticker = '',
```
lubej, line 29: Nit: I would prefer if it was `null`.

src/app/components/RoundedBalance/index.tsx
```diff
@@ -94,7 +91,7 @@ export const RoundedBalance: FC<RoundedBalanceProps> = ({
       <Tooltip
         arrow
         placement="top"
-        title={t('common.valueInToken', { ...getPreciseNumberFormat(value), ticker: tickerName })}
+        title={t('common.valueInToken', { ...getPreciseNumberFormat(value), ticker: ticker })}
```
lubej, line 94: Shorthand `ticker: ticker` -> `ticker`? Or do you think it is less readable?

lubej (approved):

src/app/components/RoundedBalance/index.tsx
lukaw3d: I'm not decided. In this case I think it is less readable, more confusable with a function parameter. Other cons:
- I get annoyed when I don't find field assignments when searching for ` <field>:`
  but ` <field>?:`, `['<field>']`, ` ["<field>"]` also break that
- Looks inconsistent if you later add any non-shorthand properties

But I love the shorthand in returns  https://github.com/oasisprotocol/wallet/blob/dc5b9373c3807fa78b563637003922ab718a5b35/src/vendors/oasisscan.ts#L137-L144 :shrug:

src/app/components/RoundedBalance/index.tsx
lukaw3d: Not sure if that could break something. I'll keep this cleanup without logic changes

PR #1544, opened 2024-09-13
Show emitting transaction address in events

lukaw3d:
Useful for account events and block events that come from different transactions

https://explorer.dev.oasis.io/mainnet/sapphire/address/0xC3ecf872F643C6238Aa20673798eed6F7dA199e9/events#events
image

### DIFF
--- a/src/app/components/RuntimeEvents/RuntimeEventDetails.tsx
+++ b/src/app/components/RuntimeEvents/RuntimeEventDetails.tsx
@@ -208,14 +208,8 @@ export const RuntimeEventDetails: FC<{
               fontWeight={400}
             />
             <br />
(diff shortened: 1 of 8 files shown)


src/app/components/RuntimeEvents/RuntimeEventDetails.tsx
```diff
@@ -208,14 +209,8 @@ export const RuntimeEventDetails: FC<{
               fontWeight={400}
             />
             <br />
-            {t('runtimeEvent.fields.emittingContract')}:
-            <br />
-            <WrappedAccountLinkWithAddressSwitch
```
buberdds, line 213: not longer used (import and component)

buberdds (approved):

PR #1546, opened 2024-09-16
Prohibit unknown denominations

lukaw3d:
Related to https://github.com/oasisprotocol/oasis-sdk/pull/1961 and https://github.com/oasisprotocol/internal-sdk/issues/4

https://explorer.dev.oasis.io/testnet/pontusxdev/address/0xdD15ae1e5F410707B965B2832e32cD9e75b6a8e5

| Before | After |
| --- | --- |
| explorer dev oasis io_testnet_pontusxdev_address_0xdD15ae1e5F410707B965B2832e32cD9e75b6a8e5 (1) | localhost_1234_testnet_pontusxdev_address_0xdD15ae1e5F410707B965B2832e32cD9e75b6a8e5 |

### DIFF
--- a/.changelog/1546.bugfix.md
+++ b/.changelog/1546.bugfix.md
@@ -1,0 +1,1 @@
+Prohibit unknown denominations
--- a/src/oasis-nexus/api.ts
+++ b/src/oasis-nexus/api.ts
@@ -219,9 +219,11 @@ const adjustRuntimeTransactionMethod = (
   isLikelyNativeTokenTransfer: boolean | undefined,
 ) => (isLikelyNativeTokenTransfer ? 'accounts.Transfer' : method)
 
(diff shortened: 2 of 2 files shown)


lubej:
Wouldn't it make more sense to show `n/a` instead of `0 n/a`. I assume that if there is an amount greater than 0, denomination would be available?

lukaw3d:
yeah. All though at the tx definitely did not transfer any ROSE, so "0 ROSE" would confer some confidence (pure "n/a" looks like it might be unknown).

"0 n/a" is a mix

lukaw3d:
Removing the number results in ugly code.
e.g. tx.fee type is not optional (tx.amount is) and can't override the type in `declare module './generated/api' {    export interface RuntimeTransaction {`

PR #1564, opened 2024-10-07
Keep eth address to fetch runtimes data

buberdds:
Issue raised by Tadej (while using staker dApp as far as I remember) / for frontend sync meeting. 

acc without `address_preimage` (keep eth addr)
https://pr-1564.oasis-explorer.pages.dev/mainnet/sapphire/address/0x95D683e4365209E5ce64f2e26e00C23207fe7Ec3
https://explorer.dev.oasis.io/mainnet/sapphire/address/0x95D683e4365209E5ce64f2e26e00C23207fe7Ec3

acc with `address_preimage` (works the same way)
https://pr-1564.oasis-explorer.pages.dev/mainnet/sapphire/address/0x8Bba8906467831962Ed5E524F4Bc47733CBF3b27
https://explorer.dev.oasis.io/mainnet/sapphire/address/0x8Bba8906467831962Ed5E524F4Bc47733CBF3b27

Notes: (thx @Andrew7234)
- the address preimage is populated whenever nexus encounters the corresponding evm address of a given oasis style address
- any evm transaction (evm.Create or evm.Call) should create preimage
- the consensus deposit txs are part of the consensusaccounts module and not the evm module, so the tx body of consensus deposit only includes the oasis-style addresses in the tx body and thus Nexus isn’t aware of the corresponding evm address


src/oasis-nexus/api.ts
```diff
@@ -416,7 +427,7 @@ export const useGetRuntimeAccountsAddress: typeof generated.useGetRuntimeAccount
           if (status !== 200) return data
           return groupAccountTokenBalances({
             ...data,
-            address_eth: getEthAddressForAccount(data, address),
+            address_eth: address.eth ?? getEthAddressForAccount(data, address.oasis),
```
lukaw3d, line 430: getEthAddressForAccount was supposed to already be doing this, if second parameter is eth address
fixed in https://github.com/oasisprotocol/explorer/pull/1039
previous issue https://github.com/oasisprotocol/explorer/issues/1037

src/app/pages/RuntimeAccountDetailsPage/hook.ts
```diff
@@ -34,19 +33,13 @@ export const useAccountTransactions = (scope: SearchScope, address: string) => {
...
     layer, // This is OK since consensus has been handled separately
     {
       limit,
       offset: offset,
-      rel: oasisAddress!,
-    },
-    {
-      query: {
-        enabled: !!oasisAddress,
-      },
```
lukaw3d, line 49: @lubej @csillag are checks like this still needed to handle some empty state

src/oasis-nexus/api.ts
```diff
@@ -243,8 +248,14 @@ export const useGetRuntimeTransactions: typeof generated.useGetRuntimeTransactio
...
             ...data,
             transactions: data.transactions.map(tx => {
+              const oasisAddress =
+                params?.rel && isValidEthAddress(params?.rel)
+                  ? getOasisAddressOrNull(params.rel)
+                  : params?.rel
+
               return {
                 ...tx,
+                to_eth: !tx.to_eth && tx.to === oasisAddress ? params?.rel : tx.to_eth,
```
lukaw3d, line 258: I would refactor this as:
```ts
const fallbackMaybeEthAddresses = [params?.rel]
const fallbackEthAddressesMap: { [oasis1Address: string]: `0x${string}` } =
  Object.fromEntries(
    fallbackMaybeEthAddresses.filter(isValidEthAddress).map((eth) => [getOasisAddressOrNull(eth), toChecksumAddress(eth)])
  )

return {
  ...tx,
  to_eth: tx.to_eth ?? fallbackEthAddressesMap[tx.to],
```

src/app/pages/RuntimeAccountDetailsPage/hook.ts
buberdds: oasisAddress used to be async. Since it is sync now and address is always true (loader would fail earlier if not) then I think it is not used to handle empty spaces.

src/oasis-nexus/api.ts
buberdds: if address map is preferable way than we don't need an array, fromEntries, filter, map as we deal with a single address.  Added a new commit with helper function.

lukaw3d: Hm, looks like we really don't have any EVM queries that support multiple address fields. nevermind

lukaw3d:
Something broke search:
- Before: 6 results https://explorer.dev.oasis.io/search?q=oasis1qzmqhh4r8g2u49aja0qsvzu68jxxk8jrw5pfc9ul
- After: 2 results https://pr-1564.oasis-explorer.pages.dev/search?q=oasis1qzmqhh4r8g2u49aja0qsvzu68jxxk8jrw5pfc9ul

buberdds:
Ok, search will behave the same way. Discord test is failing (not related to changes)

lukaw3d:
Ooooh, so searching for "0x95D683e4365209E5ce64f2e26e00C23207fe7Ec3" would have still found the 4 results

buberdds:
Yes, search functionality is not affected by this PR. For now we want to tweak account view with eth address once you navigate from dApp. Will look at search later this week.

src/oasis-nexus/api.ts
```diff
@@ -876,6 +894,11 @@ export const useGetRuntimeEvents: typeof generated.useGetRuntimeEvents = (
                         event.body.amount.Denomination ||
                         getTokensForScope({ network, layer: runtime })[0].ticker,
                     },
+                    owner:
+                      event.body?.owner && event.body.owner === oasisAddress ? params?.rel : event.body.owner,
+                    from:
+                      event.body?.from && event.body.from === oasisAddress ? params?.rel : event.body.from,
+                    to: event.body?.to && event.body.to === oasisAddress ? params?.rel : event.body.to,
```
lukaw3d, line 901: These should be saved into owner_eth, from_eth, to_eth fields if they don't exist

lukaw3d: These don't run for consensus_accounts.undelegate_start event, because this code block was meant to normalize `Amount` field. undelegate_start only has `Shares` field.

I'd rather move it into a block that needlessly runs for all events

.changelog/1564.bugfix.md
```diff
@@ -0,0 +1 @@
+Keep eth address to fetch runtimes data
```
lukaw3d, line 1: I'm not sure what this means

lukaw3d (approved):

src/oasis-nexus/api.ts
buberdds: thx

PR #1566, opened 2024-10-08
Update favicon.svg to be compatible with darkmode

CedarMist:
Uses same icon as from docs.oasis.io


buberdds:
hm this update should go other way around. Docs is not using the correct favicon. The new one was introduced during rebrand and is used by all other products. I am not UI or brand designer so I will leave decision to @donouwens who created favicon.

CedarMist:
iirc the updated favicon from @donouwens was to make it work in dark mode by adding a white background because the blue color was contrasting really badly in dark mode. However, this made the favicon much smaller and more difficult to discern amongst other favicons, and makes the detail of the Oasis logo blurry and harder to see.

This updated favicon works in light mode (with the brand blue color) and dark mode (by turning white), while retaining the detail of the icon and has a size equivalent to the majority of favicons on the web.

tjanez:
> iirc the updated favicon from @donouwens was to make it work in dark mode by adding a white background because the blue color was contrasting really badly in dark mode. However, this made the favicon much smaller and more difficult to discern amongst other favicons, and makes the detail of the Oasis logo blurry and harder to see.

Agreed. Having a smaller Oasis logo because of white background makes the favicon harder to see and doesn't look well compared to favicons of other sites.

> This updated favicon works in light mode (with the brand blue color) and dark mode (by turning white), while retaining the detail of the icon and has a size equivalent to the majority of favicons on the web.

Agreed regarding light mode. However, in the dark mode, the icon didn't turn to white for me on Chromium or Firefox:
Screenshot from 2024-12-13 09-31-01 (Copy)
Screenshot from 2024-12-13 09-32-31 (Copy)

lukaw3d:
Closing due to https://github.com/oasisprotocol/explorer/pull/1870

PR #1575, opened 2024-10-17
Make Pontus-X layer hidden

buberdds:
Opening PR just in case.

Is `pontusxtest` a hidden layer or it is fine that we are listing it in prod Explorer ?


buberdds:
Based on feedback we keep as is "we don't want to hide it completely"

PR #1578, opened 2024-10-18
Prevent crash when network error occur while fetching account names

buberdds:
Make Explorer usable (gracefully handle network errors) when account names feature is not working.

For some reason I cannot access raw.githubusercontent.com today (works when I switch to another Internet provider).
Turns out that we are not handling axios errors correctly when fetching account names. We're not handling potential issues like network issues, server unreachability, or certificate errors. When such error occurs Explorer is unusable.
  
Screenshot from 2024-10-18 10-37-55
can be repro with
https://explorer.oasis.io/mainnet/sapphire -> dev tools -> mainnet_consensus.json -> block request domain

### DIFF
--- a/.changelog/1578.bugfix.md
+++ b/.changelog/1578.bugfix.md
@@ -1,0 +1,1 @@
+Prevent crash when network error occur while fetching account names
--- a/src/app/hooks/useAccountMetadata.ts
+++ b/src/app/hooks/useAccountMetadata.ts
@@ -15,6 +15,7 @@ export const useAccountMetadata = (scope: SearchScope, address: string): Account
   const isPontusX = scope.layer === Layer.pontusxtest || scope.layer === Layer.pontusxdev
   const pontusXData = usePontusXAccountMetadata(address, { enabled: isPontusX })
   const oasisData = useOasisAccountMetadata(scope.network, scope.layer, getOasisAddress(address), {
+    useErrorBoundary: false,
     enabled: !isPontusX,
   })
   return isPontusX ? pontusXData : oasisData


src/app/data/oasis-account-names.ts
```diff
@@ -40,25 +40,39 @@ const dataSources: Record<Network, Partial<Record<Layer, string>>> = {
...
+      // Register the metadata in its native form
+      list.push(metadata)
+      map.set(metadata.address, metadata)
+    })
+    return {
+      map,
+      list,
+    }
+  } catch (error) {
+    if (axios.isAxiosError(error)) {
```
lukaw3d, line 65: I think ignoring errors should be handled per use - and I thought it already was:
- https://github.com/oasisprotocol/explorer/blob/12468641e89f5f06b65b6098fde88f859fed9368/src/app/data/oasis-account-names.ts#L106
- https://github.com/oasisprotocol/explorer/blob/12468641e89f5f06b65b6098fde88f859fed9368/src/app/components/Account/AccountLink.tsx#L93

buberdds: this is weird, but Ok. our pattern is "disable suspense's default behavior of throwing errors to the error boundary" and we do it per use like here https://github.com/oasisprotocol/explorer/blob/master/src/app/data/oasis-account-names.ts#L104.

lukaw3d: even worse:
- react-query's default behavior is: don't throw error (v5 useErrorBoundary -> throwOnError)
- we change defaults to throw on 5xx, and on disconnect (because no status code, because no CORS, because cloudflare) https://github.com/oasisprotocol/explorer/blob/2955d9df60a282414575c214412a4e906891a0b7/src/index.tsx#L17-L27
- and override the default when result isn't important (e.g. coingecko price)

I'd prefer if we ignored errors inside the displaying component. E.g. transfer value UI component knows that it isn't criticial if coingecko fails. But I took a shortcut in https://github.com/oasisprotocol/explorer/commit/bc66fc4d7230b9416527d40da54f8c5d66d94c28 and ignored errors inside the hooks, because it would be too much work to refactor.

Current
```
<SearchResultsPage
  const tokenPrices =
    useAllTokenPrices() ->
      useGetTokenPricesFromGecko() ->
        // Ignore errors. hopefully every use in every parent hook and all its sub-component isn't essential
        useQuery(useErrorBoundary: false)
  <SearchResultsView tokenPrices={tokenPrices}
    <GlobalSearchResultsView tokenPrices={tokenPrices}
      <SearchResultsList tokenPrices={tokenPrices}
        <RuntimeTransactionDetailView tokenPrices={tokenPrices}
          <dd><CurrentFiatValue amount={transaction.amount} ...tokenPrices /></dd>
```

Prefer
```
<SearchResultsPage
  <SearchResultsView
    <GlobalSearchResultsView
      <SearchResultsList
        <RuntimeTransactionDetailView
          const tokenPrices = 
            // converted tx value display isn't essential, ignore errors
            useAllTokenPrices(useErrorBoundary: false) ->
              useGetTokenPricesFromGecko(useErrorBoundary) ->
                useQuery(useErrorBoundary)
          <dd><CurrentFiatValue amount={transaction.amount} ...tokenPrices /></dd>
```

But this conflicts with my other preference: most UI components should be pure, so lets move fetching towards Page components.

:shrug:

lukaw3d (approved):

Issue #1589, opened 2024-10-23
[UX]: Update Consensus tx method iconography


buberdds:
### IMPACT

- [ ] High
- [ ] Medium
- [X] Low

### URGENCY

- [ ] Critical
- [ ] Major
- [X] Medium
- [ ] Minor

### DESCRIPTION OF ISSUE

Iconography is missing lots of tx methods. In addition, some are not clear to users based on internal feedback (for example Start undelegate) 

consensus.Meta
keymanager.PublishEphemeralSecret
keymanager.PublishMasterSecret
keymanager.UpdatePolicy
registry.DeregisterEntity
registry.ProveFreshness
registry.UnfreezeNode
roothash.Evidence
roothash.SubmitMsg
staking.Burn

and new types added in v24.0 (not handled by Nexus yet, but we will need it anyway)        
keymanager/churp.Apply        
keymanager/churp.Confirm        
keymanager/churp.Create
keymanager/churp.Update    
vault.AuthorizeAction      
vault.CancelAction   vault.Create

### DEVICE / VIEWPORT

- [ ] Desktop
- [ ] Mobile

donouwens:
@buberdds here's the icon proposal: https://docs.google.com/document/d/1-lq3L6uP8OIUpn3f0ftBuL-ybOvPv-BWWfkhzK7kQ-o/edit?usp=sharing 

Let me know if this covers all, and if you have any concerns.

PR #1565, opened 2024-10-08
Display EVM function name in verified contract calls

lukaw3d:
Part of https://github.com/oasisprotocol/explorer/issues/1222
(extracted small subset of https://github.com/oasisprotocol/explorer/pull/1516)

https://pr-1565.oasis-explorer.pages.dev/mainnet/sapphire/address/0x5b82acbDe21bda0E9E277BF29A0F84f8deB5F1A7?page=100

| Before | After |
| --- | --- |
| explorer dev oasis io_mainnet_sapphire_address_0x5b82acbDe21bda0E9E277BF29A0F84f8deB5F1A7_page=100 | localhost_1234_mainnet_sapphire_address_0x5b82acbDe21bda0E9E277BF29A0F84f8deB5F1A7_page=100 |

PR #1624, opened 2024-11-04
Update Consensus tx methods labels

buberdds:
- Based on a feedback from Slack we want to align labels with runtimes ("Let’s stick with title case for consistency, please.")
- other adjustment are postponed after UI framework work https://github.com/oasisprotocol/explorer/issues/1589#issuecomment-2451499484


lubej (approved):

lubej:
It would also be possible to just add `text-transform: uppercase` to labels - which would be simpler IMO.

buberdds:
`text-transform` was internationally removed in https://github.com/oasisprotocol/explorer/pull/1565#discussion_r1792189537  to render (EVM function names)

Issue #1597, opened 2024-10-23
[Feature]: Localnet support


ptrus:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### SUMMARY

Introduce a Localnet mode, for using explorer with local networks

### DESCRIPTION

Using Explorer (and Nexus) in the dApp development docker image was suggested in [GitHub Issue #632](https://github.com/oasisprotocol/oasis-web3-gateway/issues/632) as it significantly improves debugging and development experience.

Initial support was implemented in [PR #637](https://github.com/oasisprotocol/oasis-web3-gateway/pull/637), but for a better user experience, we should introduce some minimal support in the frontend as well.

### Proposal: introduce localnet mode (configurable vie .env) with the following changes:
- Support configuring Paratimes through .env, allowing localnet setups to have either a single Paratime or any combination of Paratimes enabled.
- Completely hide non-configured Paratimes (e.g., instead of showing "Coming soon," don't display other Paratimes at all).
- Remove "Verify through Sourcify" and other references to Sourcify (since Sorucfy cannot be used with local network).
- Possibly remove "Transactions per day, week, month" - anything that needs at least a day of data to show meaningful information. Since localnets are usually short-lived, there isn't much historical data, which makes these graphs appear empty.
- Don't display Network Change Proposals
- Decide on a color theme for localnet (similar to how we use blue for mainnet and yellow for testnet).

Note: This list is kept minimal for now, as we don't want to spend too much time on this - just the bare minimum needed for better experience. If any changes look complex we can also postpone them. In the future, we might introduce more localnet-specific features if this becomes popular and widely used. The only "blocker" issue at this point is configuring the paratimes.

I don't have a strong preference on whether we introduce a "localnet" mode that enables all these changes at once, or if each of these changes should have its own separate configuration option.

---

Btw for anyone interested, initial localnet support already works and can be tested via (https://github.com/oasisprotocol/oasis-web3-gateway/pull/637/files):

```bash
$ git clone git@github.com:oasisprotocol/oasis-web3-gateway.git
$ git checkout ptrus/feature/docker-localnet
$ docker build -t sapphire:test -f  docker/sapphire-localnet/Dockerfile .
$ docker run  docker run -e -p5432:5432 -p8544-8547:8544-8547 -p3000:3000  --rm --name test -ti sapphire:test

# Once the network is initialized (takes ~1 minute) visit localhost:3000
```

PR #1628, opened 2024-11-07
Localnet support

buberdds:
Targets https://github.com/oasisprotocol/explorer/issues/1597

As this is more freestyle I need to get feedback from backend folks too.

Control localnet via .env file `# LOCALNET SETTINGS`

Screenshot from 2024-11-07 11-01-34


ptrus:
Nice, this works! I tested it in https://github.com/oasisprotocol/oasis-web3-gateway/pull/641

ptrus (approved):

src/app/components/ContractVerificationIcon/AbiPlaygroundLink.tsx
```diff
@@ -21,6 +21,10 @@ export const AbiPlaygroundLink: FC<{
       [Layer.emerald]: `${externalLinks.dapps.abiPlayground}?network=42261&contractAddress=${address_eth}`,
       [Layer.sapphire]: `${externalLinks.dapps.abiPlayground}?network=23295&contractAddress=${address_eth}`,
     },
+    [Network.localnet]: {
+      [Layer.emerald]: '',
+      [Layer.sapphire]: '',
```
lukaw3d, line 26: `${externalLinks.dapps.abiPlayground}?network=23293&contractAddress=${address_eth}`

abi-playground supports it now https://github.com/oasisprotocol/abi-playground/commit/49facdc2e49ccbbf4d2074662c920ee64fa16805#diff-179ba0e03b98cb6ec52dd161e395e58906c36fba6bda405e160ea195ba9424e0L61 but localnet doesn't have sourcify verification, so it won't get activated

PR #1633, opened 2024-11-13
Enable search by validator name

buberdds:
branch
https://pr-1633.oasis-explorer.pages.dev/mainnet/consensus/search?q=dust
master
https://explorer.dev.oasis.io/mainnet/consensus/search?q=dust

### DIFF
--- a/src/app/components/Validators/ValidatorLink.tsx
+++ b/src/app/components/Validators/ValidatorLink.tsx
@@ -15,7 +15,7 @@ type ValidatorLinkProps = {
   name?: string
   network: Network
   alwaysTrim?: boolean
-  highlightedPart?: string
+  highlightedPartOfName?: string
 }
 
 export const ValidatorLink: FC<ValidatorLinkProps> = ({
@@ -23,7 +23,7 @@ export const ValidatorLink: FC<ValidatorLinkProps> = ({
   name,
   network,
   alwaysTrim,
-  highlightedPart,
+  highlightedPartOfName,
 }) => {
   const { isTablet } = useScreenSize()
   const to = RouteUtils.getValidatorRoute(network, address)
@@ -36,7 +36,7 @@ export const ValidatorLink: FC<ValidatorLinkProps> = ({
           address={address}
           name={name || validatorName}
           to={to}
-          highlightedPart={highlightedPart}
+          highlightedPart={highlightedPartOfName}
         />
       ) : (
         <DesktopValidatorLink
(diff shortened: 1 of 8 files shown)


src/app/hooks/useSearchForValidatorByName.ts
```diff
@@ -0,0 +1,45 @@
...
+  return matchedAddresses
+}
+
+export const useSearchForValidatorByName = (
+  network: Network,
+  nameFragment: string | undefined,
+): AccountNameSearchResults => {
+  const { isLoading, isError, error, data } = useGetConsensusValidatorsAddressNameMap(network)
+  if (isError) {
+    console.log('Failed to load Validator data', error)
```
lubej, line 29: Do we want `console.log` here?

src/app/pages/SearchResultsPage/hooks.ts
```diff
@@ -234,6 +235,21 @@ export function useNamedAccountConditionally(
   }
 }
 
+export function useNamedValidatorConditionally(nameFragment: string | undefined) {
+  const queries = RouteUtils.getEnabledNetworksForLayer(Layer.consensus).map(network =>
+    // eslint-disable-next-line react-hooks/rules-of-hooks
+    useSearchForValidatorByName(network, nameFragment),
```
lubej, line 241: Don't particularly like this, it should be something like `useSearchForValidatorByNames`, which would receive an array of networks.

src/app/components/Validators/ValidatorLink.tsx
```diff
@@ -15,15 +15,15 @@ type ValidatorLinkProps = {
...
+  highlightedPartOfName?: string
 }
 
 export const ValidatorLink: FC<ValidatorLinkProps> = ({
   address,
   name,
   network,
   alwaysTrim,
-  highlightedPart,
+  highlightedPartOfName,
```
lubej, line 26: Just a nit. But naming doesn't seems consistent, I would keep `highlightedPart`.

lubej (approved):

src/app/hooks/useSearchForValidatorByName.ts
buberdds: removed

src/app/pages/SearchResultsPage/hooks.ts
buberdds: Name updated, but I will keep that pattern where we create a hook just to brake a rule of hooks. This is how entire search was built. Will create an issue to refactor all at once.

buberdds: https://github.com/oasisprotocol/explorer/issues/1635

PR #1638, opened 2024-12-02
Use correct query param in token transfers card

buberdds:
- Waits for Nexus as query is missing index https://github.com/oasisprotocol/nexus/issues/804

Requested by @ptrus 

> The contract_address only returns event that are emitted by the contract. So this is what we want on the transfer page for contracts.

sample url: https://pr-1638.oasis-explorer.pages.dev/mainnet/sapphire/token/0x6665a6Cae3F52959f0f653E3D04270D54e6f13d8

### DIFF
--- a/.changelog/1638.bugfix.md
+++ b/.changelog/1638.bugfix.md
@@ -1,0 +1,1 @@
+Use correct query param in token Transfers card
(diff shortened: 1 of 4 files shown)


ptrus:
This PR has the backend part: https://github.com/oasisprotocol/nexus/pull/808

ptrus (approved):

PR #1642, opened 2024-12-03
Turn off polling in tokens list

buberdds:
- Limit number of request as prd endpoint is not stable
- Refreshing this list every 5 sec when it is sorted by market cap or holders is not needed


lubej (approved):
I don't think that it does much difference, but sure. Unless the deploy is really near.

Issue #528, opened 2023-06-14
[Bug]: Status of failed consensus deposits is not communicated


kostko:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### SUMMARY

When a consensus deposit fails, the explorer does not indicate that.

### STEPS TO REPRODUCE

1. Create a deposit transaction in the runtime, e.g. using the CLI, such that the deposit will in fact fail (e.g. because the user does not have enough allowance or balance in the consensus layer). For example I used `oasis account deposit 10000`.
2. The CLI will report that the deposit transaction has succeeded, but the deposit has in fact failed (this is only known in the next round):
```
Broadcasting transaction...
Transaction included in block successfully.
Round:            550814
Transaction hash: c72d51780c9db7614fb4419d4c82a711ac0c3a430f2737e1f38961d7d019229a
Execution successful.
Waiting for deposit result...
Error: deposit failed with error code 5 from module staking
```
3. Looking at the [block explorer for this transaction](https://explorer.stg.oasis.io/mainnet/sapphire/tx/c72d51780c9db7614fb4419d4c82a711ac0c3a430f2737e1f38961d7d019229a) it just shows success as status. While this is technically correct when showing the status of the transaction itself, the deposit itself failed and this is not indicated anywhere on the explorer.

### EXPECTED BEHAVIOR

Some indication that the deposit itself failed even though the transaction succeeded.

lukaw3d:
@donouwens with current API we can not tell if "but the deposit failed" (an event in the next block). If we could tell then I would argue for red color schema.

more accurate message example:
![image](https://github.com/oasisprotocol/explorer/assets/3758846/38877f68-8a02-4a9c-b373-093d435768e8)

kostko:
I think this looks ok. For the message it would be great if we could say (for deposits): "Deposit has started, final result will be known in the next block." or similar. Then instead of "Partial success" it could also say "Started" or similar.

ptrus:
Starting with https://github.com/oasisprotocol/nexus/pull/820 Nexus now handles these transactions (see the changelog for description on how it's handled).

It also works on explorer without additional changes. The transaction result will be marked as 'unknown'  until the event is received, and then it will be marked as either successful or failed.

However it would be great to also handle this on the explorer side and display the nice messages/icons as were discussed above.

lukaw3d:
current solution could be confusing: our tx status doesn't match grpc and oasisscan.

And we only show caveats while status is pending

[...]

PR #1647, opened 2024-12-06
 Handle runtime consensus transactions status

buberdds:
Sync with https://github.com/oasisprotocol/nexus/pull/820

Fixes https://github.com/oasisprotocol/explorer/issues/528

For some runtime Consensus.* methods we want to show more meaningful status when `success` prop does not exist.

Screenshot from 2024-12-06 11-05-56
Screenshot from 2024-12-06 11-21-07

### DIFF
--- a/src/locales/en/translation.json
+++ b/src/locales/en/translation.json
@@ -459,13 +459,19 @@
     }
   },
   "transaction": {
+    "delegate": "Delegate",
+    "deposit": "Deposit",
     "header": "Transaction",
-    "warningMultipleTransactionsSameHash": "Please make sure you're looking at the right transaction. There is more than one transaction with this hash, which is extremely rare. We are showing the most recent successful one.",
     "rawData": "Raw Data",
+    "started": "Started",
+    "startedDescription": "{{ method }} has started, final result will be known in the next block",
     "tooltips": {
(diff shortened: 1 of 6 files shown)


lubej (approved):

src/locales/en/translation.json
```diff
@@ -453,13 +453,19 @@
     }
   },
   "transaction": {
+    "delegate": "Delegate",
+    "deposit": "Deposit",
     "header": "Transaction",
-    "warningMultipleTransactionsSameHash": "Please make sure you're looking at the right transaction. There is more than one transaction with this hash, which is extremely rare. We are showing the most recent successful one.",
     "rawData": "Raw Data",
+    "started": "Started",
+    "startedDescription": "{{ method }} has started, final result will be known in the next block",
```
lukaw3d, line 461: "Undelegate has started" is a misleading combination

buberdds: What label would be more clear for you exactly?

lukaw3d: Not sure. But we've been using "Start to undelegate" and "Undelegate finished" 14 days later for successful events. And this is the start-of-the-start-to

PR #30, opened 2022-12-21
Dashboard: update labels in latest transactions

buberdds:
RuntimeTransactionMethod will be replaced by auto generated types, but they are not defined on BE yet.
RuntimeTransactionLabel needs to be a component to access translation hooks (I will add i18next today)

PR #1679, opened 2025-01-09
Add support for filtering TX lists for method type

csillag:
This PR exposes the runtime tx method search feature added to Nexus in https://github.com/oasisprotocol/nexus/pull/824.

It adds support for filtering runtime TX lists for method type.

Depends on:
- [x] https://github.com/oasisprotocol/nexus/pull/890
 
It also adds the same feature for consensus TX lists, for which the Nexus support has been there for a while.

It works on all consensus and runtime dashboards, transaction list, also including transactions for blocks, accounts and validators.

I suggest reviewing the PR commit by commit, because there are multiple semi-independent small changes.

## The account transactions page looks like this:

image

Or this, when there are no results: 

image

## Latest transactions page, on mobile:

image

## For consensus:

image

### DIFF
--- a/src/app/components/ConsensusTransactionMethod/index.tsx
+++ b/src/app/components/ConsensusTransactionMethod/index.tsx
@@ -18,6 +18,8 @@ import Tooltip from '@mui/material/Tooltip'
 import { tooltipDelay } from '../../../styles/theme'
 import { ConsensusTxMethod } from '../../../oasis-nexus/api'
 import { COLORS } from '../../../styles/theme/colors'
+import { SelectOptionBase } from '../Select'
+import { exhaustedTypeWarning } from '../../../types/errors'
 
 type MethodIconProps = {
   border?: boolean
@@ -171,30 +173,9 @@ const MethodIconWithTruncatedLabel: FC<MethodIconProps> = props => {
 }
 
 const getConsensusTransactionLabel = (t: TFunction, method: ConsensusTxMethod | undefined): string => {
-  /**
-   * TODO: Missing values:
-   *
-   *  consensusMeta
-   *  keymanagerPublishEphemeralSecret
-   *  keymanagerPublishMasterSecret
-   *  keymanagerUpdatePolicy
-   *  registryDeregisterEntity
-   *  registryProveFreshness
-   *  registryUnfreezeNode
-   *  roothashEvidence
-   *  roothashSubmitMsg
-   *  stakingBurn
-   *  keymanager/churpApply
-   *  keymanager/churpConfirm
-   *  keymanager/churpCreate
-   *  keymanager/churpUpdate
-   *  vaultAuthorizeAction
-   *  vaultCancelAction
-   *  vaultCreate
-   */
-
   // Please note: when updating this, keep it in sync
   // with the knownConsensusTxMethods array below!
+
   switch (method) {
     case ConsensusTxMethod.stakingTransfer:
       return t('transactions.method.stakingTransfer')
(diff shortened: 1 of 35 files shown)


csillag:
As discussed with @donouwens, the type filter list for the consensus layer is too long to be convenient. Ideally, it should be replaced with a combo-box with type-ahead search.

However, we plan to support multiple values anyway (later, when nexus will add support for that), so we will replace the widget with a searchable combo-box at that time.

Which means, the current version is as good as it will get, in the current round of implementation.

src/app/components/ConsensusTransactionMethod/index.tsx
```diff
@@ -170,6 +171,128 @@ const MethodIconWithTruncatedLabel: FC<MethodIconProps> = props => {
   )
 }
 
+const getConsensusTransactionLabel = (t: TFunction, method: ConsensusTxMethod | undefined): string => {
+  // Please note: when updating this, keep it in sync
+  // with the knownConsensusTxMethods array below!
+
+  switch (method) {
```
lubej, line 178: Wouldn't this be better if there was an object map, having O(1) time complexity looking up each of the elements.

src/app/components/Select/index.tsx
```diff
@@ -96,6 +97,54 @@ const TertiaryButton = forwardRef(
   },
 )
 
+export const StyledLightSelectButton = styled(StyledSelectButton)(() => ({
```
lubej, line 100: Not a fan of this, creating new components, rather I would add some kind of prop to `StyledSelectButton`, that would change the styling based on that via `slotProps`.

src/app/components/Select/index.tsx
```diff
@@ -96,6 +97,54 @@ const TertiaryButton = forwardRef(
...
+}))
+
+export const StyledLightSelectOption = styled(StyledSelectOption)(() => ({
+  color: COLORS.grayDark,
+  [`&:hover:not(.${optionClasses.disabled}),
+  &.${optionClasses.highlighted}`]: {
+    backgroundColor: COLORS.grayMediumLight,
+  },
+  [`&.${optionClasses.disabled}`]: {
+    // backgroundColor: COLORS.white,
```
lubej, line 121: Remove dead code

src/app/components/Select/index.tsx
```diff
@@ -96,6 +97,54 @@ const TertiaryButton = forwardRef(
...
+  },
+}))
+
+const LightTertiaryButton = forwardRef(
```
lubej, line 125: This duplication seems unnecessary, I would rather see css changes based on props.

src/app/components/SubPageCard/index.tsx
```diff
@@ -14,6 +14,10 @@ type StyledComponentProps = {
   title?: ReactNode
   subheader?: string
   action?: ReactNode
+  /**
+   * An optional second title which will be displayed under title / subheader / action
+   */
+  title2?: ReactNode
```
lubej, line 20: `subtitle` maybe?

src/app/components/Table/index.tsx
```diff
@@ -166,6 +175,7 @@ export const Table: FC<TableProps> = ({
           <TablePagination {...pagination} />
         </Box>
       )}
+      {!isLoading && !rows?.length && emptyMessage && <CardEmptyState label={emptyMessage} />}
```
lubej, line 178: This seems like it would render the `<table />` and then the empty message bellow it?

src/app/components/ConsensusTransactionMethod/index.tsx
csillag: Maybe. We have the exact same system for Runtime transactions, [here](https://github.com/oasisprotocol/explorer/blob/master/src/app/components/RuntimeTransactionMethod/index.tsx#L16), introduced [ages ago](https://github.com/oasisprotocol/explorer/pull/30/files#diff-e116fc2f1020c3761f1b254a9b5f32cd8ab21a4036827ef69b0f6bc46fd006b8). If we want to refactor this, it should be done consistently, on both ends.

However it's not entirely trivial because we need to rely on a `t` function, which is _not_ available by default in the module namespace, so we can only prepare the object map when the first call happens. Then maybe we can try to cache and reuse it, but I feel that this might be premature optimization and opens the possibility of race conditions at the time of first rendering... not sure it's worth the effort.

What do you think?

src/app/components/Select/index.tsx
csillag: Removed

src/app/components/SubPageCard/index.tsx
csillag: The problem is that currently we have `title`, `subheader` (and `action`), all displayed in the same row.
If we introduced `subtitle`, it would be very easy to accidentally confuse it with `subheader`.

That was my thinking, but I don't have a strong opinion. If you think subtitle would be better, I am open to rename it. What do you think?

src/app/components/Table/index.tsx
csillag: The table header, and the empty message, yes. Like this:

image

I find this more useful than hiding the whole table.

In any case, this was required in the initial version, when the filter widget for the TX type was placed in the table column header for type. In that arrangement, it was absolutely necessary that the table header stays, even if there is no data, so that we can change the filter to _get_ data. But we have changed the design, so now we could live with a disappearing table. Would you prefer that?

src/app/components/ConsensusTransactionMethod/index.tsx
```diff
@@ -170,6 +171,128 @@ const MethodIconWithTruncatedLabel: FC<MethodIconProps> = props => {
...
+    case ConsensusTxMethod.stakingBurn:
+      return t('transactions.method.stakingBurn')
+    case ConsensusTxMethod.vaultAuthorizeAction:
+      return t('transactions.method.vault.authorizeAction')
+    case ConsensusTxMethod.vaultCancelAction:
+      return t('transactions.method.vault.cancelAction')
+    case ConsensusTxMethod.vaultCreate:
+      return t('transactions.method.vault.create')
+    default:
+      return method || t('common.unknown')
```
lukaw3d, line 246: Use exhaustedTypeWarning

src/app/components/ConsensusTransactionMethod/index.tsx
```diff
@@ -170,6 +171,128 @@ const MethodIconWithTruncatedLabel: FC<MethodIconProps> = props => {
...
+  ConsensusTxMethod.roothashSubmitMsg,
+  ConsensusTxMethod.stakingAddEscrow,
+  ConsensusTxMethod.stakingReclaimEscrow,
+  ConsensusTxMethod.governanceSubmitProposal,
+  ConsensusTxMethod.stakingTransfer,
+  ConsensusTxMethod.beaconVRFProve,
+  ConsensusTxMethod.vaultAuthorizeAction,
+  ConsensusTxMethod.vaultCancelAction,
+  ConsensusTxMethod.vaultCreate,
+]
```
lukaw3d, line 286:
```suggestion
// List of known consensus ts types, to offer in filter
// Please keep them in alphabetical order
const knownConsensusTxMethods = [
  ConsensusTxMethod.stakingAllow,
  ConsensusTxMethod.stakingAmendCommissionSchedule,
  ConsensusTxMethod.governanceCastVote,
  ConsensusTxMethod.stakingBurn,
  ConsensusTxMethod.consensusMeta,
  ConsensusTxMethod.stakingWithdraw,
  ConsensusTxMethod.registryDeregisterEntity,
  ConsensusTxMethod.roothashExecutorCommit,
  ConsensusTxMethod.roothashExecutorProposerTimeout,
  ConsensusTxMethod['keymanager/churpApply'],
  ConsensusTxMethod['keymanager/churpConfirm'],
  ConsensusTxMethod['keymanager/churpCreate'],
  ConsensusTxMethod['keymanager/churpUpdate'],
  ConsensusTxMethod.keymanagerPublishEphemeralSecret,
  ConsensusTxMethod.keymanagerPublishMasterSecret,
  ConsensusTxMethod.keymanagerUpdatePolicy,
  ConsensusTxMethod.beaconPVSSCommit,
  ConsensusTxMethod.beaconPVSSReveal,
  ConsensusTxMethod.registryRegisterEntity,
  ConsensusTxMethod.registryRegisterNode,
  ConsensusTxMethod.registryRegisterRuntime,
  ConsensusTxMethod.registryProveFreshness,
  ConsensusTxMethod.registryUnfreezeNode,
  ConsensusTxMethod.roothashEvidence,
  ConsensusTxMethod.roothashSubmitMsg,
  ConsensusTxMethod.stakingAddEscrow,
  ConsensusTxMethod.stakingReclaimEscrow,
  ConsensusTxMethod.governanceSubmitProposal,
  ConsensusTxMethod.stakingTransfer,
  ConsensusTxMethod.beaconVRFProve,
  ConsensusTxMethod.vaultAuthorizeAction,
  ConsensusTxMethod.vaultCancelAction,
  ConsensusTxMethod.vaultCreate,
] satisfies ConsensusTxMethod[]

// eslint-disable-next-line @typescript-eslint/no-unused-vars
const typeTestExhaustiveArray =
  undefined as unknown as ConsensusTxMethod satisfies (typeof knownConsensusTxMethods)[number]
```

lukaw3d (commented):
Typescript can make sure these are kept in sync

src/app/components/ConsensusTransactionMethod/index.tsx
csillag: Added

src/app/components/Select/index.tsx
csillag: I did this now, but I am not sure if that made the code simpler or not.

src/app/components/Select/index.tsx
csillag: I did this now, but I am not sure if that made the code simpler or not.

src/app/components/Table/index.tsx
csillag: Since there are no longer any filtering in the headers, so this is no longer necessary, so I have revised this decision. Now the table is fully hidden again.

csillag:
Please note that filtering for runtime tx type can get some inconsistent results until https://github.com/oasisprotocol/nexus/issues/883 is implemented.

csillag:
This will be greatly improved when https://github.com/oasisprotocol/nexus/pull/890 lands, because it will make it possible to search for the "transfer" type coherently.

csillag:
> This will be greatly improved when [oasisprotocol/nexus#890](https://github.com/oasisprotocol/nexus/pull/890) lands, because it will make it possible to search for the "transfer" type coherently.

Landed! (Now waiting for deployment)

src/app/components/ConsensusTransactionMethod/index.tsx
lukaw3d: Dont cache it. It will cause headaches if we add language switching

src/app/components/ConsensusTransactionMethod/index.tsx
```diff
@@ -178,135 +309,42 @@ const getConsensusTransactionMethod = (
...
     default:
-      return (
-        <MethodIcon
-          color="gray"
-          icon={<QuestionMarkIcon />}
-          label={method || t('common.unknown')}
-          {...props}
-        />
-      )
+      return <MethodIcon color="gray" icon={<QuestionMarkIcon />} label={label} {...props} />
```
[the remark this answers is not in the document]
csillag: Added

src/app/components/ConsensusTransactionMethod/index.tsx
```diff
@@ -320,3 +358,9 @@ export const ConsensusTransactionMethod: FC<ConsensusTransactionMethodProps> = (
 
   return <>{getConsensusTransactionMethod(t, method, truncate)}</>
 }
+
+export type WantedConsensusTxMethod = ConsensusTxMethod | 'all'
```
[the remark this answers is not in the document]
csillag: Renamed to `ConsensusTxMethodFilterOption`.

src/app/components/RuntimeTransactionMethod/index.tsx
```diff
@@ -117,3 +141,18 @@ export const RuntimeTransactionMethod: FC<RuntimeTransactionLabelProps> = ({ tra
...
 }
+
+export const getRuntimeTransactionMethodFilteringParam = (
+  method: string,
+): Partial<GetRuntimeTransactionsParams> => {
+  switch (method) {
+    case 'all':
+      return {}
+    case 'accounts.Transfer':
+      return { method: 'native_transfers' }
```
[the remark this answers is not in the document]
csillag: Added explanation, copied the documentation of the magic values, and added link to API spec.

src/app/components/Table/index.tsx
```diff
@@ -101,71 +109,77 @@ export const Table: FC<TableProps> = ({
...
   extraHorizontalSpaceOnMobile = false,
   alwaysWaitWhileLoading = false,
+  emptyMessage = undefined,
 }) => {
   const { isMobile } = useScreenSize()
 
-  if (!isLoading && !rows?.length) {
+  const isEmpty = !isLoading && !rows?.length
+
+  if (isEmpty && !emptyMessage) {
```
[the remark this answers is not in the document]
csillag: Simplified

src/app/pages/ConsensusTransactionsPage/index.tsx
```diff
@@ -69,7 +76,9 @@ export const ConsensusTransactionsPage: FC = () => {
   const { isLoading, isFetched, data } = transactionsQuery
 
   const transactions = data?.data.transactions
-  setBeforeDateFromCollection(data?.data.transactions[0].timestamp)
+  if (transactions?.length) {
```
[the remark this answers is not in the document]
csillag: That might have been the original intention, but it was just throwing an exception when `data` was defined, but the `transactions` list was empty. Now I have restored the original intention.

src/app/components/Select/index.tsx
```diff
@@ -25,56 +25,81 @@ import { WithOptionalOwnerState } from '@mui/base/utils'
...
+    overflow: 'auto',
+    outline: 0,
+    background: isLight ? COLORS.white : theme.palette.tertiary.main,
+    filter: 'drop-shadow(0px 4px 4px rgba(0, 0, 0, 0.25))',
+    maxHeight: '50vh',
+  }
+})
+
+export const StyledSelectOption = styled(Option)(({ theme, className }) => {
+  const isLight = className?.split(' ').includes(LIGHT_MARKER_CLASS)
```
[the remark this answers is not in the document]
csillag: > Bool prop -> class name -> parse to bool var -> conditional css string is awful.

Yes is is. (Was.)

Now I have moved to `Bool prop -> bool prop -> conditional CSS`

Not sure it's much better, though.

PR #1687, opened 2025-01-14
Tune displaying account names

csillag:


.changelog/1687.trivial.md
```diff
@@ -0,0 +1 @@
+Fine-tune whan to display account names
```
buberdds, line 1: typo

src/app/components/Account/RuntimeAccountDetailsView.tsx
```diff
@@ -76,12 +76,7 @@ export const RuntimeAccountDetailsView: FC<RuntimeAccountDetailsViewProps> = ({
         <AccountAvatar account={account} />
       </StyledListTitleWithAvatar>
       <dd>
-        <AccountLink
-          showAddressAsName
```
buberdds, line 80: @donouwens 
Just to confirm we want to show duplicated names in contracts? 

Screenshot from 2025-01-15 10-42-20
Screenshot from 2025-01-15 10-42-07


instead of 

Screenshot from 2025-01-15 10-45-43

.changelog/1687.trivial.md
csillag: Fixed typo

src/app/components/Account/RuntimeAccountDetailsView.tsx
csillag: For context, here is the link: https://pr-1687.oasis-explorer.pages.dev/mainnet/sapphire/address/0x6665a6Cae3F52959f0f653E3D04270D54e6f13d8

csillag: Here is another link to another address page, which is not a token: https://pr-1687.oasis-explorer.pages.dev/testnet/pontusxtest/address/0x938224aC9e4832C517818B147BB1f6b10ADdCd26

csillag: Maybe we should hide the name selectively just for token contracts?

csillag: Updated behavior so that in account details, we don't show the account name if we also have a token name, but show it otherwise.

src/app/components/Account/RuntimeAccountDetailsView.tsx
```diff
@@ -77,7 +77,7 @@ export const RuntimeAccountDetailsView: FC<RuntimeAccountDetailsViewProps> = ({
       </StyledListTitleWithAvatar>
       <dd>
         <AccountLink
-          showAddressAsName
+          showOnlyAddress={!!token?.name}
```
buberdds, line 80: Isn't `!!contract` more appropriate here?

csillag: What is the rationale for hiding the name for all contracts?

(Please note that we can also have other named contracts, not just token contracts. Tokens contracts are already handled.)

buberdds: ah we show other row with name only for `token` so I think we should use the same condition in both places.

csillag: The name field is mandatory for `ERC-20` tokens, but not for other tokens.
That is why the `name`  field (of the token data structure) is defined like this:
```
  /** Name of the token, as provided by token contract's `name()` method. */
  name?: string;
```

So it's possible that we have tokens without names. So displaying a token link doesn't necessarily mean that we have a valid name there. That's why I used `token.name` for the condition.

This way, if the token doesn't have a name, but the smart contract does, we can still display it.

buberdds (approved):

PR #1688, opened 2025-01-14
Add support for merging all networks in the layer selector 

csillag:
This PR adds support for merging all networks in the layer selector,
so that instead of a hierarchical menu, we get a flat one.

On the Pontus-X instance, using this feature, instead of this:

image

and this:

image

We can have this:

image

Please note that you only want to activate this if you don't have
layers that are named the same on multiple networks.

(If you do, you can override names using the spacialScopeNames
config option in config.ts.)

   * * *
   
Merging this PR doesn't change behavior, unless to turn the `mergeNetworksInLayerSelector`
flag from false to true.

### DIFF
--- a/src/app/components/LayerPicker/LayerMenu.tsx
+++ b/src/app/components/LayerPicker/LayerMenu.tsx
@@ -14,10 +14,11 @@ import { Network } from '../../../types/network'
 import { orderByLayer } from '../../../types/layers'
 import { useScreenSize } from '../../hooks/useScreensize'
 import { useScopeParam } from '../../hooks/useScopeParam'
+import { SearchScope } from '../../../types/searchScope'
 
 type BaseLayerMenuItemProps = {
   divider: boolean
-  layer: Layer
+  targetScope: SearchScope
 }
 
 const LayerMenuItemCaption: FC<PropsWithChildren> = ({ children }) => (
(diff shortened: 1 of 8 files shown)


csillag:
This can be seen in action at https://pr-1355.oasis-explorer.pages.dev/pontusx/dev

.changelog/1866.trivial.md
```diff
@@ -0,0 +1 @@
+Add support for merging all networks in the layer selector
```
lukaw3d, line 1: filename typo

csillag: Fixed

lukaw3d (approved):

src/app/components/LayerPicker/LayerMenu.tsx
```diff
@@ -51,78 +52,81 @@ export const DisabledLayerMenuItem: FC<BaseLayerMenuItemProps> = ({ divider, lay
...
-}) => {
-  const currentScope = useScopeParam()
-  const [hoveredLayer, setHoveredLayer] = useState<undefined | Layer>()
-  const options = Object.values(Layer)
+type LayerMenuOption = {
+  scope: SearchScope
+  enabled: boolean
+}
+
+const getOptionsForNetwork = (network: Network, activeScope?: SearchScope | undefined): LayerMenuOption[] =>
```
lukaw3d, line 107: `activeScope?: SearchScope | undefined`
this parameter is never provided?

csillag: Oops, that was accidentally dropped. But we need is so I added it back now.. Thanks, good catch.

Issue #1703, opened 2025-01-17
[Feature]: Display all signers for runtime transactions


csillag:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### SUMMARY

Runtime transactions can have multiple signers. We should display them all.

### DESCRIPTION

Currently, Nexus's API only gives us the first signer, but that is going to change:
- [ ] https://github.com/oasisprotocol/nexus/issues/884

When the missing data becomes available, Explorer should be able to show multiple senders in the "From:" section of Runtime TX details.

Example TX with multiple signers: `9b6f77229709326983fa513d33aa0972a7543197118f4454693fe13541fdf26b`

PR #1705, opened 2025-01-18
Display multiple signers for runtime transactions

csillag:
Sometimes a TX has multiple signers. In that case, we should show them all in the "from" field in the TX details, like this:

Solves:
- #1703

image

Live sample [here](https://pr-1705.oasis-explorer.pages.dev/testnet/sapphire/tx/9b6f77229709326983fa513d33aa0972a7543197118f4454693fe13541fdf26b).

Depends on:
- [x] https://github.com/oasisprotocol/nexus/pull/885

### DIFF
--- a/src/app/components/Account/ContractCreatorInfo.tsx
+++ b/src/app/components/Account/ContractCreatorInfo.tsx
@@ -25,7 +25,7 @@ const TxSender: FC<{ scope: SearchScope; txHash: string; alwaysTrim?: boolean }>
   }
   const query = useGetRuntimeTransactionsTxHash(scope.network, scope.layer, txHash)
   const tx = query.data?.data.transactions[0]
-  const senderAddress = tx?.sender_0_eth || tx?.sender_0
+  const senderAddress = tx?.signers[0].address_eth ?? tx?.signers[0].address
 
   return query.isLoading ? (
     <Skeleton
--- a/.changelog/1705.feature.md
+++ b/.changelog/1705.feature.md
@@ -1,0 +1,1 @@
+Add support for displaying multiple signers of runtime transactions
--- a/src/app/components/Transactions/RuntimeTransactions.tsx
+++ b/src/app/components/Transactions/RuntimeTransactions.tsx
@@ -126,10 +126,11 @@ export const RuntimeTransactions: FC<TransactionsProps> = ({
                     <AccountLink
                       labelOnly={
                         !!ownAddress &&
-                        (transaction.sender_0_eth === ownAddress || transaction.sender_0 === ownAddress)
+                        (transaction.signers[0].address_eth === ownAddress ||
+                          transaction.signers[0].address === ownAddress)
                       }
                       scope={transaction}
-                      address={transaction.sender_0_eth || transaction.sender_0}
+                      address={transaction.signers[0].address_eth ?? transaction.signers[0].address}
                       alwaysTrim
                     />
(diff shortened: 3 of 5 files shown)


src/app/components/Account/ContractCreatorInfo.tsx
```diff
@@ -25,7 +25,8 @@ const TxSender: FC<{ scope: SearchScope; txHash: string; alwaysTrim?: boolean }>
   }
   const query = useGetRuntimeTransactionsTxHash(scope.network, scope.layer, txHash)
   const tx = query.data?.data.transactions[0]
-  const senderAddress = tx?.sender_0_eth || tx?.sender_0
+  const senderAddress =
+    tx?.signers?.[0].address_eth ?? tx?.sender_0_eth ?? tx?.signers?.[0].address ?? tx?.sender_0
```
lukaw3d, line 29: Is this a necessary change? If it is then it should be fixed on the API - it is too awful to use like this

.changelog/1705.trivial.md
```diff
@@ -0,0 +1 @@
+Add support for displaying multiple signers of runtime transactions
```
lukaw3d, line 1: I would mark this as a feature, not trivial

src/app/components/Account/ContractCreatorInfo.tsx
lukaw3d: I would assume `sender_0 === signers[0]` and add a runtime check inside useGetRuntimeTransactionsTxHash:
`if (!!signers[0] && signers[0].address !== sender_0) throw new Error('Send this URL to nexus people.')`

and then I would stick to almost always using sender_0. Or always pretend sender_0 field doesn't exist (is signers field ever missing?).

csillag: >  it should be fixed on the API

It _has been_ fixed in the API.

The `signers` array is the new, more correct solution. It has been introduced recently, deployed this morning. It will always exist, from now on. The old `sender_0` field has now been marked as deprecated, and will eventually disappear.

We only need to look at both for a short, transitional period. In a few weeks, when we can be sure that all deployed Nexus instances have v0.5.3 or newer, we can completely remove the fallback to the old, deprecated field.

.changelog/1705.trivial.md
csillag: OK, renamed as feature.

src/app/components/Account/ContractCreatorInfo.tsx
csillag: >  The old `sender_0` field has now been marked as deprecated, and will eventually disappear.

Btw the `deprecated` flag is tunneled all the way via Orval, so we get this in the IDE:

image

So, again, we can get rid of the deprecated fields soon-ish, simplifying this significantly.

lukaw3d: Ah, good

(if it will take a few months: could move `signers = signers ?? [{address_eth: sender_0_eth, address: sender}]` into useGetRuntimeTransactionsTxHash)

lukaw3d (approved):

src/app/components/Account/ContractCreatorInfo.tsx
csillag: Good idea. Moved all the legacy-data transformation to the API layer, with appropriate TODO comment.

PR #1706, opened 2025-01-18
Display link to proposer in consensus block data

csillag:
When displaying consensus block details, we should also include a link to the validator that has proposed the block.
Other explorers do that, too.

image

Please note that this data has been already been displayed in the latest block list, like this:

image

It was only missing from the individual block detail pages, because it was not served by the backend, but that limitation is now gone, so we can display it.

Depends on:
- [x] Nexus: https://github.com/oasisprotocol/nexus/pull/888

### DIFF
--- a/.changelog/1706.trivial.md
+++ b/.changelog/1706.trivial.md
@@ -1,0 +1,1 @@
+Show proposer link in consensus block details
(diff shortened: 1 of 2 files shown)


csillag:
Design review from @donouwens:

> This one looks good to me, great shout. Thanks!

lubej (approved):

PR #1727, opened 2025-02-04
Use custom query key for validators map

buberdds:


csillag (approved):
Reviewed. LGTM.

lukaw3d:
Explanation: useGetConsensusValidatorsAddressNameMap and useGetConsensusValidators both use the underlaying `generated.useGetConsensusValidators`, and transform the response in different ways. Issue is, both were using the same default queryKey.

PR #1778, opened 2025-02-28
Account details: add link to validator details - salvage extra features

csillag:
Pr #1466 is now obsolete, it has been implemented using another way. However, there are some features buried in that PR that might be worth salvaging. I will do that here.


csillag:
obsolete

PR #1780, opened 2025-03-02
Preserve tx_method query arg on block prev/next button

### LINKED ISSUE #1743: [Bug]: Filter by type working incorrectly when switching layer and/or network

- [x] I have searched the existing issues

### SUMMARY

_No response_

### STEPS TO REPRODUCE

- https://explorer.oasis.io/mainnet/sapphire
- Latest Transactions -> Filter by type -> select any value
- click layer picker at the top -> select testnet Sapphire or Emerald
- filter is showing value, but it is not used in API query

### EXPECTED BEHAVIOR

_No response_

### LINKED ISSUE #1779: [Bug]: Filter forgotten when walking through blocks

- [x] I have searched the existing issues

### SUMMARY

When clicking next/prev buttons, filter is not considered

### STEPS TO REPRODUCE

1. Open https://explorer.oasis.io/mainnet/consensus/block/23508465
2. Change Filter by type (e.g. Burn)
3. Click previous or next block arrow
4. Filter is still "Burn", but transactions weren't filtered actually.

### EXPECTED BEHAVIOR

Filter should remain active when you move to the next/previous block.

The same behavior should be fixed for ParaTime blocks, e.g. https://explorer.oasis.io/mainnet/sapphire/block/7821641

csillag:
Make it so that we preserve the tx method filtering setting when stepping between blocks.

Also fix some smaller issues and do some cleanup under the hood.

Fixes #1743 
Fixes #1779 

(I suggest review commit by commit.)

### DIFF
--- a/src/app/components/Select/index.tsx
+++ b/src/app/components/Select/index.tsx
@@ -183,7 +183,15 @@ const SelectOption = <T extends SelectOptionBase>({
 interface SelectCmpProps<T extends SelectOptionBase> {
   label?: ReactNode
   options: T[]
+  /**
+   * If you want to use this component in uncontrolled mode, you can provide the detauls / initial value here.
+   */
   defaultValue?: T['value']
+
+  /**
+   * You can provide a value if you want to use this component in controlled mode
+   */
+  value?: T['value'] | null
   handleChange?: (selectedOption: T['value'] | null) => void
   placement?: PopupProps['placement']
   className?: string
@@ -198,6 +206,7 @@ const SelectCmp = <T extends SelectOptionBase>({
   label,
   options,
   defaultValue,
+  value,
   handleChange,
   placement = 'bottom-start',
   className,
@@ -224,6 +233,7 @@ const SelectCmp = <T extends SelectOptionBase>({
       <CustomSelect<T['value']>
         id={selectId}
         defaultValue={defaultValue}
+        {...(value === undefined || value === null ? {} : { value })}
         onChange={onChange}
         root={root}
         listbox={listbox}
--- a/src/app/pages/ConsensusTransactionsPage/index.tsx
+++ b/src/app/pages/ConsensusTransactionsPage/index.tsx
@@ -16,12 +16,9 @@ import { useRequiredScopeParam } from '../../hooks/useScopeParam'
 import { VerticalList } from '../../components/VerticalList'
 import { ConsensusTransactionDetailView } from '../ConsensusTransactionDetailPage'
 import { useConsensusListBeforeDate } from '../../hooks/useListBeforeDate'
-import { useTypedSearchParam } from '../../hooks/useTypedSearchParam'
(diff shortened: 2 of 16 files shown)


matevz (approved):
Works as expected, thanks.

src/app/pages/ConsensusTransactionsPage/index.tsx
```diff
@@ -16,22 +16,17 @@ import { useRequiredScopeParam } from '../../hooks/useScopeParam'
...
+import { getConsensusTransactionMethodFilteringParam } from '../../components/ConsensusTransactionMethod'
 import Box from '@mui/material/Box'
 
 export const ConsensusTransactionsPage: FC = () => {
   const [tableView, setTableView] = useState<TableLayout>(TableLayout.Horizontal)
   const { t } = useTranslation()
   const { isMobile } = useScreenSize()
   const pagination = useSearchParamsPagination('page')
-  const [method, setMethod] = useTypedSearchParam<ConsensusTxMethodFilterOption>('method', 'any', {
-    deleteParams: ['date'],
```
lukaw3d, line 33: This fixes:
- https://explorer.dev.oasis.io/mainnet/consensus/tx?page=3
- set Filter by type
- stayed on page 3

After PR:
- resets to page 1

src/app/utils/route-utils.ts
```diff
@@ -130,8 +141,13 @@ export abstract class RouteUtils {
     return `${this.getScopeRoute(scope)}/block`
   }
 
-  static getBlockRoute = (scope: SearchScope, blockHeight: number) => {
-    return `${this.getScopeRoute(scope)}/block/${encodeURIComponent(blockHeight)}`
+  static getBlockRoute = (
+    scope: SearchScope,
+    blockHeight: number,
+    searchParams?: URLSearchParams,
+    preserveParams: string[] = [],
```
lukaw3d, line 148: I hate it

lukaw3d (approved):

src/app/components/Select/index.tsx
```diff
@@ -224,6 +233,7 @@ const SelectCmp = <T extends SelectOptionBase>({
       <CustomSelect<T['value']>
         id={selectId}
         defaultValue={defaultValue}
+        {...(value === undefined || value === null ? {} : { value })}
```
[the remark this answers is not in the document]
csillag: As far as can see from my experiments, in controlled mode, there is just now way to set the value from a defined value to undefined. I guess we just need to use a default/placeholder value.

src/app/utils/route-utils.ts
csillag: Would you prefer a more generally useful `options` object, for which 'preserveParams` would be just one possilble key?

lukaw3d: no. Ideally I wouldn't preserve params. But if we want to, then getBlockRoute should be the function that decides which params to preserve

PR #1783, opened 2025-03-04
Fix consensus list view fees

lubej:


lubej:
Example here:

`/mainnet/consensus/address/oasis1qq70q4d84xpl36fd9esxl6glwtkf92hmvqudjmv0`

PR #1785, opened 2025-03-04
Use new API fields to detect empty accounts in Consensus search

lukaw3d:


src/oasis-nexus/api.ts
```diff
@@ -145,20 +145,26 @@ declare module './generated/api' {
...
     return (
       available === '0' &&
       debonding_delegations_balance === '0' &&
       delegations_balance === '0' &&
       escrow === '0' &&
       total === '0' &&
       nonce === 0 &&
-      size === 'XXS'
+      stats.num_txns === 0 &&
+      !first_activity
```
csillag, line 166: I am not exactly clear of the definition of "activity", but doesn't this last check also cover everything else, making all other conditions redundant?

The API docs says

> The second-granular consensus time of the block in which this account was first active. Dates before Cobalt (2021-04-28) are approximate.

Are there cases when `first_activity` is missing, but there is still some activity we can detect from other fields? (I'll try to clarify this with backend.)

csillag, line 166: I am not exactly clear of the definition of "activity", but doesn't this last check also cover everything else, making all other conditions redundant?

The API docs says

> The second-granular consensus time of the block in which this account was first active. Dates before Cobalt (2021-04-28) are approximate.

Are there cases when `first_activity` is missing, but there is still some activity we can detect from other fields? (I'll try to clarify this with backend.)

csillag (approved):
Reviewed, LGTM.

csillag:
This needs a rebase.

src/oasis-nexus/api.ts
```diff
@@ -145,21 +145,26 @@ declare module './generated/api' {
...
-    //   available === '0' &&
-    //   debonding_delegations_balance === '0' &&
-    //   delegations_balance === '0' &&
-    //   escrow === '0' &&
-    //   total === '0' &&
-    //   nonce === 0 &&
-    //   size === 'XXS'
-    // )
-    // TODO: we should also check the number of transactions, where it becomes available
+    const {
```
lukaw3d, line 148: Issue? Can't find the designated burn address anymore
https://explorer.dev.oasis.io/search?q=oasis1qzq8u7xs328puu2jy524w3fygzs63rv3u5967970
https://pr-1785.oasis-explorer.pages.dev/search?q=oasis1qzq8u7xs328puu2jy524w3fygzs63rv3u5967970

| Before | After |
| --- | --- |
|  explorer dev oasis io_search_q=oasis1qzq8u7xs328puu2jy524w3fygzs63rv3u5967970      |      pr-1785 oasis-explorer pages dev_search_q=oasis1qzq8u7xs328puu2jy524w3fygzs63rv3u5967970  |

src/oasis-nexus/api.ts
lukaw3d: It might cover everything but I'm not sure

src/oasis-nexus/api.ts
csillag: This  might be a special case for Nexus. We could add a workaround on either backend or frontend...
(Backend preferable)

lukaw3d: Frontend fix but I would rather abandon and ignore: https://github.com/oasisprotocol/explorer/pull/1812

PR #1793, opened 2025-03-08
Generalize TableConfig as LocalSettings

csillag:
We might want to add other local options which are stored on LocalStorage, and changed based on used input.

The infrastructure added for `TableConfig` is perfect for all similar settings, just needs a more fitting name.

This provides the base for:
- [ ] https://github.com/oasisprotocol/explorer/pull/1794


lubej:
I am not opposed to renaming TableConfig. But `LocalConfig` naming is ambiguous to me at least. I would prefer it to be `UserConfig`, `UserSettings` or `LocalSettings`. Config to me kind of implies, that the config could be coming from somewhere else, than the application itself. Whereas Settings, to me implies it is coming from the application itself. Maybe it is just me. @buberdds or @lukaw3d what do you think?

src/app/providers/LocalConfigContext.ts
```diff
@@ -0,0 +1,16 @@
+import { createContext } from 'react'
+import { TableAgeType } from '../../types/table-age-type'
+
+/**
+ * This is where we store local state
+ */
+export interface LocalConfigProviderState {
+  ageHeaderType: TableAgeType
```
lubej, line 8: Also here, if we would start adding more options. I would prefer it would be a single local storage key and it being stored as a JSON of options( `{[option-key]: option-value }`), opposed to creating new local storage for each of the option.

csillag:
> I would prefer it to be `UserConfig`, `UserSettings` or `LocalSettings`. 

Renamed to `LocalSettings`, as suggested.

src/app/providers/LocalConfigContext.ts
csillag: >  I would prefer it would be a single local storage key and it being stored as a JSON of options

Reorganized accordingly

src/app/providers/LocalSettingsProvider.tsx
```diff
@@ -0,0 +1,46 @@
...
+    }
+  }, [state])
+
+  const setAgeHeaderType = (ageHeaderType: TableAgeType) => {
+    setState(prevState => ({
+      ...prevState,
+      ageHeaderType,
+      dirty: true,
+    }))
+  }
```
lubej, line 38: I would refactor this as well to `setLocalSetting(key: keyof LocalSettings, value)`, not sure if type of value can be inferred without generic type here. And probably remove the dirty flag(and above `useEffect`), as it calls `setState` twice, I would prefer if we would avoid unnecessary re-renders.

csillag: Updated accordingly. (No more `dirty` and `useEffect()`, also there is a `setLocalSetting()` function to be called by the individual wrappers, which we still need for type safefy.

csillag: With the current setup, care must be taken when adding new wrapper functions to use the correct types (matching with those defined in the type `LocalSetting`.)

csillag: It's all fine now, no more type safety issues. (Using `Partial<LocalSettings>` instead of `keyof LocalSettings`)

lubej: I meant something like this:
```typescript
  const setValues = <T extends keyof LocalSettings>(key: T, value: LocalSettings[T]) => {
    setState(prevState => ({
      ...prevState,
      ...{ [key]: value },
    }))

    localStorage.set(StorageKeys.LocalSettings, {
      ...state,
      ...{ [key]: value },
    })
  }

  const setAgeHeaderType = (ageHeaderType: TableAgeType) => setValues('ageHeaderType', ageHeaderType)
```
And just expose `setValue`(probably rename it), don't see a reason why we would need to define "wrapper" function for every key:value pair we add to LocalSettings map.

lubej (approved):
I would replace `setValues`, otherwise looks fine by me.

src/app/providers/LocalSettingsProvider.tsx
csillag: Yeah, refactored accordingly.

src/app/providers/LocalSettingsContext.ts
```diff
@@ -0,0 +1,11 @@
+import { createContext } from 'react'
+import { LocalSettings } from '../../types/local-settings'
+
+export interface LocalSettingsProviderContext {
+  readonly settings: LocalSettings
+  changeSetting: (key: keyof LocalSettings, value: any) => void
```
lukaw3d, line 6: No typesafety :scream:

csillag: Thats a copy-paste error, what we wanted (and really have in the implementation) is this: `<T extends keyof LocalSettings>(key: T, value: LocalSettings[T]) `

Will fix in the next related PR.

csillag: Fixed in https://github.com/oasisprotocol/explorer/pull/1794/commits/0748a432f74967d5810695c9e84cc5dca0c92cd3

PR #1794, opened 2025-03-08
When auto-redirecting from landing page, preserve scope 

csillag:
There are some network / layer configurations that (if configured in the `.env`) result in not showing the graph on the landing page, but instead immediately redirecting to the dashboard of a Runtime.

This change makes it so that if we change to a layer using the layer selector, that layer is preserved, and next time we redirect from the landing page, we go to this layer.

This was a feature request for the Pontus-X browser.

This implementation depends on:
- [x] #1793. 

(Hint for testing: uncomment `REACT_APP_FIXED_LAYER` in .env, or see this live in action on the pontus-x branch: https://pr-1355.oasis-explorer.pages.dev/)

### DIFF
--- a/.changelog/1794.trivial.md
+++ b/.changelog/1794.trivial.md
@@ -1,0 +1,1 @@
+When auto-redirecting from landing page, restore stored scope
--- a/src/types/local-settings.ts
+++ b/src/types/local-settings.ts
@@ -1,12 +1,15 @@
 import { TableAgeType } from './table-age-type'
+import { SearchScope } from './searchScope'
 
 /**
  * This is where we store local state
  */
 export interface LocalSettings {
   ageHeaderType: TableAgeType
+  preferredScope: SearchScope | undefined
 }
 
 export const defaultLocalSettings: LocalSettings = {
   ageHeaderType: TableAgeType.Distance,
+  preferredScope: undefined,
 }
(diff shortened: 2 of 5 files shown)


.changelog/1794.trivial.md
```diff
@@ -0,0 +1 @@
+When auto-redirecting from landing page, preserve scope
```
lukaw3d, line 1: I would change "preserve" to "store"
("preserve" to me implies it isn't removed during a redirect. "store" implies longer time storage)

src/types/local-settings.ts
```diff
@@ -1,12 +1,15 @@
 import { TableAgeType } from './table-age-type'
+import { SearchScope } from './searchScope'
 
 /**
  * This is where we store local state
  */
 export interface LocalSettings {
   ageHeaderType: TableAgeType
+  preferredScope: SearchScope | undefined
```
lukaw3d, line 9: Does pontus-x team really want this to be based on network selector clicks?

I would rename it to lastViewedScope and store it based on viewing history instead :shrug:

lukaw3d (approved):

src/types/local-settings.ts
csillag: Yes, according to their feedback, thie current behavior is exactly what they want.

.changelog/1794.trivial.md
csillag: Updated to:

> When auto-redirecting from landing page, restore stored scope

PR #1795, opened 2025-03-08
Enable Oasis account name lookup on pontus-x runtimes

csillag:
On Pontus-X runtimes, we need to look for named accounts in two sources:
 - The named address repository at Oasis Nexus
 - The named address repository at deltaDAO

This commit re-enables the first source, which was somehow lost along the way. (The second source was always working properly. that's why we didn't notice earlier.)

### Before:

image

### After:

image

### DIFF
--- a/src/app/data/oasis-account-names.ts
+++ b/src/app/data/oasis-account-names.ts
@@ -34,6 +34,10 @@ const dataSources: Record<Network, Partial<Record<Layer, string>>> = {
       'https://raw.githubusercontent.com/oasisprotocol/nexus/main/named-addresses/testnet_emerald.json',
     [Layer.sapphire]:
       'https://raw.githubusercontent.com/oasisprotocol/nexus/main/named-addresses/testnet_sapphire.json',
+    [Layer.pontusxdev]:
+      'https://raw.githubusercontent.com/oasisprotocol/nexus/main/named-addresses/testnet_pontusxdev.json',
+    [Layer.pontusxtest]:
+      'https://raw.githubusercontent.com/oasisprotocol/nexus/main/named-addresses/testnet_pontusxtest.json',
   },
   [Network.localnet]: {
     [Layer.consensus]: undefined,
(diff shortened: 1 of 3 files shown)


src/app/hooks/useAccountMetadata.ts
```diff
@@ -23,10 +23,13 @@ export const useAccountMetadata = (scope: SearchScope, address: string): Account
...
   })
   const oasisData = useOasisAccountMetadata(scope.network, scope.layer, getOasisAddress(address), {
-    enabled: !isPontusX && !isLocalnet(scope.network),
+    enabled: !isLocalnet(scope.network),
     useErrorBoundary: false,
   })
-  const registryData = isPontusX ? pontusXData : oasisData
+  if (address === 'oasis1qp3r8hgsnphajmfzfuaa8fhjag7e0yt35cjxq0u4') {
+    console.log('oasis data', oasisData)
+  }
```
lukaw3d, line 31: TODO

src/app/data/oasis-account-names.ts
```diff
@@ -34,6 +34,10 @@ const dataSources: Record<Network, Partial<Record<Layer, string>>> = {
       'https://raw.githubusercontent.com/oasisprotocol/nexus/main/named-addresses/testnet_emerald.json',
     [Layer.sapphire]:
       'https://raw.githubusercontent.com/oasisprotocol/nexus/main/named-addresses/testnet_sapphire.json',
+    [Layer.pontusxdev]:
+      'https://raw.githubusercontent.com/oasisprotocol/nexus/main/named-addresses/testnet_pontusxdev.json',
```
lukaw3d, line 38: (can we just copy in pontusx's other account names too, to simplify our code?)

lukaw3d (approved):

src/app/hooks/useAccountMetadata.ts
csillag: Removed. thanks.

src/app/data/oasis-account-names.ts
csillag: Their list changes much more often than hours.

Plus, it's easy to keep the source info correctly if the sources are separate... but maybe we can rethink this when we make this a proper nexus API.

PR #1796, opened 2025-03-08
Clean up name attribution by Oasis/deltaDAO

csillag:
Some account names come from Oasis, some from deltaDAO.
This commit makes the difference clear in the tooltips.

Built on (depends on):
- [ ] #1795 

### Tooltip for Oasis identities:

image

### Tooltip for pontus-x (deltaDAO) identities:

image

### Tooltip for self-acclaimed names (not changed):

image

### DIFF
--- a/.changelog/1796.trivial.md
+++ b/.changelog/1796.trivial.md
@@ -1,0 +1,1 @@
+Clean up name attribution by Oasis/deltaDAO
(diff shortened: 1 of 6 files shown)


csillag:
Desing feedback from @donouwens 

>  both the assigned by Oasis/deltaDAO have more authority and therefore more trustworthy. So maybe keep those both green and the self-proclaimed gray

lukaw3d (approved):

PR #1777, opened 2025-02-28
ROFL apps support

buberdds:
PR contains minor fixes needed for a release.

Most commits were extracted to:
- status badge https://github.com/oasisprotocol/explorer/pull/1819
- new ROFL instance event  https://github.com/oasisprotocol/explorer/pull/1820
- ROFL link component https://github.com/oasisprotocol/explorer/pull/1827
- Tweak TableHeaderAge component  https://github.com/oasisprotocol/explorer/pull/1829
- ROFL apps list page https://github.com/oasisprotocol/explorer/pull/1834
- ROFL app instances card https://github.com/oasisprotocol/explorer/pull/1837
- ROFL app metadata and policy cards https://github.com/oasisprotocol/explorer/pull/1840
- ROFL app details card https://github.com/oasisprotocol/explorer/pull/1841
- Provide missing data to ROFL apps list https://github.com/oasisprotocol/explorer/pull/1842
- Enable ROFL app details page https://github.com/oasisprotocol/explorer/pull/1848
- Setup ROFL app transactions card https://github.com/oasisprotocol/explorer/pull/1851
- Add ROFL to home page graph https://github.com/oasisprotocol/explorer/pull/1853
- Include ROFL apps in search results https://github.com/oasisprotocol/explorer/pull/1859
- ROFL apps minor UI tweaks https://github.com/oasisprotocol/explorer/pull/1861
- ROFL app instance transactions tab https://github.com/oasisprotocol/explorer/pull/1863

PR #1820, opened 2025-03-17
Handle new ROFL instance event

buberdds:
Extracted from https://github.com/oasisprotocol/explorer/pull/1777 and https://github.com/oasisprotocol/explorer/pull/1781

We are not handling new ROFL event (dev crash / prod warning)

### DIFF
--- a/src/locales/en/translation.json
+++ b/src/locales/en/translation.json
@@ -531,6 +531,7 @@
     "roflAppCreated": "ROFL App Created",
     "roflAppUpdated": "ROFL App Updated",
     "roflAppRemoved": "ROFL App Removed",
+    "instanceRegistered": "ROFL Instance Registered",
     "evmLog": "EVM log message",
     "filter": {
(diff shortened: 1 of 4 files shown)


lubej (approved):

src/locales/en/translation.json
```diff
@@ -656,6 +657,7 @@
   "rofl": {
     "active": "Active",
     "inactive": "Inactive",
+    "rak": "RAK",
```
lukaw3d, line 660: Can this be "Runtime Attestation Key", or do we need to save space? (and is this even Runtime Attestation Key)

buberdds: It's "RAK" because we are using this abbreviation in ROFL app detail page (but in table context, where space matter). Anyway wanted to use the same label across different places. Do we have place to show full name in events details - yes. 

I am waiting for backend feedback too.

PR #1828, opened 2025-03-19
Use checksum capitalization for Emitting contract

lukaw3d:


src/app/components/RuntimeEvents/RuntimeEventDetails.tsx
```diff
@@ -184,7 +184,7 @@ const RuntimeEventDetailsInner: FC<{
       )
     case RuntimeEventType.evmlog: {
       const { parsedEvmLogName } = parseEvmEvent(event)
-      const emittingEthAddress = `0x${Buffer.from(event.body.address, 'base64').toString('hex')}`
+      const emittingEthAddress = getEthAccountAddressFromBase64(event.body.address)
```
buberdds, line 187: missing import 
shouldn't this be normalized in API layer?

buberdds (approved):

src/app/components/RuntimeEvents/RuntimeEventDetails.tsx
lukaw3d: Ah, yeah, it doesn't depend on other info than event.type. Will move

PR #1829, opened 2025-03-19
Tweak TableHeaderAge component

buberdds:

- need custom label in header 
- currently header is misaligned compared to content cell due to Button width and padding

### DIFF
--- a/src/app/components/TableHeaderAge/index.tsx
+++ b/src/app/components/TableHeaderAge/index.tsx
@@ -8,7 +8,11 @@ import { useLocalSettings } from '../../hooks/useLocalSettings'
 import { tooltipDelay } from '../../../styles/theme'
 import { getTimeZone } from '../../hooks/useFormattedTimestamp'
 
-export const TableHeaderAge: FC = () => {
+type TableHeaderAgeProps = {
+  label?: string
+}
+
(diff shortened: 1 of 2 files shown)


src/app/components/TableHeaderAge/index.tsx
```diff
@@ -47,15 +58,15 @@ export const TableHeaderAge: FC = () => {
...
         >
-          <Button variant="text" onClick={() => changeSetting('ageHeaderType', TableAgeType.DateTime)}>
+          <StyledButton variant="text" onClick={() => changeSetting('ageHeaderType', TableAgeType.DateTime)}>
             <Typography
               sx={{
                 fontWeight: 700,
               }}
             >
-              {t('common.age')}
+              {label || t('common.age')}
```
lubej, line 67: Not sure why only Distance would need custom label? In this case it would make sense to have a Map <TableAgeType, title>? Or am I missing something?

lubej (approved):

src/app/components/TableHeaderAge/index.tsx
buberdds: For now it is all I need, if this part will need another adjustment I will open PR in the future.

PR #1830, opened 2025-03-21
Prevent empty span from rendering in AccountLink on mobile

kaja-osojnik:
Fixes extra white space on mobile by removing an unnecessary empty <span> in the "From" address field of the Transaction table.

Before:
Screenshot 2025-03-21 at 12 45 10

After:
Screenshot 2025-03-21 at 12 49 38

### DIFF
--- a/src/app/components/Account/AccountLink.tsx
+++ b/src/app/components/Account/AccountLink.tsx
@@ -182,14 +182,16 @@ export const AccountLink: FC<Props> = ({
   return (
     <WithTypographyAndLink scope={scope} address={address} mobile labelOnly={labelOnly}>
       <>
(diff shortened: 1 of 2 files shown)


src/app/components/Account/AccountLink.tsx
```diff
@@ -182,14 +182,22 @@ export const AccountLink: FC<Props> = ({
...
+              pattern={highlightedPartOfName}
+              extraTooltip={tooltipTitle}
+            />
+          </Box>
+        ) : (
           <AdaptiveHighlightedText
             text={showAccountName ? accountName : ''}
             pattern={highlightedPartOfName}
             extraTooltip={tooltipTitle}
           />
```
lukaw3d, line 199: This isn't needed. Can only ever show `text={''}`

PR #1848, opened 2025-03-28
Enable ROFL app details page

buberdds:

### DIFF
--- a/src/app/pages/RoflAppDetailsPage/index.tsx
+++ b/src/app/pages/RoflAppDetailsPage/index.tsx
@@ -29,6 +29,8 @@ import { Secrets } from './Secrets'
 import { RouterTabs } from '../../components/RouterTabs'
 import { instancesContainerId } from '../../utils/tabAnchors'
 import { RoflAppDetailsContext } from '../RoflAppDetailsPage/hooks'
+import { MetaDataCard } from './MetaDataCard'
+import { PolicyCard } from './PolicyCard'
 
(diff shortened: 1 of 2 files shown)


src/app/pages/RoflAppDetailsPage/index.tsx
```diff
@@ -1,13 +1,68 @@
...
+  const { isFetched, isLoading, data } = useGetRuntimeRoflAppsId(scope.network, Layer.sapphire, id)
+  const roflApp = data?.data
+
+  if (!roflApp && isFetched) {
+    throw AppErrors.NotFoundRoflApp
+  }
+
+  return (
+    <PageLayout>
+      <SubPageCard featured title={roflApp?.metadata['net.oasis.rofl.name'] || t('rofl.header')}>
```
buberdds, line 40: show it only when isFetched is true to avoid switching label

src/app/utils/route-utils.ts
```diff
@@ -295,6 +295,22 @@ const validateConsensusAddressParam = (address: string) => {
   return isValid
 }
 
+const validateRoflAppIdParam = (id: string) => {
+  // Oasis client SDK does not provide a way to validate ROFL app id
+  const isValid = id.startsWith('rofl1')
```
lukaw3d, line 300: Pretty sure it's still bech32, so `oasis.address.fromBech32('rofl', `

lukaw3d (approved):

PR #1853, opened 2025-04-01
Add ROFL to home page graph

buberdds:
Add ROFL bubble. Can be merge when MVP is accepted.


lubej (approved):
The spacing looks a bit disproportionate. I would guess that all bubbles branching from Sapphire should be updated. Unless it is exactly as in designs.

lubej:
> You are correct. Need to tweak it.

Looks good now 👍

PR #1876, opened 2025-04-08
Improve detecting if API is out-of-date

lukaw3d:
Fixes https://github.com/oasisprotocol/explorer/issues/882

### DIFF
--- a/src/app/components/OfflineBanner/hook.ts
+++ b/src/app/components/OfflineBanner/hook.ts
@@ -24,7 +24,7 @@ export const useIsApiReachable = (
 export type FreshnessInfo = {
   unavailable?: boolean
   outOfDate?: boolean
-  outOfDateReason?: false | 'indexer' | 'blocks'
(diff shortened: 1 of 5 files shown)


src/app/components/OfflineBanner/hook.ts
```diff
@@ -24,6 +24,7 @@ export const useIsApiReachable = (
 export type FreshnessInfo = {
   unavailable?: boolean
   outOfDate?: boolean
```
buberdds, line 26: It looks like redundancy in having two vars for handling one state. Doesn't `outOfDate` duplicate outOfDateReason? Would be simpler to have single source of truth for this state checking.

buberdds (approved):

src/app/components/OfflineBanner/hook.ts
lukaw3d: Yes
That can be a separate refactoring, I think it would require some hundred lines of changes in existing code and I'd rather avoid conflicts

PR #1877, opened 2025-04-08
Make runtime tx methods and event types stricter

lukaw3d:


buberdds (approved):

buberdds:
This needs some tweaks. Transactions created by ROFL instance are encrypted and they do not have known method 
`https://testnet.nexus.oasis.io/v1/sapphire/rofl_apps/rofl1qp55evqls4qg6cjw5fnlv4al9ptc0fsakvxvd9uw/instance_transactions?limit=10&offset=0`

lukaw3d:
looks like method is never `undefined` and our types were wrong. It's `""`

PR #1895, opened 2025-04-14
Create ROFL app instance details page

buberdds:
~For now 1k limit in API call is fine. In the near future we will have dedicated endpoint to get instance details.~

### DIFF
--- a/src/app/pages/RoflAppInstanceDetailsPage/index.tsx
+++ b/src/app/pages/RoflAppInstanceDetailsPage/index.tsx
@@ -2,10 +2,9 @@ import { FC } from 'react'
 import { useHref, useParams } from 'react-router-dom'
 import { useTranslation } from 'react-i18next'
 import Typography from '@mui/material/Typography'
-import { Layer, RoflInstance, useGetRuntimeRoflAppsIdInstances } from '../../../oasis-nexus/api'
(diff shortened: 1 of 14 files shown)


src/app/pages/RoflAppInstanceDetailsPage/index.tsx
```diff
@@ -0,0 +1,118 @@
...
+  const { t } = useTranslation()
+  const scope = useRequiredScopeParam()
+  const id = useParams().id!
+  const rak = useParams().rak!
+  const txLink = useHref('')
+  const [method, setMethod] = useTypedSearchParam('method', 'any', {
+    deleteParams: ['page'],
+  })
+  const context: RoflAppInstanceDetailsContext = { scope, id, rak, method, setMethod }
+  const instancesQuery = useGetRuntimeRoflAppsIdInstances(scope.network, Layer.sapphire, id, {
```
lubej, line 31: This seems highly inefficient. And I could see potential problem with pagination i.e. instances not being in the page you are querying.

buberdds: yes, I described this in PR description. For now there is no chance we will hit an issue, because number of instances is very low per app.

buberdds: New endpoint was added in Nexus, need to bump PR

lubej (approved):

PR #1902, opened 2025-04-16
Fix RoundedBalance prop combinations

lukaw3d:


src/app/components/RoundedBalance/__tests__/index.test.tsx
```diff
@@ -91,4 +93,45 @@ describe('RoundedBalance', () => {
       expect(screen.getByText('99,999.00000…')).toBeInTheDocument()
     })
   })
+
+  for (const compactLargeNumbers of [false, true]) {
```
buberdds, line 97: Every possible combination of params creates tons of useless snapshots. Can we define some test cases instead?

```
  describe('should match snapshot for key configurations', () => {
    const testCases = [
      // common
      { value: '0.002231', ticker: 'ROSE' },

      // compactLargeNumbers
      { value: '15000000', compactLargeNumbers: true },
      { value: '15000000', compactLargeNumbers: false },

      // showSign
      { value: '5', showSign: true },
      { value: '-0.0047956', showSign: true },

      // link 
      {
        value: '5',
        ticker: 'wROSE',
        tickerAsLink: true,
        tokenAddress: '0x8Bc2B030b299964eEfb5e1e0b36991352E56D2D3',
      },

      // Very small value
      { value: '0.00000000000002231' }, 
      // Very large value
      { value: '1111111111111111111111111111111111111111111111111111' }, 

      // Additional combos / whatever
      { value: '15000000', compactLargeNumbers: true, showSign: true, ticker: 'wROSE' },
      { value: '0.002231', tickerAsLink: true, tokenAddress: '0x8Bc2B030b299964eEfb5e1e0b36991352E56D2D3' },
    ]

    testCases.forEach(params => {
      it(`should match snapshot ${util.inspect(params)}`, () => {
         ...
      })
    })
  })
  ```

buberdds (approved):

src/app/components/RoundedBalance/__tests__/index.test.tsx
lukaw3d: We should also have more normal tests above but I also want to have all the snapshots because it is not implemented in a composable way

Issue #1950, opened 2025-05-06
LinkableDiv gets covered by sticky header


lukaw3d:
Lets fix this in a separate pullrequest (+ I don't think this is the most reliable and maintainable approach)

_Originally posted by @lukaw3d in https://github.com/oasisprotocol/explorer/pull/1890#discussion_r2069137120_

lukaw3d:
cc @kaja-osojnik that approach didn't work if user clicks the same link again

lukaw3d:
CSS has `scroll-margin` and `scroll-padding` for this

PR #1953, opened 2025-05-06
Fix LinkableDiv getting covered by sticky header

lukaw3d:
Fixes https://github.com/oasisprotocol/explorer/issues/1950

### DIFF
--- a/src/app/components/PageLayout/LinkableDiv.tsx
+++ b/src/app/components/PageLayout/LinkableDiv.tsx
@@ -1,3 +1,4 @@
+import GlobalStyles from '@mui/material/GlobalStyles'
 import { FC, useEffect, useRef } from 'react'
 import { useLocation } from 'react-router-dom'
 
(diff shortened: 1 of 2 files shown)


src/app/components/PageLayout/LinkableDiv.tsx
```diff
@@ -25,5 +26,11 @@ export const LinkableDiv: FC<JSX.IntrinsicElements['div'] & { id: string }> = pr
     }
   }, [id, hash, element])
 
-  return <div {...props} ref={divRef} />
+  return (
+    <>
+      {/* Don't scroll the target underneath sticky Header */}
+      <GlobalStyles styles="html { scroll-padding-top: 180px; }" />
```
buberdds, line 32: We use index.css as global styles. Why we can't define this CSS prop there? Can this affect layout somehow or it's just preference for this one particular case?

buberdds (approved):

PR #1960, opened 2025-05-14
ROFL app details page fixes

csillag:
**Tweak 1**: implement page title shortening in table and mobile mode

| Before | After | Comment |
| ----- | ---- | --- |
|  image | image  | No change in desktop size |
| image |image | Automatic shortening of title at tablet and mobile sizes  |

**Tweak 2**: Enable highlighting matching ROFL addresses on mouse hover

(Currently this is only enabled on desktop.)

image

**Tweak 3**:

Enable tool-tip on shortened app ids:

image

### DIFF
--- a/src/app/components/Link/index.tsx
+++ b/src/app/components/Link/index.tsx
@@ -1,10 +1,10 @@
-import { FC } from 'react'
+import { FC, PropsWithChildren } from 'react'
 import { Link as RouterLink } from 'react-router-dom'
 import { useScreenSize } from '../../hooks/useScreensize'
 import MuiLink from '@mui/material/Link'
 import Typography from '@mui/material/Typography'
 import { COLORS } from '../../../styles/theme/colors'
-import { trimLongString } from '../..//utils/trimLongString'
+import { trimLongString } from '../../utils/trimLongString'
 import { HighlightedText } from '../HighlightedText'
 import Box from '@mui/material/Box'
 import { AccountMetadataSourceIndicator } from '../Account/AccountMetadataSourceIndicator'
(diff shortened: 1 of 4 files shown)


csillag:
**Tweak 3**:

Enable tooltip on shortened app ids:

image

src/app/components/Link/index.tsx
```diff
@@ -118,11 +203,16 @@ const DesktopLink: FC<DesktopLinkProps> = ({ address, name, to, alwaysTrim, high
       </WithHighlighting>
     )
   }
+  const label = name ? <HighlightedText text={name} pattern={highlightedPart} /> : address
   return (
     <WithHighlighting address={address}>
-      <MuiLink component={RouterLink} to={to}>
-        {name ? <HighlightedText text={name} pattern={highlightedPart} /> : address}
-      </MuiLink>
+      {labelOnly ? (
```
lukaw3d, line 209: Any alternative ideas for future to avoid very component adding labelOnly mode?

lukaw3d (approved):

src/app/components/Link/index.tsx
csillag: > Any alternative ideas for future to avoid every component adding labelOnly mode?

So, the rationale was this:

We want to do a lot of things in our links. (Show a name, or an address, or both; do highlighting based on pointer and search pattern, do shortening, either fixed or adapting length, specific formatting). There are many combinations of this, some is based on configuration, some context dependent, etc.

We want to do all of these consistently both on links and labels. Duplicating the code would suck, so we keep everything together, and use the `labelOnly` flag to switching off linking when not needed.

An alternative would be to always have a pair of components, like `<WhateverLink>` and `<WhateverLabel>`, which would share their internal formatting and other behavior logic, except the link thing itself. Should we try that?

PR #1972, opened 2025-05-21
Show seconds in formatted timestamp

buberdds:
Slack req. Change affects block,tx,account,validator details and offline banner.


csillag (approved):
Reviewed. LGTM.

Issue #1907, opened 2025-04-18
Fix search results on narrow display


lukaw3d:
- open https://explorer.dev.oasis.io/mainnet/sapphire/search?q=rose
- make window 700px wide
- broken contract name display
- broken "verified" badge
- broken "at" in creator
- (not sure what to do about newline in balance)
- hover contract name: broken tooltip
- ![Image](https://github.com/user-attachments/assets/c84e142e-a0b4-4c32-96a3-ec02268bf9ee)

I think we should use `flex-wrap: wrap;` by default. Simply `* { flex-wrap: wrap; }` fixes most of these

Issue #1931, opened 2025-04-25
Tweak contract verification row on mobile


buberdds:
Vertical list on mobile:
![Image](https://github.com/user-attachments/assets/2b4d77de-85db-4af8-a118-e1d5851348ab)

lukaw3d:
Related https://github.com/oasisprotocol/explorer/issues/1907

PR #1974, opened 2025-05-21
Fix details view wrapping on mobile

lukaw3d:
Fixes https://github.com/oasisprotocol/explorer/issues/1907
image

### DIFF
--- a/src/app/components/Account/ContractCreatorInfo.tsx
+++ b/src/app/components/Account/ContractCreatorInfo.tsx
@@ -12,7 +12,6 @@ import { AppErrors } from '../../../types/errors'
 import { AccountLink } from './AccountLink'
 import Box from '@mui/material/Box'
 import Skeleton from '@mui/material/Skeleton'
-import { useScreenSize } from '../../hooks/useScreensize'
 
 const TxSender: FC<{ scope: SearchScope; txHash: string; alwaysTrim?: boolean }> = ({
   scope,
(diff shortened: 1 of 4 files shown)


src/app/components/Account/ContractCreatorInfo.tsx
```diff
@@ -48,26 +47,17 @@ export const ContractCreatorInfo: FC<{
...
-        flexDirection: isMobile ? 'column' : 'row',
-        alignItems: isMobile ? 'flex-start' : 'center',
-        gap: isMobile ? 0 : 2,
-        minWidth: '25%',
-      }}
-    >
+    <>
       <TxSender scope={scope} txHash={creationTxHash} alwaysTrim={alwaysTrim} />
-      <Box>{t('contract.createdAt')}</Box>
+      <Box>&nbsp;{t('contract.createdAt')}&nbsp;</Box>
```
buberdds, line 58: I would remove the first &nbsp; as addr has as extra padding anyway (highlighting). on the other hand it doesn't matter as hash is misaligned as well due to MaybeWithTooltip wrapper?

buberdds (approved):

buberdds:
Fixes https://github.com/oasisprotocol/explorer/issues/1931 too

src/app/components/Account/ContractCreatorInfo.tsx
lukaw3d: I don't think we want highlighting to affect layout with its padding, so I added it in anticipation of fixed highlighting

PR #1987, opened 2025-05-28
Add support for filtering Rofl app list by name

csillag:
The Nexus API has already added support for filtering the Rofl app list by name in https://github.com/oasisprotocol/nexus/pull/997.

In Explorer, currently this is only unitized in the global search since #1973/. (So that we can find Rofl apps from the global search field.)

This PR adds name/based filtering to the Rofl app list page itself.

I suggest reviewing commit by commit, because the different commits add different (small) features, and also there were some preparatory code refactorings, which would otherwise obscure the actual changes.

## Screenshots

Filtering the list on the desktop

image

When no match:

image

On mobile:

image

On mobile, vertical mode:

image

No match on mobile:

image

### DIFF
--- a/src/app/pages/RoflAppsPage/index.tsx
+++ b/src/app/pages/RoflAppsPage/index.tsx
@@ -1,46 +1,36 @@
-import { FC, useEffect, useState } from 'react'
+import { FC } from 'react'
 import { useTranslation } from 'react-i18next'
 import Divider from '@mui/material/Divider'
-import { useGetRuntimeRoflApps } from '../../../oasis-nexus/api'
 import { AppErrors } from '../../../types/errors'
-import { NUMBER_OF_ITEMS_ON_SEPARATE_PAGE, paraTimesConfig } from '../../../config'
+import { paraTimesConfig } from '../../../config'
 import { useScreenSize } from '../../hooks/useScreensize'
 import { useRuntimeScope } from '../../hooks/useScopeParam'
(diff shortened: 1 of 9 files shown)


csillag:
The playwright tests are failing because of something with the cookie consent form, probably unrelated.

csillag:
> The playwright tests are failing because of something with the cookie consent form, probably unrelated.

No, actually the tests are broken by having an auto-focus field on one of the pages. That is not crucial, I have disabled it now, and now all the tests are passing. Later it would be nice to update the tests so that they are not broken by auto-focus.

src/app/pages/RoflAppsPage/index.tsx
```diff
@@ -1,88 +1,128 @@
...
+    layer,
+  )
+  const searchBar = (
+    <TableSearchBar
+      value={wantedNameInput}
+      onChange={setWantedNameInput}
+      placeholder={t('rofl.searchByNameOrKeyword')}
+      warning={nameError}
+      fullWidth={isMobile}
+      // autoFocus={!isMobile}
```
lukaw3d, line 101: This wasn't breaking matomo e2e tests, they are just flaky. I re-ran the CI on previous commit and it passed now :shrug:

csillag: Ah ok, in that case I'm re-enabling it.

csillag:
Closing in favor of #2013

PR #1994, opened 2025-05-29
Votes page: adjust filter positions on layout

csillag:
@donouwens has requested some adjustments on the network proposal votes page.

Changes made:
 - ~Add missing checkmark from the selected pill~ - moving to separate PR
 - Change the order of the pills and the search bar
 - Make it so that the components are further apart, the pills right next to the title
 - Adjust the drop shadow around the search bar

| Before | After |
|---------|----------|
| image | image   |
|image | image  |

### DIFF
--- a/src/app/components/Search/TableSearchBar.tsx
+++ b/src/app/components/Search/TableSearchBar.tsx
@@ -18,10 +18,29 @@ export interface TableSearchBarProps {
   placeholder: string
   warning?: string
   value: string
+  /**
+   * Should we go 100% width?
+   */
+  fullWidth?: boolean
+
+  /**
+   * Width of the whole component
+   *
+   * Note: fullWidth overrides this.
+   */
+  width?: string | number
+
   onChange: (value: string) => void
 }
 
-export const TableSearchBar: FC<TableSearchBarProps> = ({ value, onChange, placeholder, warning }) => {
+export const TableSearchBar: FC<TableSearchBarProps> = ({
+  value,
+  onChange,
+  placeholder,
+  warning,
+  fullWidth,
+  width = 250,
+}) => {
   const { isTablet } = useScreenSize()
 
   const [isWarningFresh, setIsWarningFresh] = useState(false)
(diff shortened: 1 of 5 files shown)


buberdds (commented):
1) changes affected tokens page negatively (header is jumping when using filter buttons)
2) imo rendering conditionally one icon in a button group is weird UX. every time user changes filter, two buttons change width. Is this really good UX pattern?

src/styles/theme/defaultTheme.ts
```diff
@@ -425,6 +425,9 @@ export const defaultTheme = createTheme({
     },
     MuiCardHeader: {
       styleOverrides: {
+        content: {
+          display: 'inline-flex',
```
buberdds, line 429: not going to test everything, but this breaks headers layout across app, samples:

 pr
Screenshot from 2025-06-02 13-32-14
Screenshot from 2025-06-02 13-30-19

vs master
Screenshot from 2025-06-02 13-30-46
Screenshot from 2025-06-02 13-30-12

csillag: OK I have completely removed fro this PR everything about the pills. So, no more checkmark, no more jumping. No more header issues, hopefully. Let's handle that in a new PR, because it is not strictly related to the vote filter layout issue.

csillag:
> Not sure I am listing all behaviors, but I assume we are aiming to have only v1 and v2, is this correct @donouwens ?> 

I have restricted it to v1 and v2 behavior now. @buberdds

src/app/components/Search/TableSearchBar.tsx
```diff
@@ -109,7 +135,7 @@ export const TableSearchBar: FC<TableSearchBarProps> = ({ value, onChange, place
         inputProps: {
           sx: {
             p: 0,
-            width: isTablet ? 110 : 300,
+            width: fullWidth ? '100%' : isTablet ? tabletWidth : desktopWidth,
```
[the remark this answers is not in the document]
csillag: Simplified accordingly.

Issue #1535, opened 2024-09-11
Buggy address trimming on mobile


lukaw3d:
Related to https://github.com/oasisprotocol/explorer/issues/1399

https://github.com/user-attachments/assets/da5e2bb0-e3c9-4c9d-8a51-604bb2da38a0

Cause: scrolling up and down expands URL bar and triggers window resize events

lukaw3d:
Can be fixed by listening to screen width changes instead of all resizes

csillag:
> Can be fixed by listening to screen width changes instead of all resizes

This suggestion has been implemented in #2000

PR #1997, opened 2025-05-30
Prevent WithHighlighting from taking up more space than original

lukaw3d:


src/app/pages/RuntimeTransactionDetailPage/index.tsx
```diff
@@ -219,11 +214,13 @@ export const RuntimeTransactionDetailView: FC<{
             <>
               <dt>{t('common.to')}</dt>
               <dd>
-                <AccountLink
-                  scope={transaction}
-                  address={(transaction?.to_eth || transaction?.to) as string}
-                />
-                <CopyToClipboard value={(transaction?.to_eth || transaction?.to) as string} />
+                <Box sx={{ display: 'inline-flex', alignItems: 'center' }}>
```
lukaw3d, line 217: Related https://github.com/oasisprotocol/explorer/issues/1535

"To" field does resize, but copy button was in its own line
image

lukaw3d: Moving out to a separate PR

src/app/components/StyledDescriptionList/index.tsx
```diff
@@ -64,6 +64,7 @@ export const StyledDescriptionList = styled(InlineDescriptionList, {
     alignItems: 'center',
     maxWidth: '100%',
     overflowX: 'hidden',
+    paddingLeft: '5px', // Needed to fix highlightedAddress being clipped by overflow-x: hidden
```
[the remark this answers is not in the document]
lukaw3d: Yeah; but `marginLeft: '-5px'` still loses the pixels, but now they are at the end. and `marginRight: '-5px'` doesn't help :shrug: 
myimage


I would keep it like this - we don't have any spacing between dt and dd anyway. If "Endorsement" in roflapps had an extra space it would show the need for a gap or padding
pr-1997 oasis-explorer pages dev_mainnet_sapphire_rofl_app_rofl1qpdzzm4h73gtes04xjn4whan84s3k33l5gx787l2 (1)

PR #2002, opened 2025-05-30
Layer handling code cleanup

csillag:
- Replace string literals with proper constants
- Introduce ConsensusScope and RuntimeScope
  for situations when we expect a specific type of scope.
- Introduce useRuntimeScope() and useConsensusScope()
  hooks which verify that we have the right kind of scope.
  (This makes checking in the individual components unnecessary.)
- Remove hard-wiring "sapphire" for all rofl-related pages
  (We already have feature flags for runtimes to enable/disable rofl)

### DIFF
--- a/src/app/components/Search/SearchSuggestionsLinksForNoResults.tsx
+++ b/src/app/components/Search/SearchSuggestionsLinksForNoResults.tsx
@@ -13,6 +13,7 @@ import { SearchScope } from '../../../types/searchScope'
 import { useScreenSize } from '../../hooks/useScreensize'
 import { SxProps } from '@mui/material/styles'
 import { Layer } from '../../../oasis-nexus/api'
+import { Network } from '../../../types/network'
 
 interface Props {
   scope: SearchScope | undefined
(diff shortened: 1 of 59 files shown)


src/app/components/Search/SearchSuggestionsLinksForNoResults.tsx
```diff
@@ -31,7 +32,7 @@ export const SearchSuggestionsLinksForNoResults: FC<Props> = ({ scope }) => {
   const { isMobile } = useScreenSize()
   const { suggestedBlock, suggestedTransaction, suggestedAccount, suggestedTokenFragment } =
     (scope?.network && scope?.layer && searchSuggestionTerms[scope.network][scope.layer]) ??
-    searchSuggestionTerms['mainnet']['sapphire']!
```
lukaw3d, line 34: I somewhat prefer string literals :shrug:

lukaw3d (approved):

src/app/components/Search/SearchSuggestionsLinksForNoResults.tsx
csillag: I am fine with either method, but up to now we had 95% constants, and 5% string literals. At least now it's uniform...

csillag: > I somewhat prefer string literals 🤷

See https://github.com/oasisprotocol/explorer/pull/2016 🤷🏻

Issue #1992, opened 2025-05-29
Collapse contract verification badge in mobile tables


lukaw3d:
![Image](https://github.com/user-attachments/assets/2f823903-8240-4f66-8c13-fa8e8782dd45)

Mocks:
![Image](https://github.com/user-attachments/assets/f468cd34-7d64-489b-ae44-942a327d6f09)

lukaw3d:
Unresolved issue: how to shorten column header "Verification"

Maybe name it "src" or ".sol"
cc @donouwens

kaja-osojnik:
PR: [#2003](https://github.com/oasisprotocol/explorer/pull/2003)

PR #2003, opened 2025-05-30
Hide label of verification badge in a table for mobile screens

kaja-osojnik:
Implements: [#1992](https://github.com/oasisprotocol/explorer/issues/1992)

Before:
Screenshot 2025-05-30 at 20 26 46

After:
Screenshot 2025-05-30 at 20 26 22

### DIFF
--- a/src/app/components/ContractVerificationIcon/index.tsx
+++ b/src/app/components/ContractVerificationIcon/index.tsx
@@ -14,8 +14,9 @@ export const verificationIconBoxHeight = 28
 
 type ContractStatusProps = {
   verificationLevel?: 'full' | 'partial'
+  hideLabel?: boolean
 }
-export const ContractStatus = ({ verificationLevel }: ContractStatusProps) => {
+export const ContractStatus = ({ verificationLevel, hideLabel }: ContractStatusProps) => {
   const { t } = useTranslation()
   const statusLabel =
     verificationLevel === 'full'
@@ -26,7 +27,9 @@ export const ContractStatus = ({ verificationLevel }: ContractStatusProps) => {
   const statusVariant =
     verificationLevel === 'full' ? 'success' : verificationLevel === 'partial' ? 'partialsuccess' : 'danger'
 
-  return <StatusBadge label={statusLabel} variant={statusVariant} />
+  const label = hideLabel ? '' : statusLabel
+
+  return <StatusBadge label={label} variant={statusVariant} />
 }
 
 export const VerificationIcon: FC<{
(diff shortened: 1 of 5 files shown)


src/app/components/common/StatusBadge.tsx
```diff
@@ -57,14 +59,19 @@ const StyledBadge = styled(Box, {
     borderRadius: 10,
     padding: 4,
     paddingLeft: 10,
-    paddingRight: 5,
+    paddingRight: compact ? 5 : 10,
```
buberdds, line 62: Currently there is no need to use `compact`  prop as padding-right value will be always overwritten on mobile by a next CSS rule
```
    [theme.breakpoints.down('sm')]: {
      padding: 4,
    },
```

src/app/components/common/StatusBadge.tsx
```diff
@@ -57,14 +59,19 @@ const StyledBadge = styled(Box, {
...
+    [theme.breakpoints.down('sm')]: {
+      padding: 4,
+    },
   }
 })
 
-export const StatusBadge = ({ label, icon, variant = 'info' }: StatusBadgeProps) => {
+export const StatusBadge = ({ label, icon, variant = 'info', hideLabelMobile = false }: StatusBadgeProps) => {
```
buberdds, line 69: I think this all could work without changes in StatusBadge too

buberdds (commented):
1) There is some `&nbsp; &nbsp;` spacing in `ContractStatus` which is now more visible with compact badges in tables. Maybe it's better to render these extra spaces conditionally?

2) I think we don't need StatusBadge.tsx changes. Once reverted I would change prop name to  `hideLabel` (or opposite `withLabel`) and just change `ContractStatus`  to sth like 
` return <StatusBadge label={hideLabel ? '' : statusLabel} variant={statusVariant} hideLabel={hideLabel} />`

src/app/components/ContractVerificationIcon/index.tsx
```diff
@@ -49,13 +55,13 @@ export const VerificationIcon: FC<{
   }
   const Component = noLink ? Box : (Link as React.ElementType)
   const componentProps = noLink ? {} : sourcifyLinkProps
-
+  const hideLabelMobile = hideLabel && isMobile
```
lukaw3d, line 58: I'll fix: table component should decide to hide label on mobile. Not ContractStatus, not VerificationIcon

src/app/components/ContractVerificationIcon/index.tsx
```diff
@@ -49,13 +55,13 @@ export const VerificationIcon: FC<{
...
   return (
     <>
       <Component {...componentProps}>
-        <ContractStatus verificationLevel={verificationLevel} />
+        <ContractStatus verificationLevel={verificationLevel} hideLabel={hideLabelMobile} />
       </Component>
-      &nbsp; &nbsp;
+      {hideLabel && isMobile ? null : '\u00A0\u00A0'}
```
lukaw3d, line 64: I'll fix: spacing here is not related to label nor isMobile. It should be displayed depending on noLink

src/app/components/common/StatusBadge.tsx
```diff
@@ -58,6 +58,9 @@ const StyledBadge = styled(Box, {
     padding: 4,
     paddingLeft: 10,
     paddingRight: 5,
+    [theme.breakpoints.down('sm')]: {
+      padding: 4,
+    },
```
lukaw3d, line 63: I'll fix: spacing here is not related to mobile. It should be displayed depending on hideLabel

Issue #1999, opened 2025-05-30
[Bug]: Single Panel XHR Failure Wipes Entire Dashboard Content


uscinski:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### SUMMARY

Currently, when viewing a dashboard composed of multiple independent panels, if even a single asynchronous data request (XHR) for one of these panels fails (e.g., due to a 503 server error/timeout), the entire dashboard content is wiped, and a generic error message is displayed ("unknown error") for the whole page. This leads to an unproductive user experience.

### STEPS TO REPRODUCE

1. Open a page (dashboard) with multiple data panels (cards) (e.g., `/mainnet/consensus`).
2. Observe that the panels load data via individual XHR requests.
3. Simulate a 503 error using the browser developer tools.
  a. Reload the page.
  b. Navigate to the network tab.
  c. Filter XHR requests.
  d. Choose one and block it (e.g., https://nexus.oasis.io/v1/consensus/accounts?limit=5).
  e. Reload the page.
4. Observe the entire page content being replaced by a full-page error message, even if other panels loaded successfully.

Please note that step 3 is particularly crafted for the sake of reproducibility. In practice, it would happen on Nexus error, e.g., due to a request timeout.

### EXPECTED BEHAVIOR

If an XHR request for a single panel fails (e.g., 503 timeout), in my opinion the following should occur:
* All other panels that successfully loaded their data should remain visible and interactive.
* Only the specific panel whose data request failed should display an error state within its own boundaries.
* The error message within the failed panel should be clear, concise, and ideally, offer an option to retry loading that specific panel's data.
* The overall dashboard layout and functionality should remain intact.

csillag:
@uscinski please see https://github.com/oasisprotocol/explorer/pull/2004 for a partial solution. (Consensus only, for now.)

csillag:
How about fixing it only for a few select cards?

csillag:
@lukaw3d should I simplify https://github.com/oasisprotocol/explorer/pull/2004 (i.e. remove the special-purpose code for marking unreachable paratimes), or kill it altogether?

csillag:
Updated PR accordingly

csillag:
With both dashboard and paratime PR merged, this is now closed.

PR #2004, opened 2025-05-30
Improve consensus dashboard resilience

csillag:
Improves  #1999 

Now a failing query won't collapse the whole consensus dashboard, only the relevant card.
Every card has been split to two component, and the internal parts are now wrapped in the appropriate error boundaries.
The same could be done to the paratime dashboard, too.
Plus, the design could obviously improved, but this already avoids the collapsing of the whole board, so definitely improves the UX.

## Before, if any of the queries have a problem:

image

Instead of this, now we get:

## Failing to load blocks:

image

## Failing to load validators:

image

## Failing to load paratimes:

image

## Failing to load accounts:

image

## Failing to load transactions:

image

## Network change proposals:

image

## TX Stats:

image

### DIFF
--- a/.changelog/2004.feature.md
+++ b/.changelog/2004.feature.md
@@ -1,0 +1,1 @@
+Improve Consensus dashboard resilience
--- a/src/app/components/ErrorBoundary/index.tsx
+++ b/src/app/components/ErrorBoundary/index.tsx
@@ -1,4 +1,4 @@
-import { Component } from 'react'
+import { Component, ReactNode } from 'react'
 import { ErrorDisplay } from '../ErrorDisplay'
 
 type HasChildren = {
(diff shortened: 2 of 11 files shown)


csillag:
The original error suggested:

> The error message within the failed panel should be clear, concise, and ideally, offer an option to retry loading that specific panel's data.

Unfortunately, the retry / reload is not as trivial as it seems, because we throw an exception, and when an error boundary catches the error, the whole components gets into an "error" state, and it is not trivial to recover from. A browser refresh seems to be the easiest solution.

csillag:
> I guess simplify. Can some card be an error boundary for runtime status

Simplified.

lukaw3d (approved):

PR #2008, opened 2025-06-02
FilterButtons: Add missing checkmark

csillag:
According to the design, we are supposed to have checkmarks on the selected filter buttons:

image

This PR adds them.

| | Before | After |
| --- | --- | --- |
| Token list  | image  | image |
| Votes on network proposals |  image |image|

### DIFF
--- a/src/app/components/FilterButtons/index.tsx
+++ b/src/app/components/FilterButtons/index.tsx
@@ -1,7 +1,6 @@
 import Box from '@mui/material/Box'
 import Chip from '@mui/material/Chip'
 import CheckIcon from '@mui/icons-material/Check'
-import Typography from '@mui/material/Typography'
 import { COLORS } from '../../../styles/theme/colors'
 
 type FilterButtonsProps<T extends string> = {
@@ -12,7 +11,7 @@ type FilterButtonsProps<T extends string> = {
 
 export const FilterButtons = <T extends string>({ options, onSelect, value }: FilterButtonsProps<T>) => {
   return (
-    <Box sx={{ display: 'inline-flex' }}>
+    <Box component="span" fontSize={25}>
       {options.map(option => {
         const selected = option.value === value
         return (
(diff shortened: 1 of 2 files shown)


src/app/components/FilterButtons/index.tsx
```diff
@@ -21,7 +22,11 @@ export const FilterButtons = <T extends string>({ options, onSelect, value }: Fi
             clickable
             color="secondary"
             label={
-              <Box sx={{ display: 'flex', alignItems: 'center', justifyContent: 'center' }}>
+              <Box
+                component="span"
+                sx={{ display: 'flex', alignItems: 'baseline', justifyContent: 'center', gap: 2 }}
+              >
+                {selected && <CheckIcon fontSize={'small'} sx={{ transform: 'translateY(4px)' }} />}
```
lukaw3d, line 29: erc721 looks misaligned inside the chip
image

lukaw3d, line 29: `alignItems: 'baseline' + 'translateY(4px)'` looks strange. how come not align center

csillag: Indeed. Fixed. (See the updated screenshot above.)

csillag: If I set "alignItems: center`, the vertical position of the buttons keep changing as I change the selection. As far as I could tell from a cursory look, this is because there is a component higher in the DOM  (some kind of card header I believe) that is also using baseline alignment, and adding the check-mark somehow messed it up. This issue doesn't occur when I use the baseline alignment.

lukaw3d: I really don't like patching bad css with bad css

src/app/components/FilterButtons/index.tsx
```diff
@@ -21,7 +22,11 @@ export const FilterButtons = <T extends string>({ options, onSelect, value }: Fi
             clickable
             color="secondary"
             label={
-              <Box sx={{ display: 'flex', alignItems: 'center', justifyContent: 'center' }}>
+              <Box
+                component="span"
+                sx={{ display: 'flex', alignItems: 'baseline', justifyContent: 'center', gap: 2 }}
+              >
+                {selected && <CheckIcon fontSize={'small'} sx={{ transform: 'translateY(4px)', mt: -5 }} />}
```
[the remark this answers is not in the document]
csillag: Yes it does. This is only for the check-mark. Adding the checkmark pushing down the text, since the layout engine got the impression that the top of the check-mark doesn't fit to the chip, so the whole thing (check-mark _end_ text) must be placed lower. This is why the ERC-721 text was too low earlier. Telling not to worry about space at the top of the check-mark via the big negative margin solved that issue. (In the sense of, patching bad css with more bad css, as you have spelled it out above.)

src/app/components/FilterButtons/index.tsx
csillag: In any case, I expect that this component (along with the others) will be overhauled / replaced during the ui-library transition really soon, so as long as it's fully functional and looks good, I wouldn't worry about the long-term maintainability of the css code under the hood.

buberdds: I think we will remove this next sprint (uil replacement), so making such adjustments now is a waste of time

PR #2013, opened 2025-06-03
Add support for filtering Rofl app list by name  (flex search version)

csillag:
This PR is an updated version of #1987, rebased on top of #2012, so that it has the latest multi-word flex search.

See demo of flex search in the Rofl app list here: https://pr-2013.oasis-explorer.pages.dev/testnet/sapphire/rofl/app?name=tg+cha

Original description follows.

===========

The Nexus API has already added support for filtering the Rofl app list by name in https://github.com/oasisprotocol/nexus/pull/997.

In Explorer, currently this is only unitized in the global search since #1973/. (So that we can find Rofl apps from the global search field.)

This PR adds name/based filtering to the Rofl app list page itself.

I suggest reviewing commit by commit, because the different commits add different (small) features, and also there were some preparatory code refactorings, which would otherwise obscure the actual changes.

## Screenshots

Filtering the list on the desktop

image

When no match:

image

On mobile:

image

On mobile, vertical mode:

image

No match on mobile:

image

### DIFF
--- a/.changelog/2013.feature.md
+++ b/.changelog/2013.feature.md
@@ -1,0 +1,1 @@
+Add support for filtering Rofl app list by name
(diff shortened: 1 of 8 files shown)


.changelog/1987.feature.md
```diff
@@ -0,0 +1 @@
+Add support for filtering Rofl app list by name
```
lukaw3d, line 1: filename

lukaw3d (approved):

PR #2016, opened 2025-06-04
Prefer string literals instead of enum constants

csillag:
This is a follow-up to #2002, where @lukaw3d [commented](https://github.com/oasisprotocol/explorer/pull/2002#discussion_r2122324404):

> I somewhat prefer string literals :shrug: 

So, this PR increases the usage of string literals, instead of `Layer` and `Network` constants.

So,
 * `Network.mainnet` => `'mainnet'`
 * `Layer.sapphire` => `'sapphire'`
 
etc.


tjanez:
@buberdds, is this approach of using string literals also good with you?

csillag:
> @buberdds, is this approach of using string literals also good with you?

@buberdds 's input (via other channels):

> I don't care really [...]  I fine with merging 

So, merging it

PR #2017, opened 2025-06-04
Don't show fake fiat zero value on Gecko failures

### LINKED ISSUE #2006: If coingecko fails we should not show fiat value=0.00
We should just show the warning
Image

csillag:
Fixes #2006.

Test URL: https://pr-2017.oasis-explorer.pages.dev/mainnet/sapphire/address/0x8D8E302aDf83948F22c8aefb26755b059f958893

### DIFF
--- a/.changelog/2017.bugfix.md
+++ b/.changelog/2017.bugfix.md
@@ -1,0 +1,1 @@
+Don't show fake fiat zero value on Gecko failures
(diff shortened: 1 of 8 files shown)


csillag:
btw, why do we have separate  `FiatMoneyAmount` and `CurrentFiatValue` components?

csillag:
OK, so it turns out the we are using `FiatMoneyAmount` for runtime accounts, and `CurrentFiatValue` for runtime and consensus TXs. It's not completely unreasonable to have two components, because one component is about a specific TX using a specific token, while the other is about a sum of values. I don't want to unite them right away.

However, the component we use for TXs did'nt have any warning about having no data from Gecgo, it just hid the USD value in that case. Now I have fixed that, too, using the same mechanism and same UI element.

Issue #2036, opened 2025-06-13
Expected type to be exhausted should not throw in production


lukaw3d:
https://explorer.oasis.io/testnet/sapphire/address/oasis1qp2ens0hsp7gh23wajxa4hpetkdek3swyyulyrmz
> Unknown runtime tx method: Expected type to be exhausted, but this type was not handled: "roflmarket.ProviderUpdateOffers"

![Image](https://github.com/user-attachments/assets/927dc3b3-4294-4bfb-8a78-daf0eed51889)

This probably broke when switching to vite:
https://github.com/oasisprotocol/explorer/blob/0adc777ead7fd195bdbc6501f243b6294f319fc3/src/types/errors.ts#L47-L51

PR #2037, opened 2025-06-13
Fix detecting production env (fix exhausted type error in production)

lukaw3d:
Fixes https://github.com/oasisprotocol/explorer/issues/2036
https://explorer.oasis.io/testnet/sapphire/address/oasis1qp2ens0hsp7gh23wajxa4hpetkdek3swyyulyrmz


lubej (approved):

src/types/global.d.ts
```diff
@@ -1,7 +1,7 @@
 declare global {
   interface ImportMeta {
     env: {
-      NODE_ENV: 'development' | 'production' | 'test'
```
lubej, line 4: With `/// <reference types="vite/client" />` you wouldn't need this line, as `PROD` is a "built-in" constant in vite.

lukaw3d: +1 moving that into https://github.com/oasisprotocol/explorer/issues/2040

PR #1516, opened 2024-08-22
Append EVM function name to unencrypted verified contract calls

lukaw3d:
Part of https://github.com/oasisprotocol/explorer/issues/1222

Issue #1222, opened 2024-02-01
Use new API fields to show parsed tx input: evm_fn_name, evm_fn_params, revert_params


lukaw3d:
https://github.com/oasisprotocol/explorer/commit/86ff7f729ec986388df07f76cb97132273751f00

lukaw3d:
Quick mocks so far
https://www.figma.com/design/8dwSyYOOm3vlgbvxPGT29p/Block-Explorer-(Post-Launch)?node-id=6092-235792&t=9jFIaNmujesljTaq-0
![image](https://github.com/oasisprotocol/explorer/assets/3758846/b335dd7f-e208-477d-888f-c7b733e038ba)

lukaw3d:
example transactions:

- with nested structure 
  https://explorer.dev.oasis.io/mainnet/emerald/tx/0x8c3de7ea493e6ce9c7a8f100739a144e588eb295dcfc2fd845ea1c0e4586e173

- multiple field types
  https://explorer.oasis.io/mainnet/emerald/tx/0x8c4ac733390969f55df6c2acff671028175a90dd26d1c490ca863a2d7e645f03

lukaw3d:
https://explorer.dev.oasis.io/mainnet/sapphire/tx/0x9ce91f01d822bb4144de04ad4fce4711eb005c7d2824bfe38f670bf2916727ca
huge list

lukaw3d:
Here are the updated (final) visuals:

Minimised state: 
https://www.figma.com/design/8dwSyYOOm3vlgbvxPGT29p/Block-Explorer-(Post-Launch)?node-id=11019-75589&t=JEx1LAgfTQ4lZZsm-1

Open state:
https://www.figma.com/design/8dwSyYOOm3vlgbvxPGT29p/Block-Explorer-(Post-Launch)?node-id=11128-74527&t=JEx1LAgfTQ4lZZsm-1

lukaw3d:
Can probably ignore `revert_params`. Enough info is already in error message:
https://explorer.dev.oasis.io/mainnet/sapphire/tx/0x6953df2693cbb055d5670a06316d0bb0724716cc0c460dc992d2315403fa8e02
https://explorer.dev.oasis.io/mainnet/sapphire/tx/0x8c005014730804b230dadb79c448451a69715a2956aae26f5c3bc951ee416818

PR #2065, opened 2025-07-09
Display evm_fn_params in transaction details

lukaw3d:
Related https://github.com/oasisprotocol/explorer/pull/1516
Part of https://github.com/oasisprotocol/explorer/issues/1222

https://pr-2065.oasis-explorer.pages.dev/testnet/sapphire/tx/0xd4559c82595299c6a39d56387c97b7c3e6c9ac49cdd691179b9ebb69ac4f35a6
https://pr-2065.oasis-explorer.pages.dev/testnet/sapphire/tx/0x22c3c69ca9d42edd0fc0f945b969f3fbe7f574aca815379475f1e636c38c766e
Example 1 | Example 2
image


src/app/utils/yamlDump.ts
```diff
@@ -0,0 +1,10 @@
+// eslint-disable-next-line no-restricted-imports
+import { dump } from 'js-yaml'
+
+export function yamlDump(obj: any) {
+  return dump(obj, {
+    replacer: (k, v) => {
+      return v
```
csillag, line 7: Why don't we need the keys ?

csillag (approved):
Reviewed. LGTM.

(I had a question but it's non-blocking.)

src/app/utils/yamlDump.ts
lukaw3d: Key can be used to change the value return format, but it isn't itself expected to be returned. This replacer is noop - until next PR

PR #2068, opened 2025-07-12
Display subcall params

lukaw3d:


src/app/utils/yamlDump.ts
```diff
@@ -4,6 +4,8 @@ import { dump } from 'js-yaml'
 export function yamlDump(obj: any) {
   return dump(obj, {
     replacer: (k, v) => {
+      if (v instanceof Map) return Object.fromEntries(v.entries())
+      if (typeof v === 'bigint') return v.toString()
```
csillag, line 8: Are we sure that v will always be either bigint or Map or something that can be used as string without further effort? Or is this something that we will just expand based on experience with real data?

lukaw3d: I checked that it covers everything in types of transactions that have been made so far. It is possible that CBOR encoding allows more types that we will need too add here (not sure if it supports Set, regex, or undefined).

```js
console.log(yamlDump({a: 5, b: new Date(), c: [], d: new Map([['1','2']]), 
  f: new Uint8Array([1,2,3]), g: 56n, h: undefined, i: null}))
>
a: 5
b: 2025-07-14T00:29:52.243Z
c: []
d:
  '1': '2'
f: !!binary AQID
g: '56'
i: null
```

lukaw3d: Added undefined as
```js
h: '!!undefined'
```

PR #2078, opened 2025-07-21
Support filtering runtime event lists by type

csillag:
Add support for filtering Runtime Event lists by type. This works on:

- Runtime Account events
- Runtime transaction events
- Runtime block events

Besides the real event types, there is a pseudo-type called "Token transfer", which actually means "EVM log message with a signature denoting a token transfer.

Currently there is one unresolved issue: token transfer are returned both when filtering for "Token transfers" and "EVM log messages", since they _are_ technically both. To solve this we need some support from the Nexus side, but it's unlikely to land during the current sprint, so I say we should merge this now, and improve later when the Nexus team has the bandwidth to add what we need.

### DIFF
--- a/src/app/components/RuntimeEvents/RuntimeEventsDetailedList.tsx
+++ b/src/app/components/RuntimeEvents/RuntimeEventsDetailedList.tsx
@@ -16,11 +16,15 @@ export const RuntimeEventsDetailedList: FC<{
   isError: boolean
   pagination: false | TablePaginationProps
   showTxHash: boolean
-}> = ({ scope, events, isLoading, isError, pagination, showTxHash }) => {
+  filtered: boolean
+}> = ({ scope, events, isLoading, isError, pagination, showTxHash, filtered }) => {
   const { t } = useTranslation()
   return (
     <>
       {isError && <CardEmptyState label={t('event.cantLoadEvents')} />}
+      {!isError && filtered && (!pagination || pagination.selectedPage === 1) && !events?.length && (
+        <CardEmptyState label={t('tableSearch.noMatchingResults')} />
+      )}
       {isLoading && <TextSkeleton numberOfRows={10} />}
       {events &&
         events.map((event, index) => (
--- a/src/app/components/RuntimeEvents/RuntimeEventsDetailedList.tsx
+++ b/src/app/components/RuntimeEvents/RuntimeEventsDetailedList.tsx
@@ -8,6 +8,8 @@ import { TextSkeleton } from '../Skeleton'
 import { RuntimeEventDetails } from './RuntimeEventDetails'
 import Box from '@mui/material/Box'
 import Divider from '@mui/material/Divider'
+import { AppErrors } from '../../../types/errors'
+import { EmptyState } from '../EmptyState'
 
 export const RuntimeEventsDetailedList: FC<{
   scope: SearchScope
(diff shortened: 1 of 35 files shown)


matevz (commented):
Works as expected, thanks!

src/app/components/RuntimeEvents/RuntimeEventsDetailedList.tsx
```diff
@@ -16,11 +16,15 @@ export const RuntimeEventsDetailedList: FC<{
...
+  filtered: boolean
+}> = ({ scope, events, isLoading, isError, pagination, showTxHash, filtered }) => {
   const { t } = useTranslation()
   return (
     <>
       {isError && <CardEmptyState label={t('event.cantLoadEvents')} />}
+      {!isError && filtered && (!pagination || pagination.selectedPage === 1) && !events?.length && (
+        <CardEmptyState label={t('tableSearch.noMatchingResults')} />
+      )}
       {isLoading && <TextSkeleton numberOfRows={10} />}
```
lukaw3d, line 28: https://pr-2078.oasis-explorer.pages.dev/mainnet/sapphire/address/0x289B7243138F376D8898F53Aedb63d943c382D02/events?event_type=accounts.burn
shows this while it is loading
image

src/locales/en/translation.json
```diff
@@ -420,7 +420,7 @@
     "totalSupply": "Total Supply",
     "totalSupplyValue": "{{ value, number }} {{ticker}}",
     "transferEventType": {
-      "transfer": "Transfer",
+      "transfer": "Token transfer",
```
lukaw3d, line 423: A naming problem:
"Token transfer" now means ERC20 tokens; but "Tokens minted" means native

https://pr-2078.oasis-explorer.pages.dev/testnet/sapphire/address/0xC3ecf872F643C6238Aa20673798eed6F7dA199e9/events?event_type=token_transfer
image

https://pr-2078.oasis-explorer.pages.dev/testnet/sapphire/address/0xC3ecf872F643C6238Aa20673798eed6F7dA199e9/events?event_type=accounts.mint
image

also native burn
image

csillag: I think the clearest solution would be to either
 * add the "Native" prefix to everything native, or
 * add the "Non-native" (or "ERC20") prefix to everything non-native. 

The problem with both approach is that we end up with labels that some could consider to be unnecessary long. Do you have a better idea?

src/app/components/RuntimeEvents/RuntimeEventsDetailedList.tsx
```diff
@@ -16,19 +18,33 @@ export const RuntimeEventsDetailedList: FC<{
...
+      })()
+    ) : filtered ? ( // Maybe it's because of filters?
+      <CardEmptyState label={t('tableSearch.noMatchingResults')} />
+    ) : (
+      // There is no special reason, there is just no data
+      <EmptyState description={t('event.cantFindMatchingEvents')} title={t('event.noEvents')} light={true} />
+    )
+  ) : (
+    // Show the data normally
     <>
```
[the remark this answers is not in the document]
csillag: Indeed it is, for the following reasons:
 - Minimizes redundancy: no repeated `if` and `return` statements
 - Because of indentation, the shape of the code kinda reflects the actual decision tree.

But this is not the hill I wanna die on, so I refactored it to use separate simple `if () return` structures.

PR #2091, opened 2025-07-25
Adjust highlighted link vertical alignment

csillag:
This is an unrelated fix extracted from #2072.

With the current CSS, links that support the hover highlight feature appear ever so slightly bellow the normal baseline, because of margin settings. This PR fixes that, restoring correct alignment.

|Before|After|
|---|---|
| image | image  |
|image | image  |

(The second screenshot is from the new endorsement display which is not part of the master branch yet, but you get the point.)

### DIFF
--- a/src/app/components/HighlightingContext/WithHighlighting.tsx
+++ b/src/app/components/HighlightingContext/WithHighlighting.tsx
@@ -25,7 +25,7 @@ export const WithHighlighting: FC<{ children: ReactNode; address: string }> = ({
         alignItems: 'center',
         verticalAlign: 'middle',
         padding: '2px 4px',
-        margin: '-3px -5px',
+        margin: '-5px -5px -3px -5px',
         ...(isHighlighted
           ? {
               background: COLORS.warningLight,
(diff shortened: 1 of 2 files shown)


src/app/components/HighlightingContext/WithHighlighting.tsx
```diff
@@ -25,7 +25,7 @@ export const WithHighlighting: FC<{ children: ReactNode; address: string }> = ({
         alignItems: 'center',
         verticalAlign: 'middle',
         padding: '2px 4px',
-        margin: '-3px -5px',
+        margin: '-5px -5px -3px -5px',
```
buberdds, line 28: margin was introduce in https://github.com/oasisprotocol/explorer/pull/1997#discussion_r2115003982. it takes into accont padding and border. How did we end up with -5px ?

csillag: Here is the screenshot after the fix in 1997:

image

Looking at the "last activity" row, it is visibly off: the address is too low.

The previous margin setting was `-3px -5px`, which translates to `-3px -5px -3px -5px`.

Now I am changing the top from "-3px" to "-5px", to pull it higher.

Visual observation confirms that this gets the vertical alignment right. Looking at old screenshots confirms that the previous setting was never fully aesthetically correct, so this is not something that will now look OK here, but break somewhere else. (Hopefully.)

PR #2094, opened 2025-07-26
Fix testnet 404 text colors

### LINKED ISSUE #2093: "No results" text on testnet search is not visible
https://explorer.dev.oasis.io/testnet/sapphire/search?q=0x4b5ca97d1F45a8b589c0C161ebB258D50F756468

Image

broken since https://github.com/oasisprotocol/explorer/pull/2050

cc @csillag

csillag:
Background: earlier, we had a 404 dialog with a bg that was mostly transparent, so it was shown in changing dark and light colors, depending on the current theme.

However, during a recent re-design, this concept has been dropped; the bg is no longer transparent, it's always blue.

Therefore, the text color now always must be white. However, this bit was missed during implementation, because it looked OK on mainnet.

This PR fixes that: sets the text color to white, independently of the current there.

Fixes #2093

| |Before | After |
| --- |---|---|
| Mainnet | image  |  image |
| Testnet | image  | image  |

### DIFF
--- a/src/app/components/EmptyState/index.tsx
+++ b/src/app/components/EmptyState/index.tsx
@@ -42,17 +42,7 @@ type EmptyStateProps = {
 
 export const EmptyState: FC<EmptyStateProps> = ({ description, title, light, minHeight = '360px' }) => {
   const content = (
-    <Box
-      sx={{
-        ...(light
-          ? { color: 'inherit' }
-          : {
-              color: COLORS.white,
-              '& span a': { color: COLORS.white },
-            }),
-        textAlign: 'center',
-      }}
-    >
+    <Box sx={{ color: light ? 'inherit' : COLORS.white, textAlign: 'center' }}>
       <Typography component="span" sx={{ fontSize: '30px', fontWeight: 500, display: 'block' }}>
         {title}
       </Typography>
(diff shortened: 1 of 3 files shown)


src/app/components/EmptyState/index.tsx
```diff
@@ -42,9 +41,18 @@ type EmptyStateProps = {
...
+      sx={{
+        ...(light
+          ? { color: 'inherit' }
+          : {
+              color: COLORS.white,
+              '& a': { color: COLORS.white },
+            }),
+        textAlign: 'center',
+      }}
+    >
```
lukaw3d, line 55: I see https://pr-2094.oasis-explorer.pages.dev/testnet/sapphire/search?q=asdf
image

different than screenshot

src/app/components/EmptyState/index.tsx
```diff
@@ -5,7 +5,6 @@ import { styled } from '@mui/material/styles'
 import lightBackgroundEmptyState from './images/background-empty-state.svg'
```
lukaw3d, line 5: commit+pr description: transient !== transparent

buberdds (approved):

PR #2101, opened 2025-07-29
Mobile: don't show the layer picker button when already in layer picker

csillag:
Both in desktop and in mobile mode, when using the app, we have a button at the top, for launching the layer picker dialog.

Clicking this button brings up this dialog.

For some reason, in the mobile view, even when this dialog was up, the button (for launching the dialog) was still visible.

I think this serves no useful purpose.
- For opening the dialog, we don't need it since it's open.
- We also don't need it for closing the dialog, because there is a big cancel button for that.
- The only marginally useful thing it still does is indicating where we come from (what was the current layer before opening the selector, but is not strictly required, because it's already marked as "active" in the menu.
- It's presence is not consistent with the desktop mode
- It's presence is counter intuitive, because we don't expect to see the button when we have already pushed it and it has already served it's purpose.

Therefore I propose removing it.

| State | Before | After | Comment |
| --- | --- | --- | --- |
| Desktop mode, selector closed | image  | image  | App in default view, button is visible |
| Desktop mode, selector open |  image | image  | Button is now hidden, since we are inside the dialog |
| Mobile mode, selector closed |  image  |  image  | App in default view, button is visible |
| Mobile mode, selector open, network list |  image  | image  | **Changed.** |
| Mobile mode, selector open, layer list | image  |  image   | **Changed.** |
| Mobile mode, selector open, layer detaild |  image  | image  | **Changed.** |


src/app/components/LayerPicker/index.tsx
```diff
@@ -137,12 +136,6 @@ const LayerPickerContent: FC<LayerPickerContentProps> = ({ isOutOfDate, onClose,
                 </TabletBackButton>
               )}
           </div>
-          <MobileNetworkButton
```
lukaw3d, line 140: I expect this was intentional design for visual balance

csillag: Maybe so. I spelled out my rationale in the PR description, and asked @juresobocan for input.

PR #2104, opened 2025-07-30
Upgrade orval to 7.10.0

lukaw3d:
- new version should fix code re-ordering issues (randomly when v1.yaml is updated the order of generated interfaces was changing)
- needed to downgrade @apidevtools/swagger-parser. We should make a small reproducible example for upstream
- needed to manually revert unknown types to any


csillag:
> needed to manually revert unknown types to any

Does this need to be done manually after every execution of orval?

Issue #2096, opened 2025-07-28
[Bug]: Spacing around Highlighting


buberdds:
Seems like we are missing some spacing between components that are used along side highlighting feature. 

This is just one example, but we need to validate this across app:
<img width="361" height="193" alt="Image" src="https://github.com/user-attachments/assets/f15efc48-e4c7-4ac9-a90a-06c922126fda" />

csillag:
I went through the code, looking at all the locations where `AccountLink` is used. I identified 3 situations when spacing was inadequate, and fixed them all individually in the PR referenced above. (See screenshots there.)

csillag:
Btw, when going through the code, I saw that in many locations, we were already using `&nbsp;` s to add spacing around the addresses, so this was definitely the approach used up to now (as opposed to add padding to the address link component itself.) So the changes in the PR are consistent with the existing approach, just make it more consistent across the codebase.

csillag:
> It is introduced by highlighting. highlighting was used without negative margin initially as it was adding white spaces. Later we added negative margins to it, but we did not align all components.

It is true that we added negative margin to the left, but only to the extend to cover the highlighting bubble. It didn't remove any pre-existing padding.

As I said, I run some experience, including completely removing the highlighting, and the spacing issue in the token title card persisted. That's why I am saying that highlighting is not the reason here.

lukaw3d:
Ah I broke this in https://github.com/oasisprotocol/explorer/pull/2003/files#diff-57eb49205f587d798a061258b0cf7f5e9af60f1e41407705bce16c8d064c7d90L58-L73
wanted to remove extra spaces when noLink; but other components relied on extra spaces bug

buberdds:
<img width="787" height="229" alt="Image" src="https://github.com/user-attachments/assets/182c0510-cb85-4e30-a936-7ab47786731e" />
<img width="1013" height="258" alt="Image" src="https://github.com/user-attachments/assets/afad4474-a4b6-4546-a50b-a7fb80837f14" />

buberdds:
if you look at border, highlighting is cut off at the bottom and at the top

csillag:
Some fixes are here: https://github.com/oasisprotocol/explorer/pull/2107

csillag:
Other half of fix is here: https://github.com/oasisprotocol/explorer/pull/2108

csillag:
[...]

PR #2107, opened 2025-07-30
AddressLink formatting fixes

csillag:
Small adjustments in AddressLink CSS. 

| | Before | After | |
| --- | --- | --- | --- |
| Transfer events squeezed small | image  | image  | |
| Minting event |  image | image  | |

Improves #2096.

### DIFF
--- a/src/app/components/AdaptiveTrimmer/hooks.ts
+++ b/src/app/components/AdaptiveTrimmer/hooks.ts
@@ -244,8 +244,19 @@ export const useAdaptiveSizing = (
           const newLargestKnownGood = Math.max(currentLength, largestKnownGood)
           setLargestKnownGood(currentLength)
 
-          if (currentLength + 1 === smallestKnownBad) {
-            // This the best we can do, for now
+          if (currentLength === smallestKnownBad) {
+            // Note: normally, when there is no overflow,
+            // currentLength should always be lower than smallestKnownBad.
+            // However, once in a while, it's possible that it will be
+            // the same value, which means that because of some race
+            // condition during rendering, at one time,
+            // we had an overflow at this length, but now we don't.
+            // In that case, let's just stop here.
+            debugLog('Anomalous results at', currentLength, '. Stopping.')
+            setAdjustmentStep('done')
+          } else if (currentLength + 1 === smallestKnownBad) {
+            // This the best we can do, for now.
+            // One more character would be too much.
             debugLog(currentLength, 'is as long as we can grow.')
             setAdjustmentStep('done')
           } else {


src/app/components/Account/AccountLink.tsx
```diff
@@ -168,12 +165,12 @@ const DesktopAccountLink: FC<
     <WithTypographyAndLink scope={scope} address={address} labelOnly={labelOnly}>
       <MaybeWithTooltip title={tooltipTitle}>
         {showAccountName ? (
-          <Box
-            component="span"
-            sx={{ display: 'inline-flex', alignItems: 'center', gap: 3, flexWrap: 'wrap' }}
-          >
-            <AccountMetadataSourceIndicator source={accountMetadata!.source} />
-            <HighlightedText text={accountName} pattern={highlightPattern} /> ({address})
```
lukaw3d, line 176: This code worked before https://github.com/oasisprotocol/explorer/pull/2072. What changed in #2072? 
This is now fixing it _differently_ so I'm concerned something else will break.

csillag: Short version: #2072 didn't create the error, just exposed it by triggering it accidentally. Now I fix both the original problem and the accidental trigger.

Here is the full history.

- There was a error in `AdaptivelyTrimmedAccountLink`: under specific conditions (if there is sufficient width to show both account name and address at the same line), it rendered ugly content.
- However, this error was not triggered very often, because `AdaptivelyTrimmedAccountLink` was only used when we were short on space.
- However, in #2072, when introducing an option for always enabling the adaptive trimmer in `AccountLink` (even if we are on desktop and have lots of space), I have [introduced an error](https://github.com/oasisprotocol/explorer/pull/2072/commits/0dca948584f02e0ae74e07fed2c9b62f6c22a054#diff-0f9929174db10e131fdee020ed7a4eb1172794a33cd49e7760f4d63948a851d8R247) by checking the wrong variable in a condition.  (I meant to check the new option, `alwaysAdapt`, but I accidentally wrote `alwaysTrimOnTablet`.) This resulted in accidentally using `AdaptivelyTrimmedAccountLink` in unintended situations, including situations with lots of available width, therefore triggering this pre-existing, but previously hidden error.
- This PR [fixes](https://github.com/oasisprotocol/explorer/pull/2107/commits/c2d620038ff9127af7e0fb69896f18840186581b) the condition (for selecting the right kind of link), _plus_ it [improves](https://github.com/oasisprotocol/explorer/pull/2107/commits/780afd48ac420d41b371980de13529edfe2d29c0) both `AdaptivelyTrimmedAccountLink` and `DesktopAccountLink` to better handle wrapping / non-wrapping situations, thereby eliminating the original error.

csillag: > so I'm concerned something else will break.

That's not an unreasonable assumption; that is why I did manual testing by slowly changing the width and checking the behavior on multiple pages. I don't see anything broken.

lukaw3d: ohh that makes sense

PR #2123, opened 2025-08-08
Enable validator uptime

buberdds:


src/app/components/UptimeStatus/index.tsx
```diff
@@ -1,61 +1,64 @@
...
+          .map((value, index) => (
+            <div
+              key={`${value}-${index}`}
+              className={cn(
+                'inline-block',
+                small
+                  ? 'w-[3px] min-w-[3px] h-[15px] rounded-[2px] mr-[1px]'
+                  : 'w-[7px] min-w-[7px] h-[45px] rounded-[4px] mr-[2px]',
+              )}
+              style={{ backgroundColor: getUptimeItemColor(value), color: COLORS.white }}
```
csillag, line 52: Current example:
`[ 1200, 1054, 1037, 1047, 929, 1200, 939, 1055, 870, 1200, 947, 1198 ]` ==> 
image

If I understand this correctly, we only show red/green bars, and nothing else. The number goes 0-1200, and if we are above 999 it's green, if below, red. But there is no distinction between 100 and 900, or 1000 and 1200. Could we, for example, make the bar length proportionate to the actual number? (Keeping the color of course, which is a good first glance indication.)

csillag (approved):
Reviewed. LGTM.

csillag:
> My understanding was that uptime is a binary concept - it's either `up` or `down`. I'm not sure what the 0-1200 range indicates (for example, when do we show `200` and when do we show `300`?).
> 
> If that range is relevant, the height of the bars makes sense to me.

Basically, each bar signifies a 2-hour long period of time. In each 2-hour period, 1200 blocks are generated by the network. The number means how much of these 1200 blocks were signed by this validator... which basically means how much was it alive and working, in that 2 hours.

So, a number of 1200 means that this validator was working full-time in the 2 hours, 600 would mean that it was offline for half of the time.

PR #2163, opened 2025-08-28
Swallow any errors when opportunistically looking up token names

csillag:
Any time we display an address, we want to check out if it has a token name. However, if no such token exists, we get a 404 error, and hence, a big ugly exception on the console, for each lookup attepmts. It looks like this:

image

This change makes it so that these errors are swallowed. (Not all token lookup errors, just the opportunistic ones, for finding out if they have a name.)

After this change, even in development mode, all we see on the console, after 5 look up attempts, is this:

image

### DIFF
--- a/.changelog/2163.internal.md
+++ b/.changelog/2163.internal.md
@@ -1,0 +1,1 @@
+Swallow any errors when opportunistically looking up token names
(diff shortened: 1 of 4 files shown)


buberdds:
Claiming that app throws useless errors and we should swallow them is imo wrong direction. There is a bug. Bug is in `useAccountMetadata`. It does not correctly check whether we deal with contract or not and sets `enabled` arg incorrectly when calling `useTokenInfo` hook.

lukaw3d:
Maybe just be a bit stricter: if 404: swallow.

PR #2178, opened 2025-09-03
Simplify hover highlighting v1 (no border)

csillag:
Earlier we had some difficulty about padding/margin about addresses that we might want to highlight on hover, because we wanted a bubble around the highlighted part. Unfortunately this makes correct spacing next to impossible, so now we are giving it up.

From now, hover highlight will simply change the background, and nothing else.

This makes all workaround unnecessary.

Also, fix some more vertical alignment issues around links, so that everything lines up nicely.

|  | Before | After: v1 (this PR) | After: v2, (#2182) | After: v3 (#2183) |
| --- | --- | --- |--- | --- |
| TX detail summery with unnamed account at 120% zoom | image  | image | image | image  |
| TX detail summary with multiple named accounts |  image  | image  | image  | image  |
| Endorsements | image  | image  | image  | image  |
| NFT owner link | image  |  image | image   | image  |

### DIFF
--- a/.changelog/2178.internal.md
+++ b/.changelog/2178.internal.md
@@ -1,0 +1,1 @@
+Simplify hover highlighting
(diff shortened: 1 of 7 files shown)


csillag:
Decision from @donouwens:

> from what I can see and understand, I think **_option 2_** is the most reliable, scalable, and solves the UX purpose best.

So, I'm closing v1 and v3, and somebody should review v2.

PR #2183, opened 2025-09-04
Simplify hover highlight v3 (keep the border and -1px horizontal margin)

csillag:
This is an alternate implementation of #2178 and #2182, which keeps the border around the hover highlight, but adds 1px of negative horizontal margin. This means that we will have proper vertical alignment, but use one pixel above and below every address. We might see some clipping at the left and right sides, if the address is squeezed somewhere really tightly. (I haven't seen this manifest anywhere yet.)

See screenshots at #2178.

### DIFF
--- a/.changelog/2178.internal.md
+++ b/.changelog/2178.internal.md
@@ -1,0 +1,1 @@
+Simplify hover highlighting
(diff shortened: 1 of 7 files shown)


csillag:
Decision from @donouwens:

> from what I can see and understand, I think **_option 2_** is the most reliable, scalable, and solves the UX purpose best.

So, I'm closing this (v3)

Issue #2204, opened 2025-09-16
Paratime Dashboard Learning Materials height


kaja-osojnik:
Learning materials not taking up full height of the card on ParaTime dashboard on desktop.

<img width="726" height="489" alt="Image" src="https://github.com/user-attachments/assets/4e504c86-f746-4d41-8360-1192cb2b4d22" />

PR #2205, opened 2025-09-16
Learning Materials content height

kaja-osojnik:
Fixes [2204](https://github.com/oasisprotocol/explorer/issues/2204)


src/app/components/LearningMaterialsCard/index.tsx
```diff
@@ -27,7 +27,9 @@ export const LearningMaterialsCard: FC<LearningMaterialsCardProps> = ({ children
           {t('common.viewAll')}
         </Link>
       </div>
-      <CardContent>{children}</CardContent>
+      <CardContent sx={{ height: '-webkit-fill-available', paddingBottom: '0!important' }}>
```
buberdds, line 30: can we use `flex: 1` instead of `height: '-webkit-fill-available'`?

Issue #2197, opened 2025-09-11
Update monospace typography


kaja-osojnik:
Update monospace MUI typography in the design and project, possibly as a variant in ui library as well.

PR #2208, opened 2025-09-23
Update mono fonts on dashboards

kaja-osojnik:
Replace mono fonts in tables on dashboard pages.
Partial implementation of issue [#2197](https://github.com/oasisprotocol/explorer/issues/2197)

Before:
screenshot_explorer_before

After:
screenshot_explorer_after

### DIFF
--- a/src/app/components/Account/AccountLink.tsx
+++ b/src/app/components/Account/AccountLink.tsx
@@ -1,10 +1,8 @@
 import { FC, ReactNode } from 'react'
 import { Link as RouterLink } from 'react-router-dom'
 import { useScreenSize } from '../../hooks/useScreensize'
-import Link from '@mui/material/Link'
 import { RouteUtils } from '../../utils/route-utils'
 import InfoIcon from '@mui/icons-material/Info'
-import Typography from '@mui/material/Typography'
 import { SearchScope } from '../../../types/searchScope'
 import { useAccountMetadata } from '../../hooks/useAccountMetadata'
 import { trimLongString } from '../../utils/trimLongString'
@@ -15,6 +13,7 @@ import { AdaptiveHighlightedText } from '../HighlightedText/AdaptiveHighlightedT
 import { AdaptiveTrimmer } from '../AdaptiveTrimmer/AdaptiveTrimmer'
 import { AccountMetadataSourceIndicator } from './AccountMetadataSourceIndicator'
 import { WithHoverHighlighting } from '../HoverHighlightingContext/WithHoverHighlighting'
+import { Link } from '@oasisprotocol/ui-library/src/components/link'
 
 const WithTypographyAndLink: FC<{
   scope: SearchScope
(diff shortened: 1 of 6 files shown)


src/app/components/Account/AccountLink.tsx
```diff
@@ -26,15 +24,9 @@ const WithTypographyAndLink: FC<{
...
-        {labelOnly ? (
-          children
-        ) : (
-          <Link component={RouterLink} to={to} sx={{ display: 'inline-flex' }}>
-            {children}
-          </Link>
-        )}
-      </Typography>
+      <span className="inline-flex font-medium text-primary">
+        {labelOnly ? children : <RouterLink to={to}>{children}</RouterLink>}
```
buberdds, line 28: This is different than prev `{labelOnly ?...` condition. Is this expected or we should keep prev logic ? In other places you keep diff color for labelOnly

src/app/components/Account/AccountLink.tsx
```diff
@@ -26,15 +24,9 @@ const WithTypographyAndLink: FC<{
...
-      <Typography variant="mono" component="span" sx={{ display: 'inline-flex' }}>
-        {labelOnly ? (
-          children
-        ) : (
-          <Link component={RouterLink} to={to} sx={{ display: 'inline-flex' }}>
-            {children}
-          </Link>
-        )}
-      </Typography>
+      <span className="inline-flex font-medium text-primary">
```
buberdds, line 27: 1) I wonder if it's not better to keep using Link (but from UIL). This way we will have color defined in one place.
2) Why inline-flex is needed?

src/app/components/Link/index.tsx
```diff
@@ -156,23 +152,15 @@ const DesktopLink: FC<DesktopLinkProps> = ({ address, name, to, alwaysTrim, trim
...
   return (
     <WithHoverHighlighting address={address}>
-      {labelOnly ? (
-        <LinkLabel>{label}</LinkLabel>
-      ) : (
-        <MuiLink component={RouterLink} to={to}>
-          {label}
-        </MuiLink>
-      )}
+      {labelOnly ? <LinkLabel>{label}</LinkLabel> : <RouterLink to={to}>{label}</RouterLink>}
```
buberdds, line 163: This is not the same as  <MuiLink component={RouterLink} as it's missing styling from Link component, right?

src/app/components/Link/index.tsx
```diff
@@ -3,7 +3,6 @@ import { Link as RouterLink } from 'react-router-dom'
 import { useScreenSize } from '../../hooks/useScreensize'
 import MuiLink from '@mui/material/Link'
 import Typography from '@mui/material/Typography'
```
buberdds, line 5: Typography not migrated to UIL

src/app/components/Account/AccountLink.tsx
buberdds: This applies to all places in this PR. Also let's keep Link name not UilLink. We do not prefix Oasis UI Lib components.

kaja-osojnik: Updated in the AccountLink.tsx. in Link/index.tsx I renamed it to UilLink to avoid a name clash with our own Link wrapper in this file.

buberdds (approved):

PR #2213, opened 2025-09-26
Trigger workflows for stable/* branches

buberdds:


.github/workflows/cloudflare-pages.yml
```diff
@@ -5,11 +5,14 @@ name: cloudflare-pages
 on:
   # A push occurs to one of the matched branches.
   push:
-    branches: [master]
+    branches:
+      - master
```
lukaw3d, line 9: missed here?

PR #2228, opened 2025-10-02
Make BuildBanner non-sticky to fix on mobile

lukaw3d:
Our sticky AppBar is now shorter so BuildBanner became slightly visible behind it. BuildBanner doesn't really need to stay sticky, so that fixes it.

Recording 2025-10-02 at 21 57 39

### DIFF
--- a/src/app/components/BuildBanner/index.tsx
+++ b/src/app/components/BuildBanner/index.tsx
@@ -20,7 +20,11 @@ export const BuildBanner: FC = () => {
     : t('banner.buildPreview', { appTitle: getAppTitle() })
 
   return (
-    <Alert variant="warning-filled" sticky>
+    <Alert
+      variant="warning-filled"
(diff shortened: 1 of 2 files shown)


src/app/components/BuildBanner/index.tsx
```diff
@@ -19,9 +19,5 @@ export const BuildBanner: FC = () => {
     ? t('banner.buildStaging')
     : t('banner.buildPreview', { appTitle: getAppTitle() })
 
-  return (
-    <Alert variant="warning-filled" sticky>
-      {message}
-    </Alert>
-  )
+  return <Alert variant="warning-filled">{message}</Alert>
```
buberdds, line 22: sticky prop applies different styles (not related to positioning only)

lukaw3d: :scream:

Issue #2222, opened 2025-09-30
MUI Box component replacement


kaja-osojnik:
Replace all MUI <Box> components across the project with appropriate elements and tailwind styling as a part of transition from MUI to Oasis ui library using shadcn.

kaja-osojnik:
Implemented in following PR's:
- AccountLink [#2223](https://github.com/oasisprotocol/explorer/pull/2223)

PR #2254, opened 2025-10-13
Update Box element

kaja-osojnik:
Partial implementation of issue [#2222](https://github.com/oasisprotocol/explorer/issues/2222).

### DIFF
--- a/src/app/components/HighlightedText/index.tsx
+++ b/src/app/components/HighlightedText/index.tsx
@@ -1,8 +1,6 @@
 import * as React from 'react'
 import { findTextMatches, NormalizerOptions, PositiveMatchInfo } from './text-matching'
 import { FC, ReactNode } from 'react'
-import { SxProps } from '@mui/material/styles'
-import Box from '@mui/material/Box'
 import { useHighlightPattern } from '../PatternHighlightingContext'
 
 export interface HighlightOptions {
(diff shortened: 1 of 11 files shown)


src/app/components/NoPreview/index.tsx
```diff
@@ -12,21 +10,15 @@ export const NoPreview: FC<NoPreviewProps> = ({ placeholderSize }) => {
   const { t } = useTranslation()
 
   return (
-    <Box
-      gap={4}
-      sx={{
+    <div
+      className="flex flex-1 flex-col items-center justify-center gap-4 text-gray-500 font-semibold bg-gray-100"
```
buberdds: - let's use text-muted-foreground instead of tailwind text-gray-500 
- why  bg-gray-100 was added ?

src/app/components/HighlightedText/index.tsx
```diff
@@ -62,16 +43,14 @@ interface HighlightedTextProps {
...
  */
-export const HighlightedText: FC<HighlightedTextProps> = ({
-  text,
-  partsToHighlight,
-  options = defaultHighlight,
-}) => {
+export const HighlightedText: FC<HighlightedTextProps> = ({ text, partsToHighlight, options = {} }) => {
   const pattern = useHighlightPattern()
-  const { sx = defaultHighlightStyle, findOptions = {} } = options
+  const { className = defaultHighlightClass, findOptions = {} } = options
```
buberdds: is there any use case for passing className to hightlight ?

PR #2255, opened 2025-10-14
Replace MUI dependencies in Search with Oasis UI Library

buberdds:


kaja-osojnik (changes requested):
Search icon on mobile not centered with heading horizontally:
Screenshot 2025-10-16 at 12 08 00

PR #2264, opened 2025-10-20
Update Box element

kaja-osojnik:
Partial implementation of issue [#2222](https://github.com/oasisprotocol/explorer/issues/2222).

### DIFF
--- a/src/app/components/AppendMobileSearch/index.tsx
+++ b/src/app/components/AppendMobileSearch/index.tsx
@@ -23,6 +23,7 @@ export const AppendMobileSearch: FC<PropsWithChildren<AppendMobileSearchProps> &
         'relative w-full',
         action ? 'grid grid-cols-[1fr_auto_1fr]' : 'flex justify-between',
         isMobile ? 'items-center' : 'items-start',
+        enableMobileSearch ? 'justify-between' : 'justify-center',
       )}
     >
       <div>{children}</div>
--- a/src/app/pages/HomePage/index.tsx
+++ b/src/app/pages/HomePage/index.tsx
@@ -1,7 +1,5 @@
 import { FC, useState } from 'react'
 import { Logotype } from '../../components/PageLayout/Logotype'
-import { styled } from '@mui/material/styles'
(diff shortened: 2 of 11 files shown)


src/app/pages/ProposalDetailsPage/ProposalVotesCard.tsx
```diff
@@ -140,9 +139,9 @@ export const ProposalVotesCard: FC = () => {
         </div>
       </div>
       <ErrorBoundary light={true}>
-        <Box sx={{ height: '704px' }}>
+        <div className="h-[704px]">
```
buberdds: can we remove this hardcoded height? does not make sense to limit table wrapper this way

src/app/components/AppendMobileSearch/index.tsx
```diff
@@ -23,6 +23,7 @@ export const AppendMobileSearch: FC<PropsWithChildren<AppendMobileSearchProps> &
         'relative w-full',
         action ? 'grid grid-cols-[1fr_auto_1fr]' : 'flex justify-between',
         isMobile ? 'items-center' : 'items-start',
+        enableMobileSearch ? 'justify-between' : 'justify-center',
```
buberdds: why this was added? how this affects layout ?

src/app/pages/HomePage/index.tsx
```diff
@@ -25,94 +24,6 @@ export const zIndexHomePage = {
   mobileTooltip: 4,
 }
 
-const HomepageLayout = styled(Box)(({ theme }) => ({
```
buberdds: we can omit migration of HomePage and its components in next PRs as we gonna nuke that page and replace with completely new layout.

src/app/components/AppendMobileSearch/index.tsx
kaja-osojnik: To center the footer on homepage on mobile:
Before:
Screenshot 2025-10-22 at 11 23 54

After:
Screenshot 2025-10-22 at 11 24 37

Issue #2153, opened 2025-08-25
[UX]: Add ability to see which node operators run Sapphire (or any other ParaTime)


tjanez:
### IMPACT

- [ ] High
- [ ] Medium
- [x] Low

### URGENCY

- [ ] Critical
- [ ] Major
- [x] Medium
- [ ] Minor

### DESCRIPTION OF ISSUE

Currently, one is only able to see node operators which run as Consensus validators.

It would be great to be able to see which node operators run ParaTime compute nodes (e.g. Sapphire compute notes).

And vice-versa, for a given node operator to be able to see which nodes (e.g. Consensus, Sapphire, Cipher, ...) it runs.

### DEVICE / VIEWPORT

- [x] Desktop
- [x] Mobile

donouwens:
Implemented initial feedback and awaiting input for next iteration. Might have to move over to the next sprint.

PR #2302, opened 2025-11-03
Implement new homepage

buberdds:
Closes https://github.com/oasisprotocol/explorer/issues/2238.

Extracted PRs:
- https://github.com/oasisprotocol/explorer/pull/2313
- https://github.com/oasisprotocol/explorer/pull/2344
- https://github.com/oasisprotocol/explorer/pull/2345
- https://github.com/oasisprotocol/explorer/pull/2346

### DIFF
--- a/src/app/pages/HomePage/EcosystemCard.tsx
+++ b/src/app/pages/HomePage/EcosystemCard.tsx
@@ -1,7 +1,8 @@
 import { FC, ReactNode } from 'react'
-import { Link } from 'react-router-dom'
 import { useTranslation } from 'react-i18next'
 import { ArrowRight } from 'lucide-react'
+import { Link as RouterLink } from 'react-router-dom'
+import { Link } from '@oasisprotocol/ui-library/src/components/link'
 import {
   Card,
   CardContent,
@@ -13,6 +14,8 @@ import { Button } from '@oasisprotocol/ui-library/src/components/ui/button'
 import { Typography } from '@oasisprotocol/ui-library/src/components/typography'
 import { Badge } from '@oasisprotocol/ui-library/src/components/badge'
 import { Skeleton } from '@oasisprotocol/ui-library/src/components/ui/skeleton'
+import { Tooltip } from '@oasisprotocol/ui-library/src/components/tooltip'
+import { Network } from '../../../types/network'
 import { RouteUtils } from '../../utils/route-utils'
 
 type EcosystemCardProps = {
@@ -26,6 +29,7 @@ type EcosystemCardProps = {
   title: ReactNode
   description: ReactNode
   legacy?: boolean
+  network?: Network
 }
 
 export const EcosystemCard: FC<EcosystemCardProps> = ({
@@ -39,6 +43,7 @@ export const EcosystemCard: FC<EcosystemCardProps> = ({
   title,
   description,
   legacy,
+  network = 'mainnet',
 }) => {
   const { t } = useTranslation()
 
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -70,11 +70,8 @@
     "react": "18.3.1",
     "react-dom": "18.3.1",
     "react-i18next": "14.1.3",
-    "react-quick-pinch-zoom": "5.1.0",
     "react-router-dom": "6.30.0",
     "recharts": "2.15.1",
-    "swiper": "11.2.5",
-    "use-resize-observer": "9.1.0",
```
lukaw3d: You removed dependencies but yarn.lock added 1000 lines
something's off

lukaw3d: oops its because I merged #2188 and thought removed lines were from vite deduplicating.

In reality we need https://github.com/oasisprotocol/rose-app/pull/307 and nevermind   https://github.com/oasisprotocol/rose-app/pull/307#issuecomment-3478469653

lukaw3d: Fixed in https://github.com/oasisprotocol/explorer/pull/2309

buberdds: thanks

tjanez:
@buberdds, thanks for working on this.

A few quick to implement suggestions (cc @juresobocan):
1. Can we make a lot more things in the Ecosystem cards clickable?
    I think users would try to click on all the blue things and would be a bit disappointed to see that clicking on e.g. block number doesn't take them to that block's details page.
    The things I would make clickable:
    - Block number (e.g. 27,343,283) on all 4 cards
    - Active nodes number (e.g. 105) on all 4 cards. I'm unsure if we currently have a page to show the node operators that run Sapphire or Emerald? But for Consensus that should be there.
    - Name of the chain (e.g. Sapphire, Consensus). I find it most intuitive that people would expect to click on the heading (or whole first row) to be able to navigate to the chosen chain.

2. I find the "deltaDao" in italics a bit awkward:
    image
    Can we try with non-italic variant?

3. I find the main action item on each card, i.e. the "Mainnet ->" button, getting lost in the gray background and would not recognize it as clickable/actionable at first sight:
    image
    Can we make it similarly prominent as e.g. action items in the ROFL App?

4. We need to provide better copy for all 4 chains' descriptions. I can help with those.

5. I find the "Online" and "Legacy" badges a bit contradictory:
    image
    The original aim was to let users know that we consider Emerald as the Legacy EVM-compatible ParaTime/chain but currently, looking at the above picture, one would think all other 3 are running and Emerald is perhaps down?
    
    We could replace the word "Online" with "Current" (or similar) so we message a consistent thing: some chains are considered current (i.e. people should develop and deploy new things on them) and some chains are considered legacy (i.e. they are still being run, but we expect developers and dApps to migrate from these chains to a newer chain).
    
    Also, is the current "orange" color good for signaling something legacy? I think we used this color to denote an experimental development build of e.g. ROFL App. And I think orange fits well in this sense. Can we try with some other color for legacy things?

buberdds:
for # 5 
Currently
- there is one hardcoded legacy status for Emerald
- other badges indicate indexing status (just like in old homepage): Online (success color) or Out-of-date (destructive color) when Indexing is lagging

tjanez:
@buberdds,  thanks for implementing the changes. A few comments:

1. Making things clickable:
    - Could you make the block number links go to the exact block number?
       It would match the behavior of clicking on the block number in the Latest Blocks table.
    - Can we make at least the Active nodes number for the Consensus clickable and link to https://explorer.dev.oasis.io/mainnet/consensus/validators?
       For the ParaTimes, I think we currently don't have a view of nodes that run a particular ParaTime. We'll need to wait for https://github.com/oasisprotocol/explorer/issues/2153 before we can link there.

2. Pontus-X link appears to not function correctly:

     https://github.com/user-attachments/assets/e0379bce-3ebf-4ae6-9d2b-23afe807791f

     At the bottom, the link says it will go to <https://www.pontus-x.eu/>, but it goes to <https://explorer.oasis.io/testnet/pontusxtest>.

3. Can we implement a quick-fix interim solution to make the buttons have the same style as the Search button next to the search bar above?

    image

    This should make them visible enough for the time being and we could move forward with the new home page.

4. My copy suggestions for the chain cards:
    - Sapphire: "Oasis main EVM-compatible chain. Supports encrypted contract state and end-to-end encrypted transactions and calls."
    - Consensus: "Oasis Layer 1 chain providing scalable, high-throughput, secure, proof-of-stake consensus run by a decentralized set of validator nodes."
    - Pontus-X by deltaDAO: "Confidential EVM-Compatible chain offering smart contract development environment focused on building Decentralized data economy. _(Mainnet coming soon)_"
    - Emerald: "Oasis original EVM-compatible chain launched in 2021. Superseded by Oasis Sapphire which supports confidential state and encrypted transactions."
    
    image

5. Chain status badges

   > For # 5 currently
   > 
   > there is one hardcoded legacy status for Emerald
   > other badges indicate indexing status (just like in old homepage): Online (success color) or Out-of-date
   > (destructive color) when Indexing is lagging

   @buberdds, thanks for explaining this!

   @juresobocan, can we keep the "normal" status badge for Emerald then (i.e. it would show green/online or red/out-of-date) and add a **legacy icon** next to the name similar to how "by deltaDAO" is added to Pontus-X?

    Upon hovering over the legacy icon, we would explain that Emerald is considered a legacy chain and what that entails.

Issue #2146, opened 2025-08-21
[UX]:  Define data for `simple` vs `advanced` data


donouwens:
### IMPACT

- [ ] High
- [ ] Medium
- [ ] Low

### URGENCY

- [ ] Critical
- [ ] Major
- [ ] Medium
- [ ] Minor

### DESCRIPTION OF ISSUE

Based on https://github.com/oasisprotocol/explorer/issues/1918 - define what data will be considered `simple` and what will be considered `advanced` in the explorer. 

### DEVICE / VIEWPORT

- [ ] Desktop
- [ ] Mobile

lukaw3d:
@donouwens if we hide the transaction format, do we want to put encrypted/unencrypted icon somewhere else?

donouwens:
@lukaw3d I think we should keep the ‘Format’ row in both views for consistency as it would be weird to place it somewhere else when switching between basic and advanced view.

PR #2342, opened 2025-11-12
Hide advanced fields

lukaw3d:
Fixes https://github.com/oasisprotocol/explorer/issues/2146

image


buberdds:
Is enabling hide/show feature in vertical lists part of a plan? We will end up with vertical lists even more detached from horizontal one (well it's already out of control). Just curious about general plan not impl.

PR #2347, opened 2025-11-14
Remove usage of MUI useMediaQuery

buberdds:


src/app/hooks/useScreensize.ts
```diff
@@ -1,12 +1,40 @@
...
+
+      setScreenSize({
+        isMobile,
+        isTablet,
+        isLaptop,
+        isDesktop,
+      })
+    }
+
+    window.addEventListener('resize', handleResize)
```
lukaw3d: Previous mui implementation used `mediaQueryList.addEventListener('change', notify);`
https://github.com/mui/material-ui/blob/dba602459c8b50d34ecbf3c6e61a3f805c7cb413/packages/mui-system/src/useMediaQuery/useMediaQuery.ts#L103-L108

to consider if we see any issues

buberdds: switched to media query change listener

Issue #2357, opened 2025-11-21
[UX]: A few nits in the ROFL search results


matevz:
### IMPACT

- [ ] High
- [x] Medium
- [ ] Low

### URGENCY

- [ ] Critical
- [ ] Major
- [ ] Medium
- [x] Minor

### PRODUCT ENVIRONMENT

KUbuntu 24

### DESCRIPTION OF ISSUE

Search results for ROFL need some love. See below.

### DEVICE / VIEWPORT

- [x] Desktop
- [x] Mobile

### STEPS TO REPRODUCE

1. Open explorer
2. Search for some ROFL, e.g. `demo-rofl`
3. Currently the search results look like this:

<img width="2236" height="1340" alt="Image" src="https://github.com/user-attachments/assets/d23b546e-f419-4512-aa48-37e5b67a3a45" />

### ALTERNATIVES AND SUGGESTIONS

1. One would assume the first line is the thing you need to click on to open it, but it points to "Sapphire ParaTime". Move the ROFL name in the first line.
2. Make ROFL name clickable and open the ROFL page. Also add a copy button.
3. The "View ROFL App" seems both out of place and we try to avoid the term "ROFL App" wherever possible. I'd just remove it.
4. I think we can also hide the "ParaTime", "TEE" field and "Active replicas" in the results to declutter a bit.

Issue #2332, opened 2025-11-12
Icons around named accounts


lukaw3d:
<img width="785" height="217" alt="Image" src="https://github.com/user-attachments/assets/e2d43ee5-b2e6-4e23-8a89-5e85e51c8f74" />

- Bookmark icon (name self-acclaimed)
- Validator name: can contain unicode/emojis "AutoStake 🛡️ Slash Protected"
- Copy icon: isn't obvious what it will copy

all together feels overloaded

PR #2358, opened 2025-11-21
Tweaks around ROFL apps search results

buberdds:
Closes https://github.com/oasisprotocol/explorer/issues/2357

- we need to keep "View App" button to keep search consistent. I am not sure removing it everywhere is expected.
- keeping "ParaTime" line below app name. Scoped search return all networks for a given layer + global search return all networks for all layers. It is easy for users to get lost without this row.
- name copy to clipboard copies app id, right?

global search
https://pr-2358.oasis-explorer.pages.dev/search?q=demo-rofl

testnet scoped search (mainnet still avail at the bottom/collapsible section)
https://pr-2358.oasis-explorer.pages.dev/testnet/sapphire/search?q=demo-rofl

### DIFF
--- a/src/app/pages/RoflAppDetailsPage/index.tsx
+++ b/src/app/pages/RoflAppDetailsPage/index.tsx
@@ -200,12 +200,20 @@ export const RoflAppDetailsViewSearchResult: FC<{
 
   return (
     <StyledDescriptionList>
+      <dt>{t('common.name')}</dt>
+      <dd>
+        <RoflAppLink
+          id={app.id}
+          name={app.metadata['net.oasis.rofl.name']}
+          network={app.network}
+          withSourceIndicator={false}
+        />
+        <CopyToClipboard value={app.metadata['net.oasis.rofl.name']} />
+      </dd>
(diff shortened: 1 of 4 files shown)


src/app/pages/RoflAppDetailsPage/index.tsx
```diff
@@ -200,12 +200,20 @@ export const RoflAppDetailsViewSearchResult: FC<{
...
     <StyledDescriptionList>
+      <dt>{t('common.name')}</dt>
+      <dd>
+        <RoflAppLink
+          id={app.id}
+          name={app.metadata['net.oasis.rofl.name']}
+          network={app.network}
+          withSourceIndicator={false}
+        />
+        <CopyToClipboard value={app.id} />
```
matevz: I would copy the app name, because you may search for a substring and the app name may be way longer and you want to copy its name, e.g. to avoid typos.

lukaw3d: I think this will make https://github.com/oasisprotocol/explorer/issues/2332 more confusing

PR #2365, opened 2025-11-30
Fix display of "x" next to numbers (don't display as "×")

lukaw3d:


buberdds (approved):
to avoid alignment issues we can merge mono font PR and target specific elements (like search input) with `font-variant-ligatures: no-contextual;`

Issue #2377, opened 2025-12-03
"2 years ago" is equal to "1 yr"?


lukaw3d:
https://explorer.dev.oasis.io/mainnet/consensus/block/17407756
<img width="313" height="146" alt="Image" src="https://github.com/user-attachments/assets/3e81555c-d30a-413c-922e-e041d76e84db" />

PR #2401, opened 2025-12-18
Fix inconsistent relative time displays

lukaw3d:
Fixes https://github.com/oasisprotocol/explorer/issues/2377

http://localhost:1234/mainnet/consensus/block/17407756 showed "1 yr" and "2 years ago" in tooltip. This could be fixed with formatDistanceStrict rounding: floor, but then http://localhost:1234/mainnet/consensus/block/26177788 "4 mo" and "3 months ago" in tooltip. So instead, reused formatDistanceToNow.

### DIFF
--- a/src/app/utils/dateFormatter.ts
+++ b/src/app/utils/dateFormatter.ts
@@ -16,9 +16,9 @@ export const intlDateFormat = (date: Date | number) => dateFormat.format(date)
 // TODO: Works only in en-US locale, as suffixes are hardcoded
 export const formatDistanceToNow = (
   date: Date | number,
-  options: { baseDate?: Date | number; locale?: string; keepSuffix?: true } = {},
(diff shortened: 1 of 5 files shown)


src/app/utils/dateFormatter.ts
```diff
@@ -16,9 +16,9 @@ export const intlDateFormat = (date: Date | number) => dateFormat.format(date)
 // TODO: Works only in en-US locale, as suffixes are hardcoded
 export const formatDistanceToNow = (
   date: Date | number,
-  options: { baseDate?: Date | number; locale?: string; keepSuffix?: true } = {},
+  options: { baseDate?: Date | number; locale?: string; keepSuffix?: true, style?: 'short' | 'long' } = {},
```
buberdds: lint

buberdds (approved):

PR #2435, opened 2026-02-13
Migrate from react-router-dom v6 to react-router v7

buberdds:
Should close security alerts
https://github.com/oasisprotocol/explorer/security/dependabot/118
https://github.com/oasisprotocol/explorer/security/dependabot/119

Current v6 throws migration warnings:
```
Warning: React Router will begin wrapping state updates in `React.startTransition` in v7
Warning: Relative route resolution within Splat routes is changing in v7
```
but we can ignore both during migration
- startTransision affects React.lazy, but Monaco editor is declared at module scope 
- routes file has no multi-segment splat paths


lubej (approved):
LGTM

PR #2458, opened 2026-09-04
Remove orval dependency

buberdds:
no need for keep api client generator for Nexus


src/oasis-nexus/package.json
```diff
@@ -1,12 +0,0 @@
-{
-  "scripts": {
-    "start": "orval",
-    "generate": "orval"
```
lubej, line 4: [generate-openapi.yml](https://github.com/oasisprotocol/explorer/blob/master/.github/workflows/generate-openapi.yml) can be removed as well

.changelog/2458.internal.md
```diff
@@ -0,0 +1 @@
+Remove orval dependency
```
lubej, line 1: Ideally we would note down the commit hash somewhere in the README, in case in the future generation is needed. Or maybe README in oasis-nexus folder.

lubej (approved):

